PR #27, opened 2020-01-14
PDPDEVTOOL-1884 additional changes - Ask question to import referenced script files only if at least one customscript object is selected for import.

roctaviano:


src/commands/ImportObjectsCommandGenerator.js
```diff
@@ -202,8 +207,14 @@ module.exports = class ImportObjectsCommandGenerator extends BaseCommandGenerato
...
 		questions.push(questionListObjectsSelection);
+		return questions;
+	}
+
+	_generateQuestionsAfterObjectSelection(selectionObjectAnswers) {
+		const questions = [];
 
-		if (this._projectInfoService.getProjectType() === PROJECT_ACP) {
+		const customScript = selectionObjectAnswers.objects_selected.find(element => element.scriptId.startsWith(CUSTOM_SCRIPT_PREFIX));
+		if (this._projectInfoService.getProjectType() === PROJECT_ACP && customScript) {
```
karl-anthony-ng: Plugins also have a prefix of "customscript". Does the auto-import of references also apply for plugins?

karl-anthony-ng: Plugins also have a prefix of "customscript". Does the auto-import of references also apply for plugins?

roctaviano: Yes, it also applies for plugins.

davidecorreu: I think it will be cleaner to use:
```
const hasCustomScript = selectionObjectAnswers.objects_selected.some(element => element.scriptId.startsWith(CUSTOM_SCRIPT_PREFIX))
```
 because .some returns directly a boolean value if any of the elements in the array accomplishes the test function.

roctaviano: Agree.

PR #28, opened 2020-01-15
feature/PDPDEVTOOL-2095: Add Bank Parser Plugin and FIConnectivityPlugin objects

davidecorreu:
Adding Bank Parser Plugin and FIConnectivityPlugin to the object types metadata

### DIFF
--- a/src/metadata/ObjectTypesMetadata.js
+++ b/src/metadata/ObjectTypesMetadata.js
@@ -273,6 +273,30 @@ module.exports = [
 		},
 	},
 	{
+		name: 'Plug-in - Financial Institution Connectivity',
+		value: {
+			name: 'Plug-in - Financial Institution Connectivity',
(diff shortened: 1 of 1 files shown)


src/metadata/ObjectTypesMetadata.js
```diff
@@ -272,6 +272,30 @@ module.exports = [
...
+		name: 'Plug-in - Financial Institution Connectivity',
+		value: {
+			name: 'Plug-in - Financial Institution Connectivity',
+			type: 'ficonnectivityplugin',
+			prefix: 'customscript_',
+            hasRelatedFiles: true,
+            relatedFiles: [
+                { type:'blankscript' }
+            ]
+		},
```
lfpvillegas: Minor one, check identation please, Maybe there are also more objects types like this in the file.

PR #30, opened 2020-01-20
PDPDEVTOOL-2404: Bug: If there is an error from SDK call, validation errors also appear instead of just the SDK error

roctaviano:


src/commands/ImportObjectsCommandGenerator.js
```diff
@@ -59,50 +59,59 @@ module.exports = class ImportObjectsCommandGenerator extends BaseCommandGenerato
...
 			const listObjectQuestions = this._generateListObjectQuestions();
-			const listObjectAnswers = await prompt(listObjectQuestions);
-
-			const paramsForListObjects = this._arrangeAnswersForListObjects(listObjectAnswers);
-			const executionContextForListObjects = new SDKExecutionContext({
-				command: this._listObjectsMetadata.name,
-				params: paramsForListObjects,
-				includeProjectDefaultAuthId: true,
-			});
+			listObjectAnswers = await prompt(listObjectQuestions);
```
davidecorreu: Do not surround this with a try/catch block. If there is an error reading the manifest file (change the name to manifestQQ,xml), the code won't show an appropriate message.

src/commands/ImportObjectsCommandGenerator.js
```diff
@@ -59,50 +59,59 @@ module.exports = class ImportObjectsCommandGenerator extends BaseCommandGenerato
...
-			const selectionObjectAnswers = await prompt(selectionObjectQuestions);
+		const { data } = listObjectsResult;
+		if (SDKOperationResultUtils.hasErrors(listObjectsResult)) {
+			SDKOperationResultUtils.logResultMessage(listObjectsResult);
+			SDKOperationResultUtils.logErrors(listObjectsResult);
+			throw SDKOperationResultUtils.getErrorMessagesString(listObjectsResult);
+		}
+		if (Array.isArray(data) && listObjectsResult.data.length === 0) {
+			NodeUtils.println(TranslationService.getMessage(MESSAGES.NO_OBJECTS_TO_LIST), NodeUtils.COLORS.RESULT);
+			return;
```
davidecorreu: If we specify a scriptid string that matches nothing in our account during the interactive questions, the flow of execution won't be stopped here, and will return undefined. Consequently we will have as result: 
```
There are no objects matching your search criteria. Select a different set of object types, and if applicable, check the specified application ID or script ID.
There are validation errors:
"scriptid" option is mandatory.
"type" option is mandatory.
"destinationfolder" option is mandatory.
```

davidecorreu: To stop the execution flow we need to 'throw' . So I think we need to just:
```
throw NodeUtils.println(TranslationService.getMessage(MESSAGES.NO_OBJECTS_TO_LIST);
```
It will show a red message (which is nice because we could not perform the import) and stop the execution flow,

PR #35, opened 2020-01-27
Fixing parameter in authentication success message and developmentMode property

vic-nik:


src/commands/SetupCommandGenerator.js
```diff
@@ -340,7 +340,6 @@ module.exports = class SetupCommandGenerator extends BaseCommandGenerator {
 					operationResult.accountInfo.roleName,
 					operationResult.authId
 					);
-				NodeUtils.println(JSON.stringify(operationResult));
```
vic-nik: Accidentally commited to previous PR

PR #36, opened 2020-02-03
Make overwrite confirmation question dynamic - depending on whether to import referenced SuiteScripts or not.

roctaviano:
Make overwrite confirmation question dynamic - depending on whether to import referenced SuiteScripts or not.


packages/node-cli/src/commands/ImportObjectsCommandGenerator.js
```diff
@@ -88,23 +88,26 @@ module.exports = class ImportObjectsCommandGenerator extends BaseCommandGenerato
...
+
+			const overwriteConfirmationQuestion = this._generateOverwriteConfirmationQuestion(anwersAfterObjectSelection);
+			overwriteConfirmationAnswer = await prompt(overwriteConfirmationQuestion);
 		} catch (error) {
 			throw TranslationService.getMessage(PROMPTING_INTERACTIVE_QUESTIONS_FAILED, NodeUtils.lineBreak, error);
-		}			
+		}
 
-		const combinedAnswers = { ...listObjectAnswers, ...selectionObjectAnswers, ...anwersAfterObjectSelection };
+		const combinedAnswers = { ...listObjectAnswers, ...selectionObjectAnswers, ...anwersAfterObjectSelection, ...overwriteConfirmationAnswer };
```
davidecorreu: Maybe we should stop the execution here instead of passing the "overwriteConfirmationAnswer". But as is the first thing checked in _executeAction() I'm fine with it.

PR #38, opened 2020-02-10
Edited README.md to include changes for the 20.1 release

x8a:
I didn't mention the 19.2 version. Should I? If so, what NPM version should I give it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,24 @@ CLI for Node.js is an interactive tool that guides you through all the steps of 
 
 ## Prerequisites
 The following software is required to work with SuiteCloud CLI for Node.js:
-- Node.js version 12.14 LTS
+- Node.js version 12.14.0 LTS or greater
 - Java JDK version 11
 
 Read the full list of prerequisites in [NetSuite's Help Center](https://system.netsuite.com/app/help/helpcenter.nl?fid=section_1558708810.html).
+
+## Supported Versions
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,15 +4,24 @@ CLI for Node.js is an interactive tool that guides you through all the steps of
...
 - Java JDK version 11
 
 Read the full list of prerequisites in [NetSuite's Help Center](https://system.netsuite.com/app/help/helpcenter.nl?fid=section_1558708810.html).
+
+## Supported Versions
+Although you may have access to older versions of CLI for Node.js, only the versions in this section are officially supported.
+
+Each version of CLI for Node.js works optimally with a specific NetSuite version. To check the  equivalence among verisons, see the following table:
+
+| NPM | NetSuite |
```
alii-gator: That seems a bit confusing to me. 
NPM version, or suitecloud cli version?
Npm itself has versions. https://nodejs.org/en/download/releases/

x8a: What I mean to say here is the CLI for Node.js version in NPM, but not the NPM version users need to have installed to make it work. As I understood, we'll need a new version number to be hosted in NPM.

margabit:
Is this the main repository README.md or the one under packages/node-cli?

x8a:
As we discussed, I'll work on the new README files and check with you before submitting them.

PR #44, opened 2020-02-14
PDPDEVTOOL-2712: Could not access system.f.netsuite.com, system.snap.…

vic-nik:
…netsuite.com, system.wg.netsuite.com > in Node CLI


packages/node-cli/__test__/validation/InteractiveAnswersValidator.test.js
```diff
@@ -186,32 +186,3 @@ describe('validateXMLCharacters', () => {
 		expect(validateXMLCharacters("mixing_Some > of them < all 'in' the &same string \"")).toEqual(failureResponse);
 	});
 });
-
-describe('validateNotProductionUrl', () => {
```
lfpvillegas: One question about the story definition. It says that F and Snap environments should also be available with --dev option. 
Is this also appliable to production account system.netsuite.com? In the story doesn't says nothing about it but maybe still makes sense to validate that in this case --dev is not used.

What do you think?

vic-nik: We only allow to provide URL in case -dev falg is introduced. If not system.netsuite.com is used by default. What I am doing in this story is removing any validation on dev URLs, so in the URL field the user will be able to enter anything (should not be empty). If the dev URL provided is invalid it will fail later with server error, e.g. Host not known

lfpvillegas: ok, it has sense!thanks viktor!

PR #45, opened 2020-02-18
PDPDEVTOOL-2735 - Removing postinstall script

jbbalbas:


packages/node-cli/src/SDKExecutor.js
```diff
@@ -39,7 +40,7 @@ module.exports.SDKExecutor = class SDKExecutor {
 			? this._authenticationService.getProjectDefaultAuthId()
 			: null;
 
-		return new Promise((resolve, reject) => {
+		return new Promise(async (resolve, reject) => {
```
vic-nik: What does async solve here?

jbbalbas: We can only use await inside an async method, and since we need to await SDKDownloadService.download() in line 56 we need to transform this into an async method

PR #50, opened 2020-03-02
PDPDEVTOOL-2793 - Renaming cli-config.js to suitecloud.config.js

jbbalbas:


x8a:
The change looks good to me. The docs are also updated with the new name.

PR #51, opened 2020-03-02
Feature/command-generators-to-action-result

karl-anthony-ng:
Enforce the use of ActionResult object as return for CommandGenerators.


packages/node-cli/src/commands/actionresult/ActionResultBuilder.js
```diff
@@ -0,0 +1,35 @@
+/*
+** Copyright (c) 2019 Oracle and/or its affiliates.  All rights reserved.
+** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+*/
+'use strict';
+const ActionResult = require('../actionresult/ActionResult');
+const ActionResultStatus = require('../actionresult/ActionResultStatus');
+
+module.exports = class ActionResultBuilder {
```
margabit: I don't think it makes sense that the builder needs a new instance in order to be used (`new ActionResultBuilder()`).
There are other builder patterns that seem more intuitive to use. See https://medium.com/@axelhadfeg/builder-pattern-using-javascript-and-es6-ec1539182e24

packages/node-cli/src/commands/actionresult/ActionResultStatus.js
```diff
@@ -0,0 +1,8 @@
+/*
+** Copyright (c) 2019 Oracle and/or its affiliates.  All rights reserved.
+** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+*/
+module.exports = {
```
margabit: This could be defined inside the `ActionResult` file and use multiple exports.

packages/node-cli/src/commands/actionresult/ActionResult.js
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		this._status = options.status;
+		this._context = options.context;
+		this._error = options.error;
+
+		Object.preventExtensions(this);
+	}
+
+	get getStatus() {
```
margabit: `getStatus` is very Java oriented. It should just be `status`. Same for other getters

packages/node-cli/src/commands/actionresult/ActionResult.js
```diff
@@ -0,0 +1,42 @@
...
+const ActionResultStatus = require('../actionresult/ActionResultStatus');
+
+module.exports = class ActionResult {
+
+	constructor(options) {
+		assert(options);
+		assert(options.status, "status is required when creating an ActionResult object.");
+
+		if (options.status === ActionResultStatus.SUCCESS) {
+			assert(options.context, "context is required when ActionResult is a success.");
```
margabit: What is context??

packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -83,10 +85,19 @@ module.exports = class CommandActionExecutor {
 				projectConfiguration: projectConfiguration,
 			});
 
+			if (!(actionResult instanceof ActionResult)) {
+				throw 'INTERNAL ERROR: Command must return an ActionResult object.';
+			}
+
+			if (actionResult._status === ActionResultStatus.ERROR) {
+				throw actionResult._error;
```
margabit: if we have a getter, this should just be `actionResult.error`

packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -83,10 +85,19 @@ module.exports = class CommandActionExecutor {
 				projectConfiguration: projectConfiguration,
 			});
 
+			if (!(actionResult instanceof ActionResult)) {
+				throw 'INTERNAL ERROR: Command must return an ActionResult object.';
+			}
+
+			if (actionResult._status === ActionResultStatus.ERROR) {
```
margabit: this should be `actionResult.status`

packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -83,10 +85,19 @@ module.exports = class CommandActionExecutor {
...
+				throw actionResult._error;
+			}
+
 			this._commandOutputHandler.showSuccessResult(actionResult, command.formatOutputFunc);
 
-			if (actionResult && actionResult.operationResult) {
-				const operationResult = actionResult.operationResult;
+			const actionResultContext = actionResult._context;
+			if (actionResultContext.operationResult) {
```
margabit: This assumes all commands return an `operationResult` which I think it's not true. OperationResult is an implementation detail of each command, but that detail should not be reflected in the ActionResult. 
Each command should read the operationResult and determine if based on the internals, the ActionResult should be success or error.

I would suggest to have a helper to transform an operationResult into an actionResult, and be used inside commands so any code outside the command generator does not need to know about SDK or OperationResults. To generate an actionResult from an operationResult, we can do something like:
`SDKOperationResultsUtils.createActionResultFrom(operationResult)`
We already have this utils class.

packages/node-cli/src/commands/actionresult/ActionResult.js
lfpvillegas: Do you refer to?

```
get status() {
		return this._status;
	}
```

or just to put status as a public variable?

packages/node-cli/src/commands/actionresult/ActionResultBuilder.js
lfpvillegas: ActionResultBuilder has been moved to ActionResult

packages/node-cli/src/commands/actionresult/ActionResultStatus.js
lfpvillegas: Done

packages/node-cli/src/commands/actionresult/ActionResult.js
lfpvillegas: Deleted context and added _data, _resultMessages variables to ActionResult,
Created sons of ActionResult for special commands like CreateProject or Deploy

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
```diff
@@ -152,7 +161,8 @@ module.exports = class AddDependenciesCommandGenerator extends BaseCommandGenera
 		return dependenciesString;
 	}
 
-	_formatOutput(operationResult) {
+	_formatOutput(actionResult) {
+		const operationResult = actionResult._context.operationResult;
```
jbbalbas: isn't `_context` a private variable inside ActionResult? If yes, we should access with a getter, like `actionResult.getContext().operationResult`

packages/node-cli/src/core/CommandActionExecutor.js
lfpvillegas: I have started to develop that the commands return actionResult, right now only deploy and create project to advance code review... I'll be working now in the rest of commands.

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
lfpvillegas: Yes, well context would not be part of actionResult anymore... I have developed only deploy and createproject.... add dependencies will be next in the list

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -221,104 +224,115 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
+				//Enclose in double quotes to also support project names with spaces
+				parentdirectory: CommandUtils.quoteString(projectAbsolutePath),
+				type: answers[COMMAND_OPTIONS.TYPE],
+				projectname: SOURCE_FOLDER,
+				...(answers[COMMAND_OPTIONS.OVERWRITE] && { overwrite: '' }),
+				...(answers[COMMAND_OPTIONS.TYPE] === ApplicationConstants.PROJECT_SUITEAPP && {
+					publisherid: answers[COMMAND_OPTIONS.PUBLISHER_ID],
+					projectid: answers[COMMAND_OPTIONS.PROJECT_ID],
+					projectversion: answers[COMMAND_OPTIONS.PROJECT_VERSION],
+				}),
```
jbbalbas: can we construct this in a more-readable way? like:
```
const params = {
  //Enclose in double quotes to also support project names with spaces
  parentdirectory: CommandUtils.quoteString(projectAbsolutePath),
  type: answers[COMMAND_OPTIONS.TYPE],
  projectname: SOURCE_FOLDER
}
if (answers[COMMAND_OPTIONS.OVERWRITE]) { 
  params.overwrite = ''
}
if (answers[COMMAND_OPTIONS.TYPE] === ApplicationConstants.PROJECT_SUITEAPP) {
  params.publisherid: answers[COMMAND_OPTIONS.PUBLISHER_ID],
  params.projectid: answers[COMMAND_OPTIONS.PROJECT_ID],
  params.projectversion: answers[COMMAND_OPTIONS.PROJECT_VERSION]
}
```

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -221,104 +224,115 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
+			};
+
+			this._fileSystemService.createFolder(this._executionPath, projectFolderName);
+
+			const actionCreateProject = new Promise(async (resolve, reject) => {
```
jbbalbas: It would be a good idea to extract this `const actionCreateProject  = new Promise(...)` to another method so this one is cleaner

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
```diff
@@ -69,18 +69,27 @@ module.exports = class AddDependenciesCommandGenerator extends BaseCommandGenera
...
+				action: this._sdkExecutor.execute(executionContext),
+				message: TranslationService.getMessage(MESSAGES.ADDING_DEPENDENCIES),
+			});
+
+			return SDKOperationResultUtils.createActionResultFrom(operationResult);
+
+
+			
+		} catch (error) {
+			return new ActionResult.Builder().withError(error).build();
```
jbbalbas: you can replace `ActionResult.Builder().with...` for `ActionResult.Builder.with...`

jbbalbas: (In all the occurrences in this PR)

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -402,35 +416,35 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
-	_formatOutput(result) {
-		if (!result) {
+	_formatOutput(actionResult) {
+		if (!actionResult) {
 			return;
 		}
-		if (SDKOperationResultUtils.hasErrors(result.operationResult)) {
-			NodeUtils.println(
+		if (SDKActionResultUtils.hasErrors(actionResult)) {
+			logger.println(
```
jbbalbas: i think `logger` was one proposal but you forgot to change it back to `NodeUtils`

packages/node-cli/src/commands/ListFilesCommandGenerator.js
```diff
@@ -66,22 +66,32 @@ module.exports = class ListFilesCommandGenerator extends BaseCommandGenerator {
...
+			};
+			return new ActionResult.Builder().withSuccess(actionResultContext).build();
+		} catch (error) {
+			return new ActionResult.Builder().withError(error).build();
+		}
 	}
 
-	_formatOutput(operationResult) {
+	_formatOutput(actionResult) {
+		const operationResult = actionResult._context.operationResult;
```
jbbalbas: same as before, I think actionResult needs either a `getContext()` or to rename it to simply `context`

packages/node-cli/src/mappers/CreateProjectActionResultMapper.js
```diff
@@ -0,0 +1,29 @@
...
+
+    createActionResultFrom(actionCreateProjectData, projectType, includeUnitTesting) {
+        if (actionCreateProjectData.operationResult.status === ActionResult.SUCCESS) {
+            return CreateProjectActionResult.Builder.withSuccess(
+                actionCreateProjectData.operationResult.data,
+                actionCreateProjectData.operationResult.resultMessage,
+                projectType,
+                actionCreateProjectData.projectDirectory,
+                includeUnitTesting,
+                actionCreateProjectData.npmInstallSuccess
```
jbbalbas: The objective of the Builder patter is exactly to not have this. Instead, we should have something like:
```
CreateProjectActionResult.Builder
  .withData(actionCreateProjectData.operationResult.data)
  .withResultMessage(actionCreateProjectData.operationResult.resultMessage)
  .withType(projectType)
  .fromDirectry(actionCreateProjectData.projectDirectory)
  .withUnitTesting()
  .withSuccessfullNpmInstalled()
```

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
lfpvillegas: yeah good catch!

packages/node-cli/src/mappers/CreateProjectActionResultMapper.js
lfpvillegas: Yeah1 True... I don't know how I can have don this :S

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
lfpvillegas: yeah, I understand what you says... Using "..." is enough to express conditionality but this is already a little bit complex.
So, it's ok for me to change it but since I'm agnostic about javascript code format let's wait for the opinion of the other people.

packages/node-cli/src/commands/DeployCommandGenerator.js
```diff
@@ -136,60 +136,64 @@ module.exports = class DeployCommandGenerator extends BaseCommandGenerator {
...
+				message: TranslationService.getMessage(MESSAGES.DEPLOYING),
+			});
+
+			var isValidate = SDKParams[COMMAND.FLAGS.VALIDATE] ? true : false;
+			var isApplyContentProtection =
+				(this._projectType === PROJECT_SUITEAPP && SDKParams[COMMAND.OPTIONS.APPLY_CONTENT_PROTECTION] === SDK_TRUE)
+					? true
+					: false;
+
+			return DeployActionResultMapper.createActionResultFrom(operationResult, isValidate, isApplyContentProtection);
```
davidecorreu: I think we should have one mapper that is able to create an ActionResult, passing the action name and the operationResult plus some extra parameters. Something, like ActionResultMapper('deploy',operationResult, params). Or ActionResultMapper('deploy')(operationResult, params).

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
lfpvillegas: Extreacted to method createProject

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
```diff
@@ -69,18 +69,27 @@ module.exports = class AddDependenciesCommandGenerator extends BaseCommandGenera
...
+				flags: [COMMAND_OPTIONS.ALL],
+				requiresContextParams: true
+			});
+
+			const operationResult = await executeWithSpinner({
+				action: this._sdkExecutor.execute(executionContext),
+				message: TranslationService.getMessage(MESSAGES.ADDING_DEPENDENCIES),
+			});
+
+			return SDKOperationResultUtils.createActionResultFrom(operationResult);
```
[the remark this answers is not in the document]
lfpvillegas: Adddependencies is not yet submitted, only createproject and deploy... That is the next one

packages/node-cli/src/commands/actionresult/ActionResult.js
```diff
@@ -0,0 +1,86 @@
...
+				this.status = SUCCESS;
+				this.data = data;
+				this.resultMessage = resultMessage;
+				return this;
+			}
+
+			withError(error) {
+				this.status = ERROR;
+				this.error = error;
+				return this;
```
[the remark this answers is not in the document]
lfpvillegas: We can use withResultMessage method. Exceptions use also withError and they don't have resultMessage.

I change the name of resultMessage though

packages/node-cli/src/commands/DeployCommandGenerator.js
lfpvillegas: hmm ok... it could be like a Factory.... I'll try it.

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
lfpvillegas: Added changes AddDependencies... please check. I leave this conversation opened until then

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
lfpvillegas: I have done the changes finally as suggested.

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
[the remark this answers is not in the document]
lfpvillegas: Done

packages/node-cli/src/utils/SDKActionResultUtils.js
```diff
@@ -0,0 +1,43 @@
+/*
```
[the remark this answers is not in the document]
lfpvillegas: Done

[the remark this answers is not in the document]
lfpvillegas: The logger will be refactored in other pull request... anyways the intent to delete SDKOperationResultUtils when this story finishes if possible... or at least delete the methods that are not used in both methods.
I leave the  converstation as opened to remind this.

packages/node-cli/src/core/CommandActionExecutor.js
lfpvillegas: Team decided to just use for the moment builders of ActionResult, since the helper at the end would add a extra layer over ActionResult

packages/node-cli/src/commands/ListFilesCommandGenerator.js
lfpvillegas: Done for all commands now

packages/node-cli/src/utils/SDKActionResultUtils.js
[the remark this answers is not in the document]
lfpvillegas: Finally this would be improved in the other pull request.

packages/node-cli/src/commands/actionresult/DeployActionResult.js
```diff
@@ -0,0 +1,85 @@
...
+
+	validateBuild(build) {
+		assert(build);
+		assert(build.status, "status is required when creating an ActionResult object.");
+		if (build.status === ActionResult.SUCCESS) {
+			assert(build.data, "data is required when ActionResult is a success.");
+		}
+		if (build.status === ActionResult.ERROR) {
+			assert(build.errorMessages, "errorMessages is required when ActionResult is an error.");
+		}
```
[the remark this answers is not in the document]
lfpvillegas: Ok good catch, I would delete this method here and it would use the one of Action Result

I would overrite completely the method in the rest of commands, because some of them don't use "data" (Proxy, SetupAccount), and if do so... they include other properties to check (CreateProject)

MOst of command anyways use the normal actionresult and have data, so it's ok to leave there the assert(build.data)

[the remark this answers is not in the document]
lfpvillegas: Well, thinking twice, I'll try at least to use the super.validateBuild in CreateProjectActionResult

packages/node-cli/src/commands/actionresult/ActionResult.js
```diff
@@ -0,0 +1,95 @@
...
+
+	get data() {
+		return this._data;
+	}
+
+	get resultMessage() {
+		return this._resultMessage;
+	}
+
+	static get Builder() {
```
jbbalbas: we could update it to:

```
class ActionResult {
  constructor(build) { ... }
  ...
  static get Builder() {
    return new ActionResultBuilder();
  }
}

class ActionResultBuilder {
  constructor() { }
  withSuccess() {
    this.status = SUCCESS;
    return this;
  }
  ...
  build() { ...}
}
module.exports.ActionResult = ActionResult;
module.exports.ActionResultBuilder = ActionResultBuilder;
```

So that, in the subActionResult that extend ActionResult we can extend as well the ActionResultBuilder from here, like:
```
class CreateProjectActionResult extends ActionResult {
  constructor(build) {
    super(build)
    ...
  }
  ...
  static get Builder() {
    return new CreateProjectActionResultBuilder();
  }
}

class CreateProjectActionResultBuilder extends ActionResultBuilder {
  withProjectType(projectType) {
    this.projectType = projectType;
    return this;
  }
  build() {...}
  ...
}
```

And you don't need to duplicate all the time the methods .withSuccess, withResultMessage, withData, withError, etc

packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -83,20 +83,32 @@ module.exports = class CommandActionExecutor {
...
-					commandUserExtension.onError(error);
+			if (actionResult.status === ActionResult.ERROR) {
+				if (actionResult.errorMessages && Array.isArray(actionResult.errorMessages) && actionResult.size > 0) {
+					throw actionResult.errorMessages;
+				}
+				else if (actionResult.resultMessage) {
+					throw actionResult.resultMessage;
+				}
+				else {
+					throw 'Unexpected Error' //TODO Code Review and Change message
```
jbbalbas: now that we always build the actionResult in the same class, we could homogenize this and have always the errors in actionResult.errorMessages, no? so that we only need to do
```
if (actionResult.status === ActionResult.ERROR) {
  throw actionResult.errorMessages;
}
```

packages/node-cli/src/core/CommandOutputHandler.js
```diff
@@ -6,9 +6,11 @@
 
 const NodeUtils = require('./../utils/NodeUtils');
 const { unwrapExceptionMessage, unwrapInformationMessage } = require('./../utils/ExceptionUtils');
+const ActionResult = require('../commands/actionresult/ActionResult');
 
 module.exports = class CommandOutputHandler {
 	showSuccessResult(actionResult, formatOutputFunction) {
+		// TODO assert(actionResult instanceof ActionResult);
```
jbbalbas: Since all the results are build now with ActionResult.Builder, I think this assert isn't neccessary. Also, you don't need the require() in line 9

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
```diff
@@ -69,18 +70,33 @@ module.exports = class AddDependenciesCommandGenerator extends BaseCommandGenera
...
+			});
+
+			const operationResult = await executeWithSpinner({
+				action: this._sdkExecutor.execute(executionContext),
+				message: TranslationService.getMessage(MESSAGES.ADDING_DEPENDENCIES),
+			});
+
+			return operationResult.status === ActionResult.SUCCESS
+			? ActionResult.Builder
+				.withSuccess()
```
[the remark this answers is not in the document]
lfpvillegas: Yes, however, proxy and setup doesn't use data, so it's a must to have withSuccess (or just success) in ActionResult.

I would use it everywhere to homogenize the call in all commands. Also the word data not specifies at all that this has been successful

As an alternative I can rename "withData' to "successWithData"

packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -83,20 +83,32 @@ module.exports = class CommandActionExecutor {
...
-				const operationResult = actionResult.operationResult;
-				if (operationResult.status === OperationResultStatus.SUCCESS && commandUserExtension.onCompleted) {
-					commandUserExtension.onCompleted(actionResult);
-				} else if (operationResult.status === OperationResultStatus.ERROR && commandUserExtension.onError) {
-					const error = SDKOperationResultUtils.getResultMessage(operationResult)
-						+ NodeUtils.lineBreak
-						+ SDKOperationResultUtils.getErrorMessagesString(operationResult);
-					commandUserExtension.onError(error);
+			if (actionResult.status === ActionResult.ERROR) {
+				if (actionResult.errorMessages && Array.isArray(actionResult.errorMessages) && actionResult.size > 0) {
```
davidecorreu: I think last condition should be actionResult.errorMessages.length>0.
Anyway we have a method that is taking care of getting the errorMessages including the resultMessage.
So we can:
```
if (actionResult.status === ActionResult.ERROR) {
   throw ActionResultUtils.getErrorMessagesString(actionResult);
}
```

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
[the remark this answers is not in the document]
jbbalbas: Then, I think its better to have `success().withData()` than the alternatives (`succesWithData()` or `withSuccess().withData()`

packages/node-cli/src/utils/ActionResultUtils.js
```diff
@@ -0,0 +1,43 @@
...
+** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+*/
+'use strict';
+
+const NodeUtils = require('./NodeUtils');
+const { ERROR } = require('../commands/actionresult/ActionResult');
+
+module.exports = {
+	getErrorMessagesString: actionResult => {
+		const { error, resultMessage } = actionResult;
```
davidecorreu: Error has been renamed to errorMessages!!!!

packages/node-cli/src/commands/actionresult/CreateProjectActionResult.js
```diff
@@ -0,0 +1,109 @@
...
+	static get Builder() {
+		return new class Builder {
+			constructor() { }
+
+			withSuccess() {
+				this.status = ActionResult.SUCCESS;
+				return this;
+			}
+
+			withError(errorMessages) {
```
davidecorreu: This is a little bit tricky. We name it withError but store errorMessages. Maybe we should rename it.

packages/node-cli/src/commands/DeployCommandGenerator.js
```diff
@@ -136,60 +137,75 @@ module.exports = class DeployCommandGenerator extends BaseCommandGenerator {
...
+				message: TranslationService.getMessage(MESSAGES.DEPLOYING),
+			});
+
+			var isServerValidation = SDKParams[COMMAND.FLAGS.VALIDATE] ? true : false;
+			var isApplyContentProtection =
+				(this._projectType === PROJECT_SUITEAPP && SDKParams[COMMAND.OPTIONS.APPLY_CONTENT_PROTECTION] === SDK_TRUE)
+					? true
+					: false;
+
+			return operationResult.status == ActionResult.SUCCESS
```
davidecorreu: We should not use ActionResult.SUCCESS in this comparison, because we are checking a OperationResult.status, that actually have the same value, but they are different concepts.
The same happens in the other command generators.

packages/node-cli/src/commands/actionresult/CreateProjectActionResult.js
```diff
@@ -0,0 +1,109 @@
+/*
+** Copyright (c) 2019 Oracle and/or its affiliates.  All rights reserved.
+** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+*/
+'use strict';
+const assert = require('assert');
+const ActionResult = require('./ActionResult');
+
+class CreateProjectActionResult extends ActionResult {
```
davidecorreu: We should rename this one to ProjectActionResult to follow the same approach as in the other "ActionsResults".

packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -83,20 +83,32 @@ module.exports = class CommandActionExecutor {
...
+			this._commandOutputHandler.showSuccessResult(actionResult, command.formatOutputFunc);
+
+			if (actionResult.status === ActionResult.SUCCESS && commandUserExtension.onCompleted) {
+				commandUserExtension.onCompleted(actionResult);
+			} else if (actionResult.status === ActionResult.ERROR && commandUserExtension.onError) {
+				const error = ActionResultUtils.getResultMessage(actionResult)
+					+ NodeUtils.lineBreak
+					+ ActionResultUtils.getErrorMessagesString(actionResult);
+				commandUserExtension.onError(error);
+			}
```
davidecorreu: I think this block will never be reached after the changes you have done.

packages/node-cli/src/core/CommandActionExecutor.js
lfpvillegas: There is at least one command that have the errorMessage empty and the errorMessage is in resultMessage.

This is a problem in JavaCLI not in Nodejs... I don't want to touch JavaCLI since the story is already too big so I would let it as it is

lfpvillegas: (and I have only found one case... I haven't tried edge cases) I will put a //TODO. I know this is not the best way of course, but we should study all JavaCLI cases

packages/node-cli/src/core/CommandActionExecutor.js
lfpvillegas: Nice!

packages/node-cli/src/core/CommandActionExecutor.js
lfpvillegas: ok just forget about it... there is a comment a David that resolves this.

I close the conversation

packages/node-cli/src/commands/actionresult/CreateProjectActionResult.js
lfpvillegas: we would rename it to error(errorMessages) this way is compared with the other renaming 
withSuccess() to success()

packages/node-cli/src/core/CommandActionExecutor.js
lfpvillegas: There could be an exception when taking commandmetadata or the projectfolder

I would leave it

packages/node-cli/src/commands/actionresult/ActionResult.js
davidecorreu: I have another similar approach to embed the Builder inside the ActionResult.js
```
class  ActionResult {
	...
	static BuilderClass = class Builder {

		constructor() { }
 
		withSuccess() {
			this.status = SUCCESS;
			return  this;
		}
		...
	}

	static  get  Builder() {
		return  new  this.BuilderClass();
	}
}
```

And then from 
```
class DeployActionResult extends ActionResult {
	...
	static  BuilderClass = class  DeployBuilder  extends  ActionResult.BuilderClass {
		isServerValidation(isServerValidation) {
			this.isServerValidation = isServerValidation;
			return  this;
		}
		appliedContentProtection(appliedContentProtection) {
			this.appliedContentProtection = appliedContentProtection;
			return  this;
		}
	build() {
		return  new  DeployActionResult({
			status:  this.status,
			...(this.data && { data:  this.data }),
			...(this.resultMessage && { resultMessage:  this.resultMessage }),
			...(this.errorMessages && { errorMessages:  this.errorMessages }),
			...(this.isServerValidation && { isServerValidation:  this.isServerValidation }),
			...(this.appliedContentProtection && { appliedContentProtection: this.appliedContentProtection })
		});
	}
}
```

lfpvillegas: I did first approach because I didn't see the second. So, what we do at the end?

packages/node-cli/src/commands/DeployCommandGenerator.js
lfpvillegas: ok, I put it in SDKOperationResultUtils for now

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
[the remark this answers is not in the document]
lfpvillegas: ok nice I implement as this

packages/node-cli/src/commands/actionresult/ActionResult.js
davidecorreu: First option is good!

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
```diff
@@ -69,18 +71,33 @@ module.exports = class AddDependenciesCommandGenerator extends BaseCommandGenera
...
+
+			const operationResult = await executeWithSpinner({
+				action: this._sdkExecutor.execute(executionContext),
+				message: TranslationService.getMessage(MESSAGES.ADDING_DEPENDENCIES),
+			});
+
+			return operationResult.status === SDKOperationResultUtils.SUCCESS
+				? ActionResult.Builder
+					.success()
+					.withData(operationResult.data)
```
vic-nik: Can withData function set successful status so we don't need to call success() additionally? Is there any case withData will need to be combined with error status?

packages/node-cli/src/commands/AddDependenciesCommandGenerator.js
```diff
@@ -152,15 +169,16 @@ module.exports = class AddDependenciesCommandGenerator extends BaseCommandGenera
 		return dependenciesString;
 	}
 
-	_formatOutput(operationResult) {
-		if (SDKOperationResultUtils.hasErrors(operationResult)) {
-			SDKOperationResultUtils.logResultMessage(operationResult);
-			SDKOperationResultUtils.logErrors(operationResult);
+	_formatOutput(actionResult) {
+		if (actionResult.error) {
+			if (actionResult.resultMessage) {
```
vic-nik: Why do we read resultMessage if the status is error? I think all errors should be in error messages in this case

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -219,36 +221,72 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
-				projectversion: answers[COMMAND_OPTIONS.PROJECT_VERSION],
-			}),
-		};
+			const params = {
+				//Enclose in double quotes to also support project names with spaces
+				parentdirectory: CommandUtils.quoteString(projectAbsolutePath),
+				type: answers[COMMAND_OPTIONS.TYPE],
+				projectname: SOURCE_FOLDER
+			}
+			if (answers[COMMAND_OPTIONS.OVERWRITE]) {
```
vic-nik: Why this change? It looked more consice and was quite readable originally

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -219,36 +221,72 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
-			type: answers[COMMAND_OPTIONS.TYPE],
-			projectname: SOURCE_FOLDER,
-			...(answers[COMMAND_OPTIONS.OVERWRITE] && { overwrite: '' }),
-			...(answers[COMMAND_OPTIONS.TYPE] === ApplicationConstants.PROJECT_SUITEAPP && {
-				publisherid: answers[COMMAND_OPTIONS.PUBLISHER_ID],
-				projectid: answers[COMMAND_OPTIONS.PROJECT_ID],
-				projectversion: answers[COMMAND_OPTIONS.PROJECT_VERSION],
-			}),
-		};
+			const params = {
```
vic-nik: Why this change? It looked more consice and was quite readable originally

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -219,36 +221,72 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
 
-		const actionCreateProject = new Promise(async (resolve, reject) => {
+			return actionCreateProjectData.operationResult.status === SDKOperationResultUtils.SUCCESS
+				? CreateProjectActionResult.Builder
+					.success()
+					.withData(actionCreateProjectData.operationResult.data)
+					.withResultMessage(actionCreateProjectData.operationResult.resultMessage)
+					.withProjectType(projectType)
+					.withProjectName(projectName)
+					.fromDirectory(actionCreateProjectData.projectDirectory)
```
vic-nik: Why fromDirectory and not withDirectory?

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -219,36 +221,72 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
+			if (answers[COMMAND_OPTIONS.TYPE] === ApplicationConstants.PROJECT_SUITEAPP) {
+				params.publisherid = answers[COMMAND_OPTIONS.PUBLISHER_ID];
+				params.projectid = answers[COMMAND_OPTIONS.PROJECT_ID];
+				params.projectversion = answers[COMMAND_OPTIONS.PROJECT_VERSION];
+
+			}
+
+			this._fileSystemService.createFolder(this._executionPath, projectFolderName);
+
+			const actionCreateProject = new Promise(this.createProject(
```
vic-nik: I would rename to createProjectAction

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -219,36 +221,72 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
+				params, answers, projectAbsolutePath, projectFolderName, manifestFilePath));
+
+			const actionCreateProjectData = await actionCreateProject;
```
vic-nik: I would rename to createProjectActionData

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -219,36 +221,72 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
+				projectname: SOURCE_FOLDER
+			}
+			if (answers[COMMAND_OPTIONS.OVERWRITE]) {
+				params.overwrite = '';
+			}
+			if (answers[COMMAND_OPTIONS.TYPE] === ApplicationConstants.PROJECT_SUITEAPP) {
+				params.publisherid = answers[COMMAND_OPTIONS.PUBLISHER_ID];
+				params.projectid = answers[COMMAND_OPTIONS.PROJECT_ID];
+				params.projectversion = answers[COMMAND_OPTIONS.PROJECT_VERSION];
+
```
vic-nik: extra line

packages/node-cli/src/commands/ImportFilesCommandGenerator.js
```diff
@@ -177,35 +179,49 @@ module.exports = class ImportFilesCommandGenerator extends BaseCommandGenerator
...
+		try {
+			if (this._projectInfoService.getProjectType() === PROJECT_SUITEAPP) {
+				throw TranslationService.getMessage(ERRORS.IS_SUITEAPP);
+			}
 
-		return executeWithSpinner({
-			action: this._sdkExecutor.execute(executionContextImportObjects),
-			message: TranslationService.getMessage(MESSAGES.IMPORTING_FILES),
-		});
+			const executionContextImportObjects = new SDKExecutionContext({
```
vic-nik: variable names should be nouns, not verbs (the verbs are for functions/methods). Can we rename to importObjectsExecutionContext?

packages/node-cli/src/commands/ImportFilesCommandGenerator.js
```diff
@@ -177,35 +179,49 @@ module.exports = class ImportFilesCommandGenerator extends BaseCommandGenerator
...
 	}
 
-	_formatOutput(operationResult) {
-		const { data } = operationResult;
+	_formatOutput(actionResult) {
 
-		if (SDKOperationResultUtils.hasErrors(operationResult)) {
-			SDKOperationResultUtils.logResultMessage(operationResult);
-			SDKOperationResultUtils.logErrors(operationResult);
+		if (ActionResultUtils.hasErrors(actionResult)) {
```
vic-nik: Why is hasErrors function not used in previous instances?

packages/node-cli/src/commands/ImportObjectsCommandGenerator.js
```diff
@@ -305,24 +307,25 @@ module.exports = class ImportObjectsCommandGenerator extends BaseCommandGenerato
...
+	async _executeAction(answers) {
+		try {
+			if (answers[ANSWERS_NAMES.OVERWRITE_OBJECTS] === false) {
+				throw TranslationService.getMessage(MESSAGES.CANCEL_IMPORT);
 			}
-		} else {
-			if (answers[COMMAND_FLAGS.EXCLUDE_FILES]) {
-				flags.push(COMMAND_FLAGS.EXCLUDE_FILES);
-				delete answers[COMMAND_FLAGS.EXCLUDE_FILES];
+
```
vic-nik: empty line to be removed

packages/node-cli/src/commands/actionresult/ActionResult.js
```diff
@@ -0,0 +1,100 @@
...
+*/
+'use strict';
+const assert = require('assert');
+
+const ERROR = "ERROR";
+const SUCCESS = "SUCCESS";
+
+class ActionResult {
+
+	constructor(build) {
```
vic-nik: what is build? Isn't it operationResult?

vic-nik: Why build? Let's call it parameters

packages/node-cli/src/commands/actionresult/ActionResult.js
```diff
@@ -0,0 +1,100 @@
...
+		this.validateBuild(build);
+
+		this._status = build.status;
+		this._data = build.data;
+		this._resultMessage = build.resultMessage;
+		this._errorMessages = build.errorMessages;
+	}
+
+	validateBuild(build) {
```
vic-nik: let's rename to validateParameters(parameters)

packages/node-cli/src/commands/actionresult/ActionResult.js
```diff
@@ -0,0 +1,100 @@
...
+		assert(build);
+		assert(build.status, "status is required when creating an ActionResult object.");
+		if (build.status === SUCCESS) {
+			assert(build.data, "data is required when ActionResult is a success.");
```
vic-nik: That means withData should set successful status by default and there's no need for success() function

packages/node-cli/src/commands/actionresult/ActionResult.js
```diff
@@ -0,0 +1,100 @@
...
+
+	static get ERROR() {
+		return ERROR;
+	}
+};
+
+
+class ActionResultBuilder {
+	constructor() { }
+
```
vic-nik: empty line to remove

packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -83,20 +83,20 @@ module.exports = class CommandActionExecutor {
...
-				const operationResult = actionResult.operationResult;
-				if (operationResult.status === OperationResultStatus.SUCCESS && commandUserExtension.onCompleted) {
-					commandUserExtension.onCompleted(actionResult);
-				} else if (operationResult.status === OperationResultStatus.ERROR && commandUserExtension.onError) {
-					const error = SDKOperationResultUtils.getResultMessage(operationResult)
-						+ NodeUtils.lineBreak
-						+ SDKOperationResultUtils.getErrorMessagesString(operationResult);
-					commandUserExtension.onError(error);
-				}
+			if (actionResult.status === ActionResult.ERROR) {
```
vic-nik: Why don't we use hasErrors consistently?

PR #52, opened 2020-03-02
Adding Dockerfile with suitecloud

jbbalbas:


Dockerfile
```diff
@@ -0,0 +1,132 @@
+FROM node:12.14
+
+## ----------------------------------------------------------------------------
+## --- INSTALLING JAVA 11 -----------------------------------------------------
+## https://github.com/docker-library/openjdk/blob/master/11/jdk/Dockerfile
+## ----------------------------------------------------------------------------
```
jbbalbas: Installing JAVA 11 from openjdk Dockerfile. Everything is copied from their repository

Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+  javac --version; \
+  java --version
+
+## ----------------------------------------------------------------------------
+## --- JAVA INSTALLATION FINISHED ---------------------------------------------
+## ----------------------------------------------------------------------------
+
+## ----------------------------------------------------------------------------
+## --- INSTALLING SUITECLOUD (Nodejs CLI) -------------------------------------
+## ----------------------------------------------------------------------------
```
jbbalbas: The real work starts here. We add suitecloud and sdfcli to the docker image, since at this point it has Node12.14 and java 11

margabit: Why do we install sdfcli?

jbbalbas: I've added it because it's a requirement of the user story PDPDEVTOOL-2660

vic-nik: @margabit We had a discussion about whether to install both CLIs and dicided to have two in one container. At this point Java CLI provides a bit more commands so maybe it's good to have both of them in the container for CI. I think we may remove sdfcli later. Do you have concerns about having both installed?

Dockerfile
```diff
@@ -0,0 +1,132 @@
...
+
+RUN npm install -g @oracle/netsuite-suitecloud-nodejs-cli --unsafe-perm;
```
vic-nik: Let's change it to @oracle/suitecloud-cli already, I will change the package name today

Dockerfile
margabit: Ok, sounds good!

PR #54, opened 2020-03-03
Change NPM package names and fix JS warnings

vic-nik:


extensions/localserver-cli-command/src/client-scripts/LoadTemplateSafe.js
```diff
@@ -28,7 +28,7 @@
 	});
 
 	function copyProperties(source, dest) {
-		for (var property in source) {
+		for (const property in source) {
```
davidecorreu: Is it going to be alright to change from var to const in this client-script? Is my only concern in the whole pull request. Just because maybe this code is run in browser.

vic-nik: Good catch, I'll revert it!

PR #66, opened 2020-03-11
Fix bug by creating new method emptyFolderRecursive with exceptionPath

lfpvillegas:


packages/node-cli/src/services/FileSystemService.js
```diff
@@ -110,6 +110,25 @@ module.exports = class FileSystemService {
 		}
 	}
 
+	emptyFolderRecursive(folderPath, exceptionPath) {
```
davidecorreu: I will name the second parameter, excludePath.

lfpvillegas: Done!

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -252,6 +252,9 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
 
 		const actionCreateProject = new Promise(async (resolve, reject) => {
 			try {
+				if(answers[COMMAND_OPTIONS.OVERWRITE]) {
```
vic-nik: formatting, space missing after if

packages/node-cli/src/services/FileSystemService.js
```diff
@@ -110,6 +110,23 @@ module.exports = class FileSystemService {
...
 	}
 
+	emptyFolderRecursive(folderPath) {
+		assert(folderPath);
+		let self = this;
+		if (existsSync(folderPath)) {
+			readdirSync(folderPath).forEach(file => {
+				let currentPath = path.join(folderPath, file);
+				if (lstatSync(currentPath).isDirectory()) {
+					// recurse
```
vic-nik: I would remove the comments, this code is readable

PR #67, opened 2020-03-11
Bugfix/pdpdevtool 2836

jbbalbas:


packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -363,6 +338,30 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
+
+	_sanitizeName(name) {
+		let cleanName = name.toLowerCase().replace(/[^a-z._\@\-\/\s]/g, '').replace(/\s/g, '_');
+		if (cleanName.indexOf('/') >= 0) {
+			return '@' + 
+				this._sanitizeString(cleanName.slice(0, cleanName.indexOf('/'))) + 
+				'/' + 
+				this._sanitizeString(cleanName.slice(cleanName.indexOf('/') + 1, cleanName.length));
+		}
+		else {
```
vic-nik: Formatting, else should be on the same line as closing brace

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -363,6 +338,30 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
 		);
 	}
 
+
+	_sanitizeName(name) {
+		let cleanName = name.toLowerCase().replace(/[^a-z._\@\-\/\s]/g, '').replace(/\s/g, '_');
+		if (cleanName.indexOf('/') >= 0) {
```
vic-nik: Since cleanName.indexIf('/') is used 3 times it would be better to extract it to variable (const)

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -363,6 +338,30 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
...
+			return '@' +
```
vic-nik: Let's use template literal instead of string concatenation

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -363,6 +338,28 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
 		);
 	}
 
+	_sanitizeName(name) {
```
vic-nik: Confirmed with Carlos. Let's use static name in template file. It's going to be "suitecloud-project", no need to differentiate between SuiteApp and ACP in the name

PR #77, opened 2020-03-18
sc localserver regression fixes

PepeBotella25:


extensions/localserver-cli-command/src/compilers/Compiler.js
```diff
@@ -46,6 +46,11 @@ module.exports = class Compiler {
 
 	_createRequireJSFile(localFolder) {
 		const src = require.resolve('requirejs');
-		fs.copyFileSync(src, path.join(localFolder, 'require.js'));
+		const dst = path.join(localFolder, 'require.js');
+		fs.copyFileSync(src, dst);
+
+		let content = fs.readFileSync(dst).toString();
+		content = content.replace('#!/usr/bin/env node', '');
+		fs.writeFileSync(dst, content);
```
jbbalbas: Instead copying the file from src to dst, then reading it from dst, remove the `#!/usr/bin/env node` string and then write again in dst it, wouldn't it be more efficient to read it from src, remove the unwanted string and write it in dst?

extensions/localserver-cli-command/src/DeployXml.js
```diff
@@ -16,12 +16,20 @@ module.exports = class DeployXml {
...
+		const isACP = manifest.$.projecttype === 'ACCOUNTCUSTOMIZATION';
+
 		let objectsPath = this.deploy.objects.path;
 		objectsPath = path.join(this._projectFolder, objectsPath.substr(1));
 		this.objectsPath = path.dirname(objectsPath);
 
 		let filesPath = this.deploy.files.path;
 		filesPath = path.join(this._projectFolder, filesPath.substr(1));
+
+		if(isACP){
```
vic-nik: Formatting, spaces are missing after if and before curly brace

extensions/localserver-cli-command/src/compilers/Compiler.js
vic-nik: +1

extensions/localserver-cli-command/src/DeployXml.js
```diff
@@ -16,12 +16,20 @@ module.exports = class DeployXml {
 
 		this.deploy = Utils.parseXml(this._projectFolder, 'deploy.xml').deploy;
 
+		const { manifest } = Utils.parseXml(this._projectFolder, 'manifest.xml');
```
vic-nik: We have constants for manifest.xml and ACCONTCUSTOMIZATION in ApplicationConstants.js
I would use those here and we can add another constrant for 'deploy.xml' under FILES

PR #78, opened 2020-03-19
Added SuiteCloud Unit Testing examples

x8a:


packages/unit-testing/README.md
```diff
@@ -32,14 +32,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+							> SuiteScripts
+                  > Suitelet.js
+							> Templates
+							> Web Site Hosting Files
+					> Objects
+					> Translations
+					deploy.xml
+					manifest.xml
+		suitecloud.config.js
+		jest.config.js
```
jbbalbas: You're mixing spaces and tabs here, and the result isn't aligned:
image
(Probably two/three tabs is also too much spacing)

packages/unit-testing/README.md
```diff
@@ -32,14 +32,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+					> AccountConfiguration
+					> FileCabinet
+							> SuiteScripts
+							> Templates
+							> Web Site Hosting Files
+					> Objects
+					> Translations
+					deploy.xml
+					manifest.xml
+		suitecloud.config.js
```
jbbalbas: same here, two/three spaces seems too much spacing

packages/unit-testing/README.md
vic-nik: Agree, one tab or max of 4 spaces should be enough

packages/unit-testing/README.md
```diff
@@ -32,14 +32,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+```
+
+To see the content of the files you need to make SuiteCloud Unit Testing work, check the following list:
+
+- `jest.config.js` file
+```javascript
+const SuiteCloudJestConfiguration = require("@oracle/suitecloud-unit-testing/jest-configuration/SuiteCloudJestConfiguration");
+
+module.exports = SuiteCloudJestConfiguration.build({
+		projectFolder: 'src',
```
vic-nik: Seems like a bit too much of spaces here

packages/unit-testing/README.md
```diff
@@ -11,7 +11,7 @@ The following software is required to use SuiteCloud Unit Testing package:
 ## `jest.config.js` File Examples
```
vic-nik: In Prerequisites section of this file it says:
The following software is required to use SuiteCloud Unit Testing package: Node.js, Jest, _tests_ folder and jest.config.js file.
I think jest.config.js file doesn't belong to this section - it is configuration and not software
 _tests_ folder is not required but is the default location for placing test files, I think it can be overriten in config so I would only add a suggestion to create _tests_ folder according to Jest conventions

x8a: As I understood, you need the **jest.config.js** file in order to configure your jest config, so that's why I added it here. However, since the file and the folder are not software, I can create a second list with required configuration and add those two there. What do you think?

Also, the **__tests__** folder is added when running `project:create` and adding unit testing, so I think it's good to add this as our config structure.

packages/unit-testing/README.md
```diff
@@ -5,13 +5,15 @@ Suitecloud Unit Testing allows you to use unit testing with [Jest](https://jestj
 The following software is required to use SuiteCloud Unit Testing package:
 - Node.js version 12.14.0 LTS or greater
 - [Jest](https://jestjs.io/docs/en/getting-started.html)
-- A `__tests__` folder in the root of your SDF project.
-- A `jest.config.js` file in the root of your SDF project.
+
+Also, to properly structure your test files, create the following elements in the root of your SDF project:
```
vic-nik: Great! I would also change 'to properly structure your test files' to smth. like 'to properly structure and run your tests'

packages/unit-testing/README.md
```diff
@@ -32,14 +34,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+
+Here you can find two examples on how to use SuiteCloud Unit Testing with a SuiteCloud project.
+
+The first example covers testing for the **N/record** module, which is fully mocked in SuiteCloud Unit Testing. Whereas the second example covers the testing of a module that is not mocked in SuiteCloud Unit Testing, by using a custom stub.
+
+>💡 You can manually mock any module that is still not supported in SuiteCloud Unit Testing.
+
+### N/record Module Example
+
+Make sure your project follows a structure similar to the one presented below:
```
vic-nik: I am not sure about this sentence, since this is an example and not a guide maybe it's better to say smth. like this instead:
Given the project structure similar to the one presented below

packages/unit-testing/README.md
```diff
@@ -32,14 +34,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+		> Web Site Hosting Files
+	> Objects
+	> Translations
+  deploy.xml
+  manifest.xml
+suitecloud.config.js
+jest.config.js
+```
+
+To see the content of the files you need to make SuiteCloud Unit Testing work, check the following list:
```
vic-nik: I think this sentence is a bit complicated I couldn't understand it right away, also it's not true that the user needs all these file in order for unit testing to work, only config files. Maybe smth. more simple like:
'Here's the content of the files in this example'

packages/unit-testing/README.md
```diff
@@ -32,14 +34,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+    });
+});
+
+```
+
+To read a comprehensive description of this example, visit [NetSuite's Help Center](https://system.netsuite.com/app/help/helpcenter.nl?fid=section_158436693543.html#subsect_158437822806).
+
+### Custom Stub Example 
+
+In order for your tests to fully work, your project needs to follow a specific structure. Use the following structure as an example:
```
vic-nik: Also not sure, we don't need to eforce the structure, it's only example. E.g. customStubs can be named differently or even ___tests___ location can be reconfigured if needed. I think we can simply remove the first sentence.

packages/unit-testing/README.md
```diff
@@ -32,14 +34,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+		> Web Site Hosting Files
+	> Objects
+	> Translations
+	deploy.xml
+	manifest.xml
+suitecloud.config.js
+jest.config.js
+```
+
+Once you have a file structure similar to the one above, you can start creating the testing files.
```
vic-nik: I would remove this line too, the test file is already in the structure above: http.test.js

packages/unit-testing/README.md
```diff
@@ -32,14 +34,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+	> Translations
+	deploy.xml
+	manifest.xml
+suitecloud.config.js
+jest.config.js
+```
+
+Once you have a file structure similar to the one above, you can start creating the testing files. 
+
+To see the content of the files you need to make SuiteCloud Unit Testing work, check the following list:
```
vic-nik: Same as previous comment on similar line

packages/unit-testing/README.md
```diff
@@ -32,14 +34,243 @@ module.exports = SuiteCloudJestConfiguration.build({
...
+		customStubs: [
+			{
+				module: "N/http",
+				path: "&lt;rootDir>/customStubs/http.js"
+			}
+		]
+});
+```
+- `http.js` file: This is the stub file. It partially mocks NetSuite's **N/http** module.
+>💡 The JSDoc annotations are copied from NetSuite's **N/http** module. Comment out the ones you do not need, or remove them completely.
```
vic-nik: I think the second sentence is not needed, it's not possible to comment annotations. The first sentence describes it quite well

packages/unit-testing/README.md
x8a: I added the same introductory sentence that the other example has.

packages/unit-testing/README.md
```diff
@@ -1,45 +1,314 @@
...
-The `jest.config.js` file must follow a specific structure. Depending on your SDF project type, check one of the following examples:
+## Getting Started
+> ⚠ SuiteCloud Unit Testing is installed as a `devDependency`.
+
+To get started with SuiteCloud Unit Testing, do the following:
 
-- For ACP:
+1. Inside of your SDF project folder, create a `src` folder.
+2. Move your project files inside of the `src` folder.
+3. To initialize NPM, from the root of your SDF project folder, run `npm init`.
```
vic-nik: I think we should say "To initialize NPM package"

packages/unit-testing/README.md
```diff
@@ -1,45 +1,314 @@
...
+        const a = 'foobar';
+        const b = 'foobar';
+        expect(a).toMatch(b);
+    });
+});
+```
+8. From the root of your SDF project folder, run `npm test`
+to run your test. You should see an output similar to following:
+```
+PASS  __tests__/BasicJest.test.js
```
vic-nik: Should be:
PASS  __tests__/sample-test.js

packages/unit-testing/README.md
```diff
@@ -1,45 +1,314 @@
...
+    √ should assert stings are equal (2ms)
 ```
+>💡 If you want to run tests with coverage, from the root of your SDF project folder, run `npm test --coverage`.
+
+**You successfully ran your first test for an SDF project!**
+
+➡ If you use SuiteCloud CLI for Node.js, you can install SuiteCloud Unit Testing when running the [`project:create`](../../packages/node-cli/README.md/##Commands) command by following the questions prompted.
+
+## Additional Configuration
+To properly structure and run your test files, create a `jest.config.js` file, inside of the root of your SDF project folder.
```
vic-nik: I would change to:
To properly run your tests against SuiteScript 2.X files, create a jest.config.js file, inside of the root of your SDF project folder.

PR #79, opened 2020-03-23
Feature/pdpdevtool 2802 - Support for file upload in Node CLI

roctaviano:
Feature/pdpdevtool 2802 - Support for file upload in Node CLI


packages/node-cli/src/services/FileSystemService.js
```diff
@@ -20,7 +20,20 @@ module.exports = class FileSystemService {
 
 		const availableDirectories = getDirectories(parentFolder);
 
-		return [parentFolder, ...availableDirectories];
+		return availableDirectories;
```
davidecorreu: Actually now this function does exactly what it says, return the directories inside of the parentFolder, but it is used also in the ImportObjectsCommandGenerator.js , and there is expecting the parantFolder in the returned array. One possible fix can be adding there (ImportObjectsCommandGenerator.js) the '/Objects' folder as one of options to pick.

packages/node-cli/src/validation/FileCabinetValidator.js
```diff
@@ -0,0 +1,24 @@
...
+const TEMPLATES_PATH = "/Templates";
+const TEMPLATES_EMAIL_TEMPLATES_PATH = "/Templates/E-mail Templates";
+const TEMPLATES_MARKETING_TEMPLATES_PATH = "/Templates/Marketing Templates";
+const WEB_SITE_HOSTING_FILES_PATH = "/Web Site Hosting Files";
+
+const UNRESTRICTED_PATHS = [SUITESCRIPTS_PATH, TEMPLATES_EMAIL_TEMPLATES_PATH, TEMPLATES_MARKETING_TEMPLATES_PATH, WEB_SITE_HOSTING_FILES_PATH];
+
+class FileCabinetValidator {
+	pathIsUnrestricted(path) {
+		//Templates folder is a special case, because it contains both restricted and unrestricted folders,
```
davidecorreu: Can we upload a file to "/Templates" ?

margabit:
Is this going to be published with the 20.1 release?

packages/node-cli/src/validation/FileCabinetValidator.js
alii-gator: Not directly under Templates..

alii-gator:
Albert this was not part of Regression, so it will not be released in 1.0.0 version

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
```diff
@@ -0,0 +1,188 @@
...
+	SUCCESS: 'SUCCESS',
+	ERROR: 'ERROR',
+};
+
+const { validateArrayIsNotEmpty, showValidationResults } = require('../validation/InteractiveAnswersValidator');
+
+module.exports = class UploadFilesCommandGenerator extends BaseCommandGenerator {
+	constructor(options) {
+		super(options);
+		this._projectInfoService = new ProjectInfoService(this._projectFolder);
```
karl-anthony-ng: Is projectInfoService used in here?

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
```diff
@@ -0,0 +1,188 @@
...
+		if (SDKOperationResultUtils.hasErrors(operationResult)) {
+			SDKOperationResultUtils.logResultMessage(operationResult);
+			SDKOperationResultUtils.logErrors(operationResult);
+			return;
+		}
+
+		if (Array.isArray(data)) {
+			const successfulUploads = data.filter(result => result.type === UPLOAD_FILE_RESULT_STATUS.SUCCESS);
+			const unsuccessfulUploads = data.filter(result => result.type === UPLOAD_FILE_RESULT_STATUS.ERROR);
+			if (Array.isArray(successfulUploads) && successfulUploads.length) {
```
karl-anthony-ng: I think filter always returns an array, so no need to if it's an array here.

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
```diff
@@ -0,0 +1,188 @@
...
+				NodeUtils.println(TranslationService.getMessage(OUTPUT.FILES_UPLOADED), NodeUtils.COLORS.RESULT);
+				successfulUploads.forEach(result => {
+					NodeUtils.println(this._getFileCabinetRelativePath(result.file.path), NodeUtils.COLORS.RESULT);
+				});
+			}
+			if (Array.isArray(unsuccessfulUploads) && unsuccessfulUploads.length) {
```
karl-anthony-ng: I think filter always returns an array, so no need to if it's an array here.

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
```diff
@@ -0,0 +1,188 @@
...
+	_formatOutput(operationResult) {
+		const { data } = operationResult;
+
+		if (SDKOperationResultUtils.hasErrors(operationResult)) {
+			SDKOperationResultUtils.logResultMessage(operationResult);
+			SDKOperationResultUtils.logErrors(operationResult);
+			return;
+		}
+
+		if (Array.isArray(data)) {
```
karl-anthony-ng: There may not be a need to check if data is an array. I think from the operation result, it'll will always return data as an array. I'm not 100% sure though, please check.

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
```diff
@@ -0,0 +1,188 @@
...
+
+		const localFileCabinetFoldersChoices = localFileCabinetSubFolders.map(transformFoldersToChoicesFunc);
+
+		return [
+			{
+				message: TranslationService.getMessage(QUESTIONS.SELECT_FOLDER),
+				type: CommandUtils.INQUIRER_TYPES.LIST,
+				name: COMMAND_ANSWERS.SELECTED_FOLDER,
+				choices: localFileCabinetFoldersChoices,
+			},
```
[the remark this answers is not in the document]
roctaviano: Agree.

packages/node-cli/src/services/FileSystemService.js
roctaviano: Agree. Adding the 'Objects' folder in ImportObjectsCommandGenerator.js

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
roctaviano: No, this should be removed.

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
roctaviano: Agree.

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
roctaviano: This is to protect from the possibility that it could be null.

karl-anthony-ng: But in this case, I think we know it'll always be an array right? But I'm not 100% sure so I'm ok if this check is necessary here.

packages/node-cli/src/commands/UploadFilesCommandGenerator.js
```diff
@@ -0,0 +1,190 @@
...
+	ERROR: 'ERROR',
+};
+
+const { validateArrayIsNotEmpty, showValidationResults } = require('../validation/InteractiveAnswersValidator');
+
+module.exports = class UploadFilesCommandGenerator extends BaseCommandGenerator {
+	constructor(options) {
+		super(options);
+		this._fileSystemService = new FileSystemService();
+		this.localFileCabinetFolder = path.join(this._projectFolder, ApplicationConstants.FOLDERS.FILE_CABINET);
```
davidecorreu: I think we can name it -> this._localFileCabinetFolder.

packages/node-cli/src/services/FileCabinetService.js
```diff
@@ -0,0 +1,54 @@
...
+const TEMPLATES_EMAIL_TEMPLATES_PATH = '/Templates/E-mail Templates';
+const TEMPLATES_MARKETING_TEMPLATES_PATH = '/Templates/Marketing Templates';
+const WEB_SITE_HOSTING_FILES_PATH = '/Web Site Hosting Files';
+
+const UNRESTRICTED_PATHS = [SUITESCRIPTS_PATH, TEMPLATES_EMAIL_TEMPLATES_PATH, TEMPLATES_MARKETING_TEMPLATES_PATH, WEB_SITE_HOSTING_FILES_PATH];
+
+module.exports = class FileCabinetService {
+	constructor(fileCabinetAbsolutePath) {
+		this._fileSystemService = new FileSystemService();
+		this.fileCabinetAbsolutePath = fileCabinetAbsolutePath;
```
davidecorreu: this._fileCabinetAbsolutePath ?

packages/node-cli/src/services/FileCabinetService.js
```diff
@@ -0,0 +1,54 @@
...
+				if (this._shouldEnterFolder(folder)) {
+					getFoldersRecursively(folder);
+				}
+			});
+		getFoldersRecursively(parentFolder);
+
+		return folders;
+	}
+
+	pathIsUnrestricted(path) {
```
davidecorreu: Maybe better isUnrestrictedPath(path) or _isUnrestrictedPath()?

PR #90, opened 2020-04-09
Fixed bug found by test

lfpvillegas:


packages/node-cli/src/commands/CreateProjectCommandGenerator.js
```diff
@@ -275,7 +275,7 @@ module.exports = class CreateProjectCommandGenerator extends BaseCommandGenerato
 
 				const operationResult = await this._sdkExecutor.execute(executionContextCreateProject);
 
-				if (SDKOperationResultUtils.hasErrors(operationResult)) {
+				if (operationResult.status === SDKOperationResultUtils.ERROR) {
```
davidecorreu, line 278: I prefer the hasErrors(operationResult) function than the straight comparison to the ERROR. And if so, maybe it would be better to use it like SDKOperationResultUtils.STATUS.ERROR

packages/node-cli/src/commands/ImportObjectsCommandGenerator.js
```diff
@@ -125,18 +124,18 @@ module.exports = class ImportObjectsCommandGenerator extends BaseCommandGenerato
 					{ name: NodeTranslationService.getMessage(YES), value: true },
 					{ name: NodeTranslationService.getMessage(NO), value: false },
 				],
-				validate: fieldValue => showValidationResults(fieldValue, validateArrayIsNotEmpty),
+				validate: (fieldValue) => showValidationResults(fieldValue, validateArrayIsNotEmpty),
```
davidecorreu, line 127: There is no need for the parenthesis when an arrow function is using only one argument. There is a lot of cases like this in the review. Was it done by your prettier? I personally don't like it, but I'll stick to what ever we decide as a team.

lfpvillegas: Yeah, it was done by prettier.. Let's discuss in Slack for going faster

packages/node-cli/src/commands/CreateProjectCommandGenerator.js
lfpvillegas: I was having an error in that function, so, because we deleted a similar function in ActionResultUtils, I decided to do the same here.... 
Also the method now would be better defined as, isStatusError
I like your second idea, I will implement it.

vic-nik: I think it's not bad, quite readable, the function will only have 1 line, not sure it's worth creating it

PR #101, opened 2020-04-21
Feature/pdpdevtool 2950

jbbalbas:


packages/vscode-extension/package.json
```diff
@@ -62,8 +72,9 @@
 		"@types/vscode": "^1.40.0",
 		"glob": "^7.1.5",
 		"mocha": "^6.2.2",
-		"typescript": "^3.6.4",
 		"tslint": "^5.20.0",
+		"typescript": "^3.6.4",
+		"vsce": "^1.75.0",
```
davidecorreu, line 77: Do we really need vsce as a dependency of  the vscode-extension?
Won't we be it installed globally?

jbbalbas: I'll remove it

extensions/localserver-cli-command/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@oracle/suitecloud-cli-localserver-command",
-	"version": "1.0.2",
+	"version": "1.0.10",
```
vic-nik, line 3: We should not upgrade the version without publishing to NPM

packages/node-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@oracle/suitecloud-cli",
-	"version": "1.0.2",
+	"version": "1.0.10",
```
vic-nik, line 3: We should not upgrade the version without publishing to NPM

packages/node-cli/package.json
```diff
@@ -22,7 +22,7 @@
 		"suitecloud": "./src/suitecloud.js"
 	},
 	"dependencies": {
-		"@oracle/suitecloud-cli-localserver-command": "^1.0.2",
+		"@oracle/suitecloud-cli-localserver-command": "^1.0.10",
```
vic-nik, line 25: Why is version upgrade needed?

packages/node-cli/src/core/sdksetup/SDKDownloadService.js
```diff
@@ -34,19 +34,21 @@ class SDKDownloadService {
 		this._fileSystemService = new FileSystemService();
 	}
 
-	download() {
+	async download() {
```
vic-nik, line 37: why does it need to be async?

jbbalbas: because we need to `await` for it in vscode-extension before executing any command

packages/node-cli/package.json
jbbalbas: Finally I found out that it's not needed, I'll revert it

PR #106, opened 2020-04-24
PDPDEVTOOL-3067 - Download sdk only if it's not present

jbbalbas:


packages/node-cli/src/core/sdksetup/SDKDownloadService.js
```diff
@@ -36,7 +36,9 @@ class SDKDownloadService {
 
 	async download() {
 		const sdkDirectory = this._fileSystemService.createFolder(HOME_PATH, FOLDERS.SUITECLOUD_SDK);
-
+		if (fs.existsSync(path.resolve(sdkDirectory, SDKProperties.getSDKFileName()))) {
+			return;
+		}
```
davidecorreu, line 41: This looks good to me. But I have one doubt that is the following: if we have our suitecloud-vscode-extension and a nodeCLI that use a different version of .jar we are going to face some problems, because when we download the versionA.jar we delete the other ones versions. Should we still remove the other .jar files versions?

jbbalbas: You're right, actually I said yesterday the same problem, and then I forgot to implement it. We could have nodeCLI using lagging and VSCode using leading version, so we shouldn't remove the jars anymore. I'll remove the deletion of the jar file

PR #114, opened 2020-05-07
PDPDEVTOOL-3071 - Download core jar standalone for cli and vscode

jbbalbas:


packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -31,6 +31,7 @@ module.exports = class CommandActionExecutor {
 		this._commandsMetadataService = dependencies.commandsMetadataService;
 		this._authenticationService = dependencies.authenticationService;
 		this._consoleLogger = dependencies.consoleLogger;
+		this._sdkPath = dependencies.sdkPath;
```
davidecorreu, line 34: Should we add another assert for dependencies.sdkPath ?

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
```diff
@@ -0,0 +1,16 @@
+/*
```
davidecorreu, line 1: In the CLI the name is SDKProperties with capitals letters at the beginning.

packages/vscode-extension/src/core/sdksetup/SdkValidator.ts
```diff
@@ -0,0 +1,29 @@
+
+import { spawn } from 'child_process';
+
+const SDK_INTEGRATION_MODE_JVM_OPTION = '-DintegrationMode';
+const SDK_CLIENT_PLATFORM_VERSION_JVM_OPTION = '-DclientPlatformVersion';
```
davidecorreu, line 5: Can we somehow re-use the constants from the CLI?

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+async function install() {
+	const sdkParentDirectory = fileSystemService.createFolder(homedir(), SUITECLOUD_FOLDER);
+	const sdkDirectory = fileSystemService.createFolder(sdkParentDirectory, VSCODE_SDK_FOLDER);
+	const fullURL = `${SDK_DOWNLOAD_URL}/${SDK_FILENAME}`;
+
+	try {
+		var downloadFilePromise = downloadFile(fullURL, sdkDirectory);
```
davidecorreu, line 38: Can we use const instead of var?

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
```diff
@@ -0,0 +1,103 @@
...
+		messageService.showInformationMessage(
+			translationService.getMessage(EXTENSION_INSTALLATION.IN_PROGRESS),
+			downloadFilePromise,
+			`$(sync~spin) ${translationService.getMessage(EXTENSION_INSTALLATION.IN_PROGRESS)}`
```
davidecorreu, line 42: Is this $(sync~spin) a kind of spinner? Can we add it instead of only in this message to all the statusBar messages adding this same string literal to the showStatusBarMessage function of the MessageService.ts?
It will be very cool to have it for all the statusBar messages.

packages/vscode-extension/src/util/ExtensionUtil.ts
```diff
@@ -27,4 +27,8 @@ export const NodeTranslationService = require('@oracle/suitecloud-cli/src/servic
 
 export const FileUtils = require('@oracle/suitecloud-cli/src/utils/FileUtils');
 
-export const SDKDownloadService = require('@oracle/suitecloud-cli/src/core/sdksetup/SDKDownloadService');
\ No newline at end of file
+export const SdkDownloadService = require('@oracle/suitecloud-cli/src/core/sdksetup/SdkDownloadService');
+
```
davidecorreu, line 31: This cli-service is no longer needed here.

PR #117, opened 2020-05-13
PDPDEVTOOL-2633 - Upload file command

jbbalbas:


packages/vscode-extension/src/commands/UploadFile.ts
```diff
@@ -0,0 +1,71 @@
...
+		const workspaceFolder = workspace.getWorkspaceFolder(activeFile);
+		if (!workspaceFolder) {
+			opts.messageService.showCommandError(this.translationService.getMessage(UPLOAD_FILE.ERRORS.INVALID_WORKSPACE));
+			return;
+		}
+
+		const cliConfigurationService = new CLIConfigurationService();
+		cliConfigurationService.initialize(workspaceFolder.uri.fsPath);
+		cliConfigurationService.getProjectFolder(this.commandName);
+		const projectFolder = cliConfigurationService.getProjectFolder();
```
lfpvillegas, line 32: One question...Why do we repeat two times getProjectFolder? Maybe one of them could be deleted.
(I have been analyzing a little bit the code and I think first one is ok, but maybe there is something I missed)

packages/vscode-extension/messages.json
```diff
@@ -13,6 +15,13 @@
 	"DEPLOY_DEPLOYING": "Deploying project...",
 	"LIST_OBJECTS_COMMAND": "List objects",
 	"LIST_OBJECTS_LISTING": "Listing objects...",
+	"UPLOAD_FILE_COMMAND": "Upload file",
+	"UPLOAD_FILE_UPLOADING": "Uploading files...",
+	"UPLOAD_FILE_OVERWRITE_FILES": "The file {0} will be overwritten in the File Cabinet. Do you want to continue?",
```
davidecorreu, line 20: Probably UPLOAD_FILE_OVERWRITE_FILE with no plural.

packages/vscode-extension/src/commands/UploadFile.ts
```diff
@@ -0,0 +1,70 @@
...
+		const activeFile = window.activeTextEditor?.document.uri;
+		if (!activeFile) {
+			opts.messageService.showCommandError(this.translationService.getMessage(UPLOAD_FILE.ERRORS.INVALID_ACTIVE_FILE));
+			return;
+		}
+		const workspaceFolder = workspace.getWorkspaceFolder(activeFile);
+		if (!workspaceFolder) {
+			opts.messageService.showCommandError(this.translationService.getMessage(UPLOAD_FILE.ERRORS.INVALID_WORKSPACE));
+			return;
+		}
```
davidecorreu, line 27: I think these checks are already covered in the ActionExecutor in line 20-21 (when checking for the executionPath) that is a kind of precondition to execute any type of action.  But probably the messages here are more accurate. Maybe we can move them to these precondition check in the ActionExecutor?

PR #122, opened 2020-05-22
feature/PDPDEVTOOL-3100 manageauth

lfpvillegas:


packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,271 @@
...
+         answers[ANSWERS_NAMES.REMOVE] = await this._confirmRemove(prompt);
+      }
+
+      return this._extractAnswers(answers);
+   }
+
+   async _selectAuthID(authIDList, prompt) {
+      var authIDs = Object.entries(authIDList).sort();
+      if (authIDs.length <= 0) {
+         throw Error(NodeTranslationService.getMessage(ERRORS.CREDENTIALS_EMPTY));
```
jbbalbas, line 98: throw `new CLIException(...)` or just the message string

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,271 @@
...
+              .withResultMessage(operationResult.resultMessage)
+              .build()
+         : ManageAccountActionResult.Builder.withErrors(
+              SdkOperationResultUtils.collectErrorMessages(operationResult)
+           ).build();
+   }
+
+   _getSpinnerMessage(answers) {
+      let message = "";
+      if (answers[COMMAND.OPTIONS.REMOVE]) {
```
jbbalbas, line 244: a `switch` statement would be better for this, since we have 4 (and 5 in the future, with REVOKE) options

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,271 @@
...
+   constructor(options) {
+      super(options);
+      this._authenticationService = new AuthenticationService(options.executionPath);
+      this._outputFormatter = new ManageAccountOutputFormatter(options.consoleLogger);
+   }
+
+   async _getCommandQuestions(prompt) {
+      const authIDList = await this._authenticationService.getAuthIds(this._sdkExecutor);
+      let answers = await this._selectAuthID(authIDList, prompt);
+      this._outputFormatter.logAccountInfo(answers[ANSWERS_NAMES.SELECTED_AUTH_ID]);
```
jbbalbas, line 83: the OutputFormatter should take care of the output of the command, no? In this case, we're asking the questions (input of the command). I don't think you should use outputFormatter here. Maybe it could be an option to extract `logAccountInfo` to an utility class, and call it both here and in OutputFormatter (if it's needed)

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,271 @@
...
+         },
+      ]);
+      if (!answer[ANSWERS_NAMES.REMOVE]) {
+         throw NodeTranslationService.getMessage(MESSAGES.CANCEL);
+      }
+      return answer[ANSWERS_NAMES.REMOVE];
+   }
+
+   _extractAnswers(answers) {
+      var commandAnswers = new Object();
```
jbbalbas, line 205: instead creating new Object() you can:
```
if (answers[ANSWERS_NAMES.ACTION] == ACTION.RENAME) {
  return {
    [COMMAND.OPTIONS.RENAME]: answers[ANSWERS_NAMES.SELECTED_AUTH_ID].authId,
    [COMMAND.OPTIONS.RENAMETO]: answers[ANSWERS_NAMES.RENAMETO],
  };
} else if (answers[ANSWERS_NAMES.ACTION] == ACTION.REMOVE) {
  return {
    [COMMAND.OPTIONS.REMOVE]: answers[ANSWERS_NAMES.SELECTED_AUTH_ID].authId,
  }
// } else if (answers[ANSWERS_NAMES.ACTION] == ACTION.REVOKE) {
//  return {
//    [COMMAND.OPTIONS.REVOKE]: answers[ANSWERS_NAMES.SELECTED_AUTH_ID].authId,
//  }
```

packages/node-cli/src/commands/outputFormatters/ManageAccountOutputFormatter.js
```diff
@@ -0,0 +1,61 @@
...
+   constructor(consoleLogger) {
+      super(consoleLogger);
+   }
+
+   formatActionResult(actionResult) {
+      if (actionResult.resultMessage) {
+         ActionResultUtils.logResultMessage(actionResult, this.consoleLogger);
+      } else if (Array.isArray(actionResult.data)) {
+         actionResult.data.forEach((message) => this.consoleLogger.result(message));
+      } else if (actionResult.data instanceof Object) {
```
jbbalbas: change this `else if instanceof Object` for an `else` statement

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: Use of switch is considered not good practice in terms of clean code. I understand that using if else is basically the same and the solution would be to implement polymorfism but I think this it's not worth enough for overcomplicating

Anyways, condition is that the object "answers" has a property. Answers could have several properties at the same time (like in the case of rename and renameto) so I don't think it would be possible to do this with the switch or it would increase more the complexity.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,266 @@
...
+      LIST: "list",
+      REMOVE: "remove",
+      RENAME: "rename",
+      RENAMETO: "renameto",
+   },
+   FLAGS: {
+      NO_PREVIEW: "no_preview",
+      SKIP_WARNING: "skip_warning",
+      VALIDATE: "validate",
+   },
```
davidecorreu, line 45: These FLAGS don't seem to be used.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: Fixed

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: We decided to mantain for the moment since it's an output for non-interactive -info option, and also it's a message shown in the console in the middle of interactive (not final output though, but don't be willing to overcomplicate by creating a new service)

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: Fixed!

packages/node-cli/src/commands/outputFormatters/ManageAccountOutputFormatter.js
lfpvillegas: Changed... but we could improve in the Java part to always recieve same type of Object from ManageAuth. However this would be a huge change for all tools

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: Deleted!

packages/node-cli/src/core/authentication/AuthenticationService.js
```diff
@@ -53,4 +64,19 @@ module.exports = class AuthenticationService {
...
+		const getAuthListContext = new SdkExecutionContext({
+			command: COMMANDS.MANAGEAUTH,
+			flags: [FLAGS.LIST],
+		});
+		const existingAuthIDsResponse = await executeWithSpinner({
+			action: sdkExecutor.execute(getAuthListContext),
+			message: NodeTranslationService.getMessage(COMMAND_SETUPACCOUNT.MESSAGES.GETTING_AVAILABLE_AUTHIDS),
+		});
+		if (existingAuthIDsResponse.status === SdkOperationResultUtils.STATUS.ERROR) {
+			throw SdkOperationResultUtils.getResultMessage(existingAuthIDsResponse);
```
davidecorreu, line 78: In general I think is will be better to use the SdkOperationResultUtils.collectErrorMessages, it can log more information if it's present.

packages/node-cli/messages.json
```diff
@@ -122,6 +126,33 @@
 	"COMMAND_LOCAL_RESOURCE_NOT_FOUND": "The process failed. The {0} file was not found in the project.",
 	"COMMAND_LOCAL_INVALID_XML": "Invalid xml {0}",
 
+	
+	"COMMAND_MANAGE_ACCOUNT_ERROR_CREDENTIALS_EMPTY": "There is no authentication ID (authID) configured for this account-role combination. Run \"suitecloud account:setup\" to set up a NetSuite account to use with SuiteCloud CLI for Node.js.",
```
davidecorreu, line 130: The first sentence of the message doesn't seem to be right, specially the account-role combination part.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,263 @@
...
+      choices.push(new inquirer.Separator());
+      let answers = await prompt([
+         {
+            type: CommandUtils.INQUIRER_TYPES.LIST,
+            name: ANSWERS_NAMES.SELECTED_AUTH_ID,
+            message: NodeTranslationService.getMessage(MESSAGES.SELECT_CONFIGURED_AUTHID),
+            choices: choices,
+         },
+      ]);
+      return answers;
```
davidecorreu, line 114: I think we can directly `return prompt([...]);` . We are already use `await` to get the result of this method.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,263 @@
...
+   constructor(options) {
+      super(options);
+      this._authenticationService = new AuthenticationService(options.executionPath);
+      this._outputFormatter = new ManageAccountOutputFormatter(options.consoleLogger);
+   }
+
+   async _getCommandQuestions(prompt) {
+      const authIDList = await this._authenticationService.getAuthIds(this._sdkExecutor);
+      let answers = await this._selectAuthID(authIDList, prompt);
+      this._outputFormatter.logAccountCredentials(answers[ANSWERS_NAMES.SELECTED_AUTH_ID]);
```
davidecorreu, line 78: I'm not sure about this line. I thought that the outputFormatter was only taking care of the final output of the command, but maybe I'm wrong. From my point of view this is like a middle step used to collect the answers.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,263 @@
...
+         name: ANSWERS_NAMES.RENAMETO,
+         message: NodeTranslationService.getMessage(QUESTIONS.NEW_NAME),
+         filter: (fieldValue) => fieldValue.trim(),
+         validate: (fieldValue) =>
+            showValidationResults(
+               fieldValue,
+               validateFieldIsNotEmpty,
+               validateFieldHasNoSpaces,
+               (fieldValue) => validateSameAuthID(fieldValue, originalAuthId),
+               (fieldValue) => validateAuthIDNotInList(fieldValue, Object.keys(authIDList)),
```
davidecorreu, line 161: As the originalAuthId is already in the authIDList maybe we can skip that validateSameAuthID ?

packages/node-cli/src/validation/InteractiveAnswersValidator.js
```diff
@@ -123,7 +123,13 @@ class InteractiveAnswersValidator {
 	validateProjectType(value) {
 		return [ApplicationConstants.PROJECT_SUITEAPP, ApplicationConstants.PROJECT_ACP].includes(value)
 			? VALIDATION_RESULT_SUCCESS
-			: VALIDATION_RESULT_FAILURE(NodeTranslationService.getMessage(ANSWERS_VALIDATION_MESSAGES.WRONG_PROJECT_TYPE));
+			: VALIDATION_RESULT_FAILURE(NodeTranslationService.getMessage(ANSWERS_VALIDATION_MESSAGES.CURRENT_AUTHID));
```
davidecorreu, line 126: I think this message was changed by mistake.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,263 @@
...
+            name: ANSWERS_NAMES.SELECTED_AUTH_ID,
+            message: NodeTranslationService.getMessage(MESSAGES.SELECT_CONFIGURED_AUTHID),
+            choices: choices,
+         },
+      ]);
+      return answers;
+   }
+
+   async _selectAction(prompt) {
+      let answer = await prompt({
```
davidecorreu, line 118: We can use const here.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,263 @@
...
+
+      if (answer[ANSWERS_NAMES.ACTION] == ACTION.EXIT) {
+         throw NodeTranslationService.getMessage(MESSAGES.CANCEL);
+      }
+
+      return answer[ANSWERS_NAMES.ACTION];
+   }
+
+   async _introduceNewName(prompt, authIDList, originalAuthId) {
+      let answer = await prompt({
```
davidecorreu, line 150: We can use const here.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,263 @@
...
+         answers[ANSWERS_NAMES.RENAMETO] = await this._introduceNewName(prompt, authIDList, selectedAuthID);
+      } else if (answers[ANSWERS_NAMES.ACTION] == ACTION.REMOVE) {
+         answers[ANSWERS_NAMES.REMOVE] = await this._confirmRemove(prompt);
+      }
+
+      return this._extractAnswers(answers);
+   }
+
+   async _selectAuthID(authIDList, prompt) {
+      var authIDs = Object.entries(authIDList).sort();
```
davidecorreu, line 91: We can use const here.

packages/node-cli/messages.json
lfpvillegas: Hey David... The message was defined as this in the story, I have just copied it. Anyways I will check with Alba if it's the definitive.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: Javier and I were discussing just the same and concluded to leave it like this because:

For suitecloud account:manageauth -info this method would be final output... so that way we reuse code.
It's still an output, because it's showing the info of the selected account. Even if it is not final output, I give you that point.
Alternative it's to create a service only for this case and I think it's making it more complex.

However, if you still have some doubt about this we could discuss rest of the team and decide if output is only final or also middle steps.

packages/node-cli/messages.json
lfpvillegas: At the end... the message it's right

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: It sends a different error message so It's worth it to have it separated.

packages/node-cli/src/commands/outputFormatters/ManageAccountOutputFormatter.js
lfpvillegas: We are not going to change OperationResult at the end... instead we are changing ActionResult with a new variable actionExecuted

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: Changed and moved logic to new Calss AccountCredentialsService

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,288 @@
...
+      if (answers[COMMAND.OPTIONS.LIST]) {
+         flags.push(COMMAND.OPTIONS.LIST);
+         delete sdkParams[COMMAND.OPTIONS.LIST];
+      }
+      const executionContext = new SdkExecutionContext({
+         command: this._commandMetadata.sdkCommand,
+         params: sdkParams,
+         flags,
+      });
+
```
jbbalbas, line 222: we could have:
`const actionExecuted = this.getActionExecuted(answers)`

and then use this action when calling `this._getSpinnerMessage(actionExecuted)`
So that we could simplify `_getSpinnerMessage` function to:
```
_getSpinnerMessage(action) {
      switch(action) {
            case ACTION.REMOVE:
                  return NodeTranslationService.getMessage(MESSAGES.REMOVING);
                  break;
            case ACTION.RENAME:
                  return NodeTranslationService.getMessage(MESSAGES.RENAMING);
                  break;
            case ACTION.LIST:
                  return NodeTranslationService.getMessage(MESSAGES.LISTING);
                  break
            case ACTION.REVOKE:
                  return NodeTranslationService.getMessage(MESSAGES.REVOKING);
                  break;
            default:
                  return NodeTranslationService.getMessage(MESSAGES.INFO, answers.info);
      }
   }
```

and we could also modify the cal to:
`this.prepareData(actionExecuted, operationResult.data)` 
and 
`.withActionExecuted(actionExecuted)`

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,288 @@
...
+         message = NodeTranslationService.getMessage(MESSAGES.LISTING);
+         // } else if (answers.hasOwnProperty(COMMAND.OPTIONS.REVOKE)) {
+         //    message = NodeTranslationService.getMessage(MESSAGES.REVOKING);
+      } else if (answers.hasOwnProperty(COMMAND.OPTIONS.INFO)) {
+         message = NodeTranslationService.getMessage(MESSAGES.INFO, answers.info);
+      }
+      return message;
+   }
+
+   prepareData(answers, data) {
```
jbbalbas, line 256: if it's private it should be `_prepareData`

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,288 @@
...
+      assert(data.hasOwnProperty(DATA_PROPERTIES.ACCOUNT_INFO));
+      let actionResultData = { authId: answers.info, accountInfo: data.accountInfo };
+      if (data.hasOwnProperty(DATA_PROPERTIES.URLS)) {
+         actionResultData[DOMAIN] = data.urls.app;
+      }
+      return actionResultData;
+   }
+
+
+   getActionExecuted(answers) {
```
jbbalbas, line 269: if it's private it should be `_getActionExecuted`

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,284 @@
...
+			delete sdkParams[COMMAND.OPTIONS.LIST];
+		}
+		const executionContext = new SdkExecutionContext({
+			command: this._commandMetadata.sdkCommand,
+			params: sdkParams,
+			flags,
+		});
+
+		const action = this._getActionExecuted(answers);
+		const message = this._getSpinnerMessage(action);
```
davidecorreu, line 223: _getSpinnerMessage() needs both action and answers as parameters. And I think we don't need to store the result in a variable, we can directly call it in the executeWithSpinner.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,284 @@
...
+
+		const operationResult = await executeWithSpinner({
+			action: this._sdkExecutor.execute(executionContext),
+			message: message,
+		});
+
+		return operationResult.status === SdkOperationResultUtils.STATUS.SUCCESS
+			? ManageAccountActionResult.Builder.withData(this._prepareData(action, operationResult.data))
+					.withResultMessage(operationResult.resultMessage)
+					.withActionExecuted(action)
```
davidecorreu, line 233: withExecutedAction ?

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,284 @@
...
+			case ACTION.RENAME:
+				return NodeTranslationService.getMessage(MESSAGES.RENAMING);
+			case ACTION.LIST:
+				return NodeTranslationService.getMessage(MESSAGES.LISTING);
+			case ACTION.REVOKE:
+				return NodeTranslationService.getMessage(MESSAGES.REVOKING);
+			case ACTION.INFO:
+				return NodeTranslationService.getMessage(MESSAGES.INFO, answers.info);
+		}
+		throw 'Error, Action inexistent';
```
davidecorreu: Either , "Error, inexistent action." or  "Error, action not implemented."

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,284 @@
...
+} = require('../services/TranslationKeys');
+
+const ACTION = {
+	LIST: 'list',
+	EXIT: 'exit',
+	INFO: 'info',
+	RENAME: 'rename',
+	REMOVE: 'remove',
+	REVOKE: 'revoke',
+};
```
davidecorreu, line 34: It will be nice to group this ACTION list only in one place, because we have the same ones in the ManageAccountOutputFormatter. Maybe in the ManageAccountActionResult, and export it from there?

packages/node-cli/src/services/AccountCredentialsService.js
```diff
@@ -0,0 +1,62 @@
...
+	COMMAND_MANAGE_ACCOUNT: {
+		MESSAGES: { ACCOUNT_INFO },
+		QUESTIONS_CHOICES,
+	},
+} = require('./TranslationKeys');
+
+const SANDBOX_ACCOUNT_ID_REGEX_PATTERN = '.+_SB\\d*$';
+const RELEASE_PREVIEW_ACCOUNT_ID_REGEX_PATTERN = '.+_RP\\d*$';
+
+module.exports = class AccountCredentialsService {
```
davidecorreu, line 20: I see this class more like a transformer/formater than a Service. I would prefer to name it AccountCredentialsFormater. And then name the methods something similar to, getAccountInfoString, and getAuthIdInfoString.
I also think it will be nice to move some of the TranslationKeys out of the COMMAND_MANAGE_ACCOUNT, because I think they are only used here.

packages/node-cli/src/commands/actionresult/ManageAccountActionResult.js
```diff
@@ -0,0 +1,73 @@
...
+const { ActionResult, ActionResultBuilder } = require('./ActionResult');
+
+const STATUS = {
+	ERROR: 'ERROR',
+	SUCCESS: 'SUCCESS',
+};
+
+const MANAGE_ACTION = {
+	LIST: 'list',
+	EXIT: 'exit',
```
davidecorreu, line 16: I don't thinks exit is a manage_action

lfpvillegas: Then, it should not be called MANAGE_ACTION

lfpvillegas: We close it for now... since this is already delaying the story

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,283 @@
...
+		if (answers.hasOwnProperty(COMMAND.OPTIONS.REMOVE)) {
+			return MANAGE_ACTION.REMOVE;
+		}
+		if (answers.hasOwnProperty(COMMAND.OPTIONS.RENAME)) {
+			return MANAGE_ACTION.RENAME;
+		}
+		assert.fail(NodeTranslationService.getMessage(ERRORS.UNKNOWN_ACTION));
+	}
+
+	_extractAuthId(answers) {
```
vic-nik, line 268: There's code duplication in _extractSelectedActipon and _extractAuthId. I'd suggest to combine them and retun one object that can be later used, e.g. { "selectedAction" : MANAGE_ACTION.INFO,  "authId" : "myid" }

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
```diff
@@ -0,0 +1,283 @@
...
+		if (answers.hasOwnProperty(COMMAND.OPTIONS.INFO)) {
+			return answers[COMMAND.OPTIONS.INFO];
+		}
+		if (answers.hasOwnProperty(COMMAND.OPTIONS.LIST)) {
+			return;
+		}
+		if (answers.hasOwnProperty(COMMAND.OPTIONS.REMOVE)) {
```
vic-nik, line 275: Are we handling REVOKE action? The check for it is present in _getSpinnerMessage.

packages/node-cli/messages.json
```diff
@@ -122,6 +126,32 @@
 	"COMMAND_LOCAL_RESOURCE_NOT_FOUND": "The process failed. The {0} file was not found in the project.",
 	"COMMAND_LOCAL_INVALID_XML": "Invalid xml {0}",
 
+	"COMMAND_MANAGE_ACCOUNT_ERROR_CREDENTIALS_EMPTY": "There is no authentication ID (authID) configured for this account-role combination. Run \"suitecloud account:setup\" to set up a NetSuite account to use with SuiteCloud CLI for Node.js.",
+	"COMMAND_MANAGE_ACCOUNT_ERROR_TAG": "The Authentication Tag/Label provided, is not valid, please review the data.",
```
vic-nik, line 130: Not sure if the first comman is needed?

packages/node-cli/messages.json
```diff
@@ -122,6 +126,32 @@
...
+	"COMMAND_MANAGE_ACCOUNT_ERROR_UNKNOWN_ACTION": "Unknown selected account:manageauth action",
```
vic-nik, line 131: Maybe "Unknown selected action for account:manageauth command" to be more readable?

packages/node-cli/messages.json
```diff
@@ -122,6 +126,32 @@
...
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_ACCOUNT_ID": "Account ID: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_ACCOUNT_NAME": "Account Name: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_ACCOUNT_TYPE": "Account Type: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_AUTHID": "Authentication ID (authID): {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_DOMAIN": "Domain: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_ROLE": "Role: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_CANCEL": "The manageauth process has been canceled.",
```
vic-nik, line 138: I think we should either say "account:manageauth" or just: "The process has been canceled"

packages/node-cli/messages.json
```diff
@@ -122,6 +126,32 @@
...
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_INFO": "Loading \"{0}\" authentication ID (authID) information:",
```
vic-nik, line 139: Are we going to change authentication ID with authentication alias?

packages/node-cli/messages.json
```diff
@@ -122,6 +123,31 @@
...
+	"COMMAND_MANAGE_ACCOUNT_ERROR_CREDENTIALS_EMPTY": "There is no authentication ID (authID) configured for this account-role combination. Run \"suitecloud account:setup\" to set up a NetSuite account to use with SuiteCloud CLI for Node.js.",
+	"COMMAND_MANAGE_ACCOUNT_ERROR_TAG": "The Authentication Tag/Label provided, is not valid, please review the data",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_ACCOUNT_ID": "Account ID: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_ACCOUNT_NAME": "Account Name: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_ACCOUNT_TYPE": "Account Type: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_AUTHID": "AuthId: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_DOMAIN": "Domain: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_ACCOUNT_INFO_ROLE": "Role: {0}",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_CANCEL": "The manageauth process has been canceled.",
+	"COMMAND_MANAGE_ACCOUNT_MESSAGES_INFO": "This are the details for CURRENTAUTHID....",
```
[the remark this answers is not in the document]
lfpvillegas: Hey Viktor... this message is outdated, please check the current one

packages/node-cli/messages.json
lfpvillegas: This message is not used at all. So I'll delete it

packages/node-cli/messages.json
lfpvillegas: Not in this story, but in a following one.

packages/node-cli/src/commands/ManageAccountCommandGenerator.js
lfpvillegas: Deleting

packages/node-cli/messages.json
lfpvillegas: Hey Viktor, I have consulted it with Alba and she thinks it fine. Also I think that this message is fine because it's consistent with other messages already in the application like "The createproject process has been canceled" and "The setupaccount process has been canceled".
In addition to that, we should think that the name of the process is not exactly the name of the command.

PR #131, opened 2020-05-29
Feature/pdpdevtool 2634

jbbalbas:


packages/node-cli/src/core/CommandActionExecutor.js
```diff
@@ -138,7 +136,7 @@ module.exports = class CommandActionExecutor {
 			const overriddenCommandArguments = beforeExecutingOutput.arguments;
 
 			const argumentsFromQuestions =
-				runInInteractiveMode || command._commandMetadata.forceInteractiveMode
+				runInInteractiveMode /*|| command._commandMetadata.forceInteractiveMode*/
```
karl-anthony-ng, line 139: Is there a reason why the commented piece of code can't be removed yet?

packages/node-cli/src/core/CommandsMetadataService.js
```diff
@@ -8,7 +8,7 @@ const path = require('path');
 const FileUtils = require('../utils/FileUtils');
 const { SDK_COMMANDS_METADATA_FILE, NODE_COMMANDS_METADATA_FILE, COMMAND_GENERATORS_METADATA_FILE } = require('../ApplicationConstants');
 const SDK_WRAPPER_GENERATOR = 'commands/SdkWrapperCommandGenerator';
-let COMMANDS_METADATA_CACHE;
+let COMMANDS_METADATA_CACHE = null;
```
karl-anthony-ng, line 11: Shouldn't leaving this undefined work the same way?

packages/vscode-extension/src/util/ExtensionUtil.ts
```diff
@@ -2,37 +2,33 @@
  ** Copyright (c) 2020 Oracle and/or its affiliates.  All rights reserved.
  ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
  */
-
 import * as vscode from 'vscode';
 
 // returns the root project folder of the active file in the editor
 // works fine with workspace with multiple project folders opened
-export function getRootProjectFolder(): string | null {
+export function getRootProjectFolder(): string | undefined {
```
karl-anthony-ng, line 9: What was the reason to change this from null to undefined?

packages/node-cli/src/core/CommandsMetadataService.js
jbbalbas: Yes, but I (personally) prefer to initialize variables and not leave them undefined. However if you prefer undefined, I can change it, the code will work the same way

packages/vscode-extension/src/util/ExtensionUtil.ts
jbbalbas: that later on, when we call this method and assign it's result to any variable, we can use the `?` parameter:
```
protected executionPath?: string;
[...]
this.executionPath = getRootProjectFolder();
```
(this code is from BaseAction.ts)

packages/node-cli/src/core/CommandActionExecutor.js
jbbalbas: I forgot to delete it. The code can't be removed, so I'll restore it

packages/node-cli/src/core/CommandsMetadataService.js
karl-anthony-ng: Just wanted to understand the change. I don't have a problem with initializing it to null.

lfpvillegas: This is an interesting conversation and I think we should talk with all the team for making a decission to make our code consistent. Maybe even make it one of our agreements.
I'm not sure but I also think that making one decission or another could lead to some implications all of us should be aware of.

packages/node-cli/src/services/AuthenticationService.js
```diff
@@ -0,0 +1,82 @@
...
+				return fileContentJson[DEFAULT_AUTH_ID_PROPERTY];
+			} catch (error) {
+				throw NodeTranslationService.getMessage(ERRORS.WRONG_JSON_FILE, projectFilePath, error);
+			}
+		}
+	},
+
+	async getAuthIds(sdkPath) {
+
+		const sdkExecutor = new SdkExecutor(sdkPath);
```
lfpvillegas, line 61: So, my question here is... we do not pass the sdkExecutor because in VsCode we cannot define it like in Node.js?

Maybe it is interesting to create another method that is

`async getAuthIds(sdkExecutor) `
And then getAuthIds(sdkPath) will only be used by VSCode and would call the previous one

`async getAuthIdsWithSDKPath(sdkPath) {
 return getAuthIds(new SdkExecutor(sdkPath);
}`

Names are not good, but just as example...

packages/vscode-extension/messages.json
```diff
@@ -19,7 +19,9 @@
 	"UPLOAD_FILE_UPLOADING": "Uploading files...",
 	"UPLOAD_FILE_OVERWRITE_FILE": "The {0} file will be overwritten in the File Cabinet. Do you want to continue?",
 	"UPLOAD_FILE_PROCESS_CANCELED": "The file upload process has been canceled. No files have been overwritten.",
-	"NOT_IN_VALID_PROJECT": "There was an error. You must have a file open in the editor.",
+	"ERROR_NO_ACTIVE_FILE": "There was an error. You must have an active file to perform this action.",
+	"ERROR_NO_ACTIVE_WORKSPACE": "There was an error. The current file does not belong to any workspace.",
+	"ERROR_NO_ACTIVE_FILE_OR_WORKSPACE": "There is no active file or the active file isn't in a workspace",
```
lfpvillegas, line 24: isn't could be "is not" to be more formal?

packages/vscode-extension/src/commands/Deploy.ts
```diff
@@ -2,30 +2,27 @@
...
 export default class Deploy extends BaseAction {
-	readonly commandName: string = 'project:deploy';
 
-	async execute(opts: { suiteCloudRunner: SuiteCloudRunner; messageService: MessageService }) {
-		const commandActionPromise = opts.suiteCloudRunner.run({
-			commandName: this.commandName,
-			arguments: {},
-		});
+	constructor() {
+		super('project:deploy');
```
lfpvillegas, line 12: You could create a constant COMMAND_NAME here

packages/vscode-extension/src/commands/ListObjects.ts
```diff
@@ -16,9 +13,12 @@ const objectTypes: {
 }[] = require('@oracle/suitecloud-cli/src/metadata/ObjectTypesMetadata');
 
 export default class ListObjects extends BaseAction {
-	readonly commandName: string = 'object:list';
+	
+	constructor() {
+		super('object:list');
```
lfpvillegas, line 18: constant

packages/vscode-extension/src/commands/ManageAccounts.ts
```diff
@@ -0,0 +1,101 @@
...
+interface SelectAuthIdItem extends QuickPickItem {
+	option: AuthIdListOption.select;
+	authId: string;
+}
+
+type AuthIdItem = NewAuthIdItem | SelectAuthIdItem;
+
+export default class ManageAccounts extends BaseAction {
+	constructor() {
+		super('account:setup');
```
lfpvillegas, line 29: Also constant here

packages/vscode-extension/src/commands/ManageAccounts.ts
```diff
@@ -0,0 +1,101 @@
...
+		const actionResult: ActionResult<AuthListData> = await accountsPromise;
+
+		if (actionResult.status === actionResultStatus.SUCCESS) {
+			const selected = await this.getAuthListOption(actionResult.data);
+			if (!selected) {
+				return;
+			} else if (selected.option === AuthIdListOption.new) {
+				this.handleNewAuth();
+			} else if (selected.option === AuthIdListOption.select) {
+				this.handleSelectAuth(selected.authId);
```
lfpvillegas, line 52: little thing... could you name the method handleSelectedAuth?

packages/vscode-extension/src/commands/ManageAccounts.ts
```diff
@@ -0,0 +1,101 @@
...
+		// ManageAccount can be executed anywhere. We don't need to be in a project folder
+		return {
+			valid: true,
+		};
+	}
+
+	protected async execute() {
+
+		const accountsPromise = AuthenticationService.getAuthIds(sdkPath);
+		this.messageService.showStatusBarMessage(`Loading the configured authentication IDs in this machine...`, true, accountsPromise);
```
lfpvillegas, line 42: Messages could go in the TranslationService way

packages/vscode-extension/src/commands/UploadFile.ts
```diff
@@ -2,32 +2,26 @@
...
-import MessageService from '../service/MessageService';
-import { COMMAND, UPLOAD_FILE, YES, NO } from '../service/TranslationKeys';
+import { COMMAND, UPLOAD_FILE, ERRORS, YES, NO } from '../service/TranslationKeys';
 import { actionResultStatus, CLIConfigurationService, ApplicationConstants, getRootProjectFolder } from '../util/ExtensionUtil';
 import BaseAction from './BaseAction';
 
 export default class UploadFile extends BaseAction {
-	readonly commandName: string = 'file:upload';
+	constructor() {
+		super('file:upload');
```
lfpvillegas, line 13: constant

PR #136, opened 2020-06-08
Bug/pdpdevtool 3282

lfpvillegas:


packages/node-cli/src/validation/InteractiveAnswersValidator.js
```diff
@@ -133,7 +133,7 @@ class InteractiveAnswersValidator {
 	}
 
 	validateAuthIDNotInList(newAuthID, authIDs) {
-		return !authIDs.includes(newAuthID)
+		return !authIDs.hasOwnProperty(newAuthID)
```
davidecorreu, line 136: This is used in ManagaAccountCommnandGenerator.js line 152 : `(fieldValue) => validateAuthIDNotInList(fieldValue, Object.keys(authIDList))`
Object.keys returns an array with the names of the authIDList, so when checking "ownproperties" of an array your are not actually cheking for the values (names of the authids) but for its indexes ( 0, 1, ... n) . 
The easy check is to run `suitecloud account:manageauth -i` and try to rename an authid to 0 or 1.

davidecorreu: We will need to use the same approach to use this validator from manageauth and setup.  So either pas a map of the authids or an array with their names from both placess.

lfpvillegas: Resolved... I change some variable names from List to Map, because I usually think in List as an Array.

packages/node-cli/src/commands/setup/SetupInputHandler.js
```diff
@@ -63,11 +63,11 @@ module.exports = class SetupInputHandler extends BaseInputHandler {
 		let authIdAnswer;
 		const choices = [];
 
-		const authIDList  = await getAuthIds(this._sdkPath);
-		if (authIDList.status === ActionResult.STATUS.ERROR) {
-			throw authIDList.errorMessages;
+		const authIDMap  = await getAuthIds(this._sdkPath);
```
davidecorreu, line 66: This const is actually an actionResult. The map of authIDs is going to be the data. I think this is the bad naming that leads what I think is an error in  line 160
`(fieldValue) => validateAuthIDNotInList(fieldValue, Object.keys(authIDMap))`
Maybe I'm wrong, but can you check it.

PR #141, opened 2020-06-11
PDPDEVTOOL-3232 / PDPDEVTOOL-3252 - Fix projectFolder for vscode and …

jbbalbas:
…SuiteCloud help header

### DIFF
--- a/packages/node-cli/src/CLI.js
+++ b/packages/node-cli/src/CLI.js
@@ -42,6 +42,11 @@ module.exports = class CLI {
 			const commandMetadataList = this._commandsMetadataService.getCommandsMetadata();
 			this._initializeCommands(commandMetadataList, runInInteractiveMode);
 
+			// If there are no arguments, we print SuiteCloud version header
+			if (!process.argv || process.argv.length <= 2) {
+				NodeConsoleLogger.info(NodeTranslationService.getMessage(TITLE, COMPATIBLE_NS_VERSION));
+			}
+
 			program
 				.version(CLI_VERSION, '--version')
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/core/SuiteCloudRunner.ts
```diff
@@ -6,13 +6,14 @@
...
 
-import { CommandActionExecutor, CommandOptionsValidator, CLIConfigurationService } from '../util/ExtensionUtil';
+import { CommandActionExecutor, CommandOptionsValidator, CLIConfigurationService, ApplicationConstants } from '../util/ExtensionUtil';
 import CommandsMetadataSingleton from '../service/CommandsMetadataSingleton';
 
 export default class SuiteCloudRunner {
 	private commandActionExecutor: any;
 
 	constructor(executionPath?: string) {
+		process.argv.push(`${ApplicationConstants.PROJECT_FOLDER_ARG}=${executionPath}`);
```
davidecorreu, line 16: Why this line does make it work in Linux?

jbbalbas: Actually, RC checked and it doesn't work in Windows either, so it's not a Linux specific bug. Also, I broke it myself with the last refactor (shame on me!)

packages/node-cli/src/CLI.js
```diff
@@ -42,6 +42,11 @@ module.exports = class CLI {
 			const commandMetadataList = this._commandsMetadataService.getCommandsMetadata();
 			this._initializeCommands(commandMetadataList, runInInteractiveMode);
 
+			// If there are no arguments, we print SuiteCloud version header
+			if (!process.argv || !process.argv.slice(2).length) {
```
vic-nik, line 46: Why do we need to slice? Can't we check (!process.argv.length || process.argv.length === 0) ?

PR #147, opened 2020-06-18
PDPDEVTOOL-3318 - Fix authid in vscode

jbbalbas:


packages/node-cli/src/commands/listobjects/ListObjectsInputHandler.js
```diff
@@ -147,9 +146,7 @@ module.exports = class ListObjectsInputHandler extends BaseInputHandler {
 			validate: (fieldValue) => showValidationResults(fieldValue, validateFieldIsNotEmpty),
 		};
 		questions.push(questionScriptId);
-
 		let answers = await prompt(questions);
-		answers[COMMAND_QUESTIONS_NAMES.AUTH_ID] = getProjectDefaultAuthId(this._executionPath);
```
davidecorreu, line 152: Maybe AUTH_ID could be removed from COMMAND_QUESTIONS_NAMES at the beginning of this file.

PR #148, opened 2020-06-18
Bug/pdpdevtool 3168 setupaccount lowercase

lfpvillegas:


packages/node-cli/src/commands/setup/SetupInputHandler.js
```diff
@@ -167,7 +167,8 @@ module.exports = class SetupInputHandler extends BaseInputHandler {
 					type: CommandUtils.INQUIRER_TYPES.INPUT,
 					name: ANSWERS.SAVE_TOKEN_ACCOUNT_ID,
 					message: NodeTranslationService.getMessage(QUESTIONS.SAVE_TOKEN_ACCOUNT_ID),
-					filter: (fieldValue) => fieldValue.trim(),
+					transformer: (answer) => answer.toUpperCase(),
```
jbbalbas, line 170: `answer` could be `fieldValue` as everywhere else, but it's a minor detail

lfpvillegas: ok now I see the change. I have done it

PR #149, opened 2020-06-22
Bug/pdpdevtool 3345

lfpvillegas:


packages/node-cli/src/commands/manageaccount/ManageAccountOutputHandler.js
```diff
@@ -13,17 +13,18 @@ module.exports = class ManageAccountOutputFormatter extends BaseOutputHandler  {
...
 
-	formatActionResult(actionResult) {
+	parse(actionResult) {
 		if (actionResult.resultMessage) {
 			ActionResultUtils.logResultMessage(actionResult, this._log);
 		}
 
 		if (actionResult.actionExecuted == MANAGE_ACTION.INFO) {
-			this._log.result(AccountCredentialsFormatter.getInfoString(actionResult.data));
+			 this._log.result(AccountCredentialsFormatter.getInfoString(actionResult.data));
```
jbbalbas, line 22: extra space

packages/node-cli/src/commands/manageaccount/ManageAccountInputHandler.js
```diff
@@ -48,13 +48,21 @@ const ANSWERS_NAMES = {
...
 module.exports = class ManageAccountInputHandler extends BaseInputHandler {
 	constructor(options) {
-        super(options);
-        this._sdkPath = options.sdkPath;
+		super(options);
+		this._sdkPath = options.sdkPath;
 	}
 
 	async getParameters(params) {
+		if (!this._runInInteractiveMode) {
```
davidecorreu, line 56: I thought that getParameters(params) was never executed if not interactive mode. But seems that is not like this. Shouldn't it be handled by our own "framework" and not execute the getPrameters or do exactly as you do here (just forward the parameters)?

packages/node-cli/src/commands/manageaccount/ManageAccountInputHandler.js
```diff
@@ -48,13 +48,21 @@ const ANSWERS_NAMES = {
...
+			return params;
+		}
+		let answers;
 		const authIDActionResult = await getAuthIds(this._sdkPath);
-		let answers = await this._selectAuthID(authIDActionResult.data, prompt);
+		if (params.hasOwnProperty(COMMAND.OPTIONS.INFO)) {
```
davidecorreu, line 61: When is this case happening it we have stop the non-interactive cases before?

packages/node-cli/src/commands/manageaccount/ManageAccountInputHandler.js
lfpvillegas: Yeah I get your point... maybe... However this was done by another story after this one, I have just done the same as in the rest of the commands to fix the bug.

Any further change about functionality or refactorization has to be done in another story because it's out of the scope of this one.

packages/node-cli/src/commands/manageaccount/ManageAccountInputHandler.js
lfpvillegas: ups... I was doing some tryouts and forget to change again... I fix it now

packages/node-cli/src/commands/manageaccount/ManageAccountInputHandler.js
```diff
@@ -48,13 +48,22 @@ const ANSWERS_NAMES = {
...
+			answers = this._extractAuthIDInfo(authIDActionResult.data, params[COMMAND.OPTIONS.INFO]);
+			delete params[COMMAND.OPTIONS.INFO];
```
davidecorreu, line 63: As this shouldn't be happening unless we do some more modifications, I think it will be better to remove it, and probably also the _extractAuthIDInfo method (If Carlos agree with the "non-bonus solution").

.p4config
```diff
@@ -0,0 +1,4 @@
+P4USER=luperez
```
karl-anthony-ng, line 1: I don't think we should be committing this file.

PR #150, opened 2020-06-23
PDPDEVTOOL-2656 - Package command

jbbalbas:


packages/node-cli/src/utils/ApplyContentProtectionUtils.js
```diff
@@ -0,0 +1,36 @@
...
+	},
+} = require('../services/TranslationKeys');
+
+function validate(args, projectFolder, commandName) {
+	let projectInfoService = new ProjectInfoService(projectFolder);
+	if (args[APPLY_CONTENT_PROTECTION] && projectInfoService.getProjectType() === PROJECT_ACP) {
+		throw NodeTranslationService.getMessage(ERRORS.APPLY_CONTENT_PROTECTION_IN_ACP);
+	}
+
+	if (args[APPLY_CONTENT_PROTECTION] && !this._projectInfoService.hasLockAndHideFiles()) {
```
davidecorreu, line 26: I think you wanted to use projectInfoService instead of this._projectInfoService

packages/node-cli/src/utils/ApplyContentProtectionUtils.js
```diff
@@ -0,0 +1,36 @@
...
+const APPLY_CONTENT_PROTECTION = 'applycontentprotection';
+
+const {
+	UTILS: {
+		APPLY_CONTENT_PROTECTION_ARGUMENT_HANDLER: { ERRORS },
+	},
+} = require('../services/TranslationKeys');
+
+function validate(args, projectFolder, commandName) {
+	let projectInfoService = new ProjectInfoService(projectFolder);
```
davidecorreu, line 21: It will fine using const instead of let.

packages/node-cli/src/utils/AccountSpecificValuesUtils.js
```diff
@@ -0,0 +1,55 @@
...
+const { PROJECT_SUITEAPP } = require('../ApplicationConstants');
+
+const ACCOUNT_SPECIFIC_VALUES = 'accountspecificvalues';
+const ACCOUNT_SPECIFIC_VALUES_OPTIONS = {
+	ERROR: 'ERROR',
+	WARNING: 'WARNING',
+};
+
+function validate(args, projectFolder) {
+	let projectInfoService = new ProjectInfoService(projectFolder);
```
davidecorreu, line 25: We can user const.

PR #153, opened 2020-07-03
Feature/pdpdevtool 2635 vs code authentication

jbbalbas:


packages/node-cli/src/core/CommandRegistrationService.js
```diff
@@ -50,7 +50,7 @@ module.exports = class CommandRegistrationService {
 
 		commandSetup.description(commandMetadata.description).action(async (options) => {
 			const actionResult = await executeCommandFunction(options);
-			process.exitCode = actionResult.status === ActionResult.STATUS.SUCCESS ? 0 : 1;
+			process.exitCode = actionResult.isSuccess() ? 0 : 1;
```
karl-anthony-ng, line 53: Is it worth it to make 0 and 1 here constants?

packages/node-cli/src/metadata/NodeCommandsMetadata.json
```diff
@@ -60,5 +60,43 @@
                 "mandatory": false
             }
         ]
+    },
+    "account:ci": {
+        "name": "account:ci",
+        "description": "seria Set up an account to use with the SuiteCloud CLI for Node.js. This command only supports non-interactive mode.",
```
karl-anthony-ng, line 66: Is the description correct? There seems to be some extra text at the start.

packages/vscode-extension/src/types/ActionResult.d.ts
```diff
@@ -10,6 +10,15 @@ export interface ActionResult<T> {
 	data: T;
 }
 
+export interface AuthenticateActionResult extends ActionResult<undefined> {
```
karl-anthony-ng, line 13: It's really undefined in ActionResult<undefined>?

jbbalbas: Yes, an AuthenticateActionResult doesn't have `data`, so it's an ActionResult with `data` property `undefined`

packages/node-cli/src/SdkExecutor.js
```diff
@@ -26,97 +26,86 @@ const UTF8 = 'utf8';
...
+					if (isIntegrationMode && output.errorCode && output.errorCode === SdkErrorCodes.NO_TBA_SET_FOR_ACCOUNT) {
+						reject(NodeTranslationService.getMessage(ERRORS.SDKEXECUTOR.NO_TBA_FOR_ACCOUNT_AND_ROLE));
+					}
+					resolve(output);
+				} catch (error) {
+					reject(NodeTranslationService.getMessage(ERRORS.SDKEXECUTOR.RUNNING_COMMAND, error));
+				}
+			} else if (code !== 0) {
+				// check if the problem was due to bad Java Version
+				const javaVersionError = this._checkIfJavaVersionIssue();
```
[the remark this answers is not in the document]
jbbalbas: Yes, I this code is old, I only extracted it into a function, but you're right, at this point we know that java version is already valid. I'll update it

PR #201, opened 2020-08-26
Bug/PDPDEVTOOL-2143

lfpvillegas:


vic-nik (approved):

packages/node-cli/src/commands/object/import/ImportObjectsOutputHandler.js
```diff
@@ -21,7 +21,8 @@ module.exports = class ImportObjectsOutputHandler extends BaseOutputHandler {
 			!actionResult.data ||
 			!(
 				(Array.isArray(actionResult.data.successfulImports) && actionResult.data.successfulImports.length) ||
-				(Array.isArray(actionResult.data.failedImports) && actionResult.data.failedImports.length)
+				(Array.isArray(actionResult.data.failedImports) && actionResult.data.failedImports.length) ||
+				(Array.isArray(actionResult.data.errorImports) && actionResult.data.errorImports.length)
```
davidecorreu, line 25: I don't see this `errorImports` in the ImportObjectsResult.java . Is this really needed?

lfpvillegas: It was failing to log an error because it was inserted in this errorImports variable. 
You have to look in  ImportObjectsAction._parsePartialResult

davidecorreu: Aha! Thanks for the clarification.

davidecorreu (approved):

PR #207, opened 2020-08-28
bug/PDPDEVTOOL-1067

lfpvillegas:


packages/node-cli/src/commands/file/import/ImportFilesAction.js
```diff
@@ -58,6 +58,10 @@ module.exports = class ImportFilesAction extends BaseAction {
 				throw NodeTranslationService.getMessage(ERRORS.IS_SUITEAPP);
 			}
 
+			if(this._runInInteractiveMode == false) {
```
karl-anthony-ng: As Javi mentioned, please change to  "if (...".

karl-anthony-ng (changes requested):

davidecorreu (approved):

vic-nik (approved):

PR #231, opened 2020-09-16
SuiteCloud Unit Testing README.md file edited

x8a:
Fixed indentation problems in a numbered list for npm publishing

### DIFF
--- a/packages/unit-testing/README.md
+++ b/packages/unit-testing/README.md
@@ -36,35 +36,35 @@ However, if you want to configure SuiteCloud Unit Testing manually, do the follo
 3. To initialize the NPM package, from the root of your SDF project folder, run `npm init`.
 >💡 A `package.json` file is created in your SDF project folder.
 4. In your `package.json` file, add the following code:
-```json
-{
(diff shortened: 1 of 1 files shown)


vic-nik (approved):

packages/unit-testing/README.md
```diff
@@ -36,35 +36,35 @@ However, if you want to configure SuiteCloud Unit Testing manually, do the follo
...
-```json
-{
-  "scripts": {
-    "test": "jest"
-  }
-}
-```
+    ```json
+    {
+    "scripts": {
```
karl-anthony-ng, line 41: Should we add one more tab here?

karl-anthony-ng (changes requested):

packages/unit-testing/README.md
x8a: I added only one because it's a small file, but we can also have two.

karl-anthony-ng (approved):

davidecorreu (approved):

PR #244, opened 2020-10-20
bugfix/PDPDEVTOOL-3765-U6-Issue-593237-SDF-Node.js-CLI-Importing-of-a…

karl-anthony-ng:
…ll-custom-objects-failed


packages/node-cli/src/commands/object/import/ImportObjectsAction.js
```diff
@@ -59,20 +66,29 @@ module.exports = class ImportObjectsAction extends BaseAction {
...
 				this._log.info(NodeTranslationService.getMessage(WARNINGS.OVERRIDE));
 				if (params[COMMAND_FLAGS.EXCLUDE_FILES]) {
 					flags.push(COMMAND_FLAGS.EXCLUDE_FILES);
 					delete params[COMMAND_FLAGS.EXCLUDE_FILES];
 				}
+
+				if (params[ANSWERS_NAMES.OBJECT_TYPE] === IMPORT_OBJECTS_COMMAND_TYPE_PARAM_ALL) {
+					scriptIdArray = (await this._getAllScriptIds(params)).map((el) => el.scriptId).join(' ').split(' ');
+				} if (params[ANSWERS_NAMES.OBJECT_TYPE] !== IMPORT_OBJECTS_COMMAND_TYPE_PARAM_ALL
+					&& params[ANSWERS_NAMES.SCRIPT_ID] === IMPORT_OBJECTS_COMMAND_SCRIPT_ID_PARAM_ALL) {
```
davidecorreu, line 87: I guess it could be an `else if(params[ANSWERS_NAMES.SCRIPT_ID] === IMPORT_OBJECTS_COMMAND_SCRIPT_ID_PARAM_ALL)`

karl-anthony-ng: Yes you are correct. Updated.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
```diff
@@ -59,20 +66,29 @@ module.exports = class ImportObjectsAction extends BaseAction {
...
+				scriptIdArray = params[ANSWERS_NAMES.SCRIPT_ID].split(' ');
 			} else {
 				this._log.info(NodeTranslationService.getMessage(WARNINGS.OVERRIDE));
 				if (params[COMMAND_FLAGS.EXCLUDE_FILES]) {
 					flags.push(COMMAND_FLAGS.EXCLUDE_FILES);
 					delete params[COMMAND_FLAGS.EXCLUDE_FILES];
 				}
+
+				if (params[ANSWERS_NAMES.OBJECT_TYPE] === IMPORT_OBJECTS_COMMAND_TYPE_PARAM_ALL) {
+					scriptIdArray = (await this._getAllScriptIds(params)).map((el) => el.scriptId).join(' ').split(' ');
```
davidecorreu, line 85: If the map() already returns an array why we need to later cascade with .join(' ').split(' ') ? I think the result will be the same original array, but probably I'm missing something.
The same happens some lines below.

karl-anthony-ng: The original map is in a different format, it includes the type. We just need the script id in this case.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
```diff
@@ -59,20 +66,29 @@ module.exports = class ImportObjectsAction extends BaseAction {
...
+				} else if (params[ANSWERS_NAMES.OBJECT_TYPE] !== IMPORT_OBJECTS_COMMAND_TYPE_PARAM_ALL
+					&& params[ANSWERS_NAMES.SCRIPT_ID] === IMPORT_OBJECTS_COMMAND_SCRIPT_ID_PARAM_ALL) {
```
davidecorreu, line 87: I think `params[ANSWERS_NAMES.OBJECT_TYPE] !== IMPORT_OBJECTS_COMMAND_TYPE_PARAM_ALL` can be removed in this statemen because it will be always `true` after the first `if`, right?

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
```diff
@@ -59,20 +66,29 @@ module.exports = class ImportObjectsAction extends BaseAction {
 
 		try {
 			const flags = [];
+			let scriptIdArray;
```
davidecorreu, line 69: If we don't initialize scripIdArray, it can lead to an error later at line 99 when trying to check `scriptIdArray.length`.
We should consider it. Because is making the following call fail
`suitecloud object:import --type usereventscript --scriptid something --destinationfolder /Objects`

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
karl-anthony-ng: You are correct. Updated.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
karl-anthony-ng: Fixed this. No need to initialize this. I added line 89 for non-interactive mode that will initialize the scriptIdArray.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
davidecorreu: What I was trying to say is that `map((el) => el.scriptId)` is already returning your desired array. I don't see why we need later to perform `.join(' ').split(' ')`.

karl-anthony-ng: Ok. Updated.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
```diff
@@ -138,4 +156,48 @@ module.exports = class ImportObjectsAction extends BaseAction {
...
+	async _getAllScriptIdsForObjectType(params) {
+		const sdkParams = {};
+		sdkParams.type = params[ANSWERS_NAMES.OBJECT_TYPE];
+
+		return this._callListObjects(params, sdkParams);
+	}
+
+	async _getAllScriptIds(params) {
+		return this._callListObjects(params, {});
+	}
```
davidecorreu, line 169: We are probably missing to forward the scriptid (if present) into the sdkParams

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
alii-gator: @davidecorreu @karl-anthony-ng 
As David said in slack, i think the scriptid here behaves as "contains".
So if a user adds a scriptid which can result in a large number of matching script ID's, this can break the execution as we have now.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
```diff
@@ -59,20 +66,31 @@ module.exports = class ImportObjectsAction extends BaseAction {
...
 				}
+
+				scriptIdArray = params[ANSWERS_NAMES.SCRIPT_ID].split(' ');
 			} else {
 				this._log.info(NodeTranslationService.getMessage(WARNINGS.OVERRIDE));
 				if (params[COMMAND_FLAGS.EXCLUDE_FILES]) {
 					flags.push(COMMAND_FLAGS.EXCLUDE_FILES);
 					delete params[COMMAND_FLAGS.EXCLUDE_FILES];
 				}
+
```
alii-gator, line 83: @karl-anthony-ng the scriptid in importobjects behaves like "contains" and not "equals"
so if a user provides a scriptid, and there are alot of matching scriptids returned, this can also break the output.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
```diff
@@ -59,20 +66,31 @@ module.exports = class ImportObjectsAction extends BaseAction {
...
+				if (params[ANSWERS_NAMES.OBJECT_TYPE] === IMPORT_OBJECTS_COMMAND_TYPE_PARAM_ALL) {
```
alii-gator, line 84: @karl-anthony-ng the scriptid in importobjects behaves like "contains" and not "equals"
so if a user provides a scriptid, and there are alot of matching scriptids returned, this can also break the output.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
karl-anthony-ng: Updated to cover this case as well.

packages/node-cli/src/commands/object/import/ImportObjectsAction.js
karl-anthony-ng: Updated code so that non-interactive mode will always make use of listobjects first so we can split the importobjects into separate requests.

alii-gator (approved):

davidecorreu (approved):

PR #260, opened 2020-11-13
bug/PDPDEVTOOL-3540-VSCode-setupaccount-providing-url-fails

x8a:
Added validation for production accounts & its message


packages/node-cli/messages.json
```diff
@@ -15,6 +15,7 @@
 	"ANSWERS_VALIDATION_MESSAGES_INVALID_EMAIL": "Invalid email address.",
 	"ANSWERS_VALIDATION_MESSAGES_MAX_LENGTH": "This field can only contain up to {0} characters.",
 	"ANSWERS_VALIDATION_MESSAGES_PROJECT_VERSION_FORMAT": "The project version must only contain digits and dots. Ensure it follows a pattern such as \"0.0.0\".",
+	"ANSWERS_VALIDATION_MESSAGES_PRODUCTION_DOMAIN": "You entered a production domain. To use a production domain, empty the filed and press Enter.",
```
victormp1988, line 18: Is Filed (field) a typo?

x8a: it is! Thanks for spotting it!

packages/node-cli/src/validation/InteractiveAnswersValidator.js
```diff
@@ -149,4 +151,16 @@ module.exports = {
 			? VALIDATION_RESULT_SUCCESS
 			: VALIDATION_RESULT_FAILURE(NodeTranslationService.getMessage(ANSWERS_VALIDATION_MESSAGES.MAX_LENGTH, maxLength));
 	},
+
+	validateNonProductionDomain(fieldValue) {
+		return !fieldValue.match(PRODUCTION_DOMAIN_REGEX)
+			? VALIDATION_RESULT_SUCCESS
+			: VALIDATION_RESULT_FAILURE(NodeTranslationService.getMessage(ANSWERS_VALIDATION_MESSAGES.PRODUCTION_DOMAIN));
```
victormp1988, line 158: I would suggest trying to use positive conditional instead of negative. It's always harder reading code with negative ones.

packages/node-cli/src/validation/InteractiveAnswersValidator.js
```diff
@@ -149,4 +151,16 @@ module.exports = {
...
+	},
+
+	validateNonProductionAccountSpecificDomain(fieldValue) {
+		return !fieldValue.match(PRODUCTION_ACCOUNT_SPECIFIC_DOMAIN_REGEX)
+			? VALIDATION_RESULT_SUCCESS
+			: VALIDATION_RESULT_FAILURE(NodeTranslationService.getMessage(ANSWERS_VALIDATION_MESSAGES.PRODUCTION_DOMAIN));
```
victormp1988, line 164: Same

packages/node-cli/src/validation/InteractiveAnswersValidator.js
x8a: I thought the same but @davidecorreu suggested to do it like this. I'm fine with either way.

lfpvillegas (approved):

packages/vscode-extension/src/commands/ManageAccounts.ts
```diff
@@ -319,7 +321,7 @@ export default class ManageAccounts extends BaseAction {
 		};
 		if (url) {
 			commandParams.url = url;
-			commandParams.dev = true;
+			commandParams.dev = url !== ApplicationConstants.PROD_ENVIRONMENT_ADDRESS;
```
karl-anthony-ng, line 324: Do we also need to check account specific url here?

karl-anthony-ng (changes requested):

packages/vscode-extension/src/commands/ManageAccounts.ts
x8a: I rolled back to the original code here, since I saw there was no difference in behavior.

karl-anthony-ng (approved):

packages/node-cli/src/validation/InteractiveAnswersValidator.js
x8a: What do you think?

victormp1988: I see! As a general rule I always try to promote positive conditional, sometimes looking from a different perspective helps to apply the rule:
* Why not calling the function validateDomain?

In any case this comment is not a blocking one so you can go ahead and decide offline 👍.

victormp1988: BTW good code 😃! You are the real Full Stack

victormp1988 (approved):

packages/node-cli/src/validation/InteractiveAnswersValidator.js
x8a: Thanks, Victor!!! Getting there 💪

PR #261, opened 2020-11-19
Bug/pdpdevtool-3540-vscode-setupaccount-providing-url-fails

x8a:


packages/node-cli/src/validation/InteractiveAnswersValidator.js
```diff
@@ -5,8 +5,8 @@
 'use strict';
 
 const ApplicationConstants = require('../ApplicationConstants');
+const { PRODUCTION_DOMAIN_REGEX, PRODUCTION_ACCOUNT_SPECIFIC_DOMAIN_REGEX } = require('../ApplicationConstants');
```
lfpvillegas, line 8: I see that ApplicationConstants have another two constants being used in this file: PROJECT_SUITEAPP and PROJECT_ACP in line 125.
I would say to use only one import of ApplicationConstants. Either we delete line 7 and add the aforementioned constants to the import of line 8, or we just delete line 8 and then the constants would be used as 'ApplicationConstants.PRODUCTION_DOMAIN_REGEX'.

lfpvillegas (changes requested):

karl-anthony-ng (approved):

packages/node-cli/src/validation/InteractiveAnswersValidator.js
x8a: Thanks, Luis!

karl-anthony-ng (approved):

lfpvillegas (approved):

PR #263, opened 2020-11-24
improvement-to-vscode-extension-build-and-dependencies

karl-anthony-ng:
PDPDEVTOOL-3786

### DIFF
--- a/packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
+++ b/packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
@@ -5,12 +5,44 @@
 'use strict';
 
 import { homedir } from 'os';
-import { join } from 'path';
-import { sdkDownloadUrl, sdkFilename } from './extension.config.json';
-
-export const SDK_DOWNLOAD_URL = sdkDownloadUrl;
-export const SDK_FILENAME = sdkFilename;
+import * as path from 'path';
+import * as fs from 'fs';
+import { ApplicationConstants } from '../../util/ExtensionUtil';
 
 export const SUITECLOUD_FOLDER = '.suitecloud-sdk';
 export const VSCODE_SDK_FOLDER = 'vscode';
-export const sdkPath = join(homedir(), SUITECLOUD_FOLDER, VSCODE_SDK_FOLDER, SDK_FILENAME);
+
+const SUITECLOUD_CLI_PACKAGE_JSON = '@oracle/suitecloud-cli/package.json';
+const EXTENSION_CONFIG_JSON_FILENAME = "extension.config.json"
+const EXTENSION_CONFIG_JSON_FILE = "./" + EXTENSION_CONFIG_JSON_FILENAME;
+
+function getSdkDownloadUrl(): string {
+	if (extensionConfigJsonFileExists()) {
+		const extensionConfigJsonFile = require(EXTENSION_CONFIG_JSON_FILE);
+		return extensionConfigJsonFile.sdkDownloadUrl
+	}
+
+	const suiteCloudCliModulePackageJsonPath = require.resolve(SUITECLOUD_CLI_PACKAGE_JSON);
+	return require(suiteCloudCliModulePackageJsonPath).sdkDownloadUrl;
+}
+
+function extensionConfigJsonFileExists() {
+	return fs.existsSync(path.resolve(__dirname, EXTENSION_CONFIG_JSON_FILENAME));
+}
+
+export function getSdkPath(): string {
+	return path.join(homedir(), SUITECLOUD_FOLDER, VSCODE_SDK_FOLDER, getSdkDownloadUrl());
+}
+
+export function getSdkFilename(): string {
+	if (extensionConfigJsonFileExists()) {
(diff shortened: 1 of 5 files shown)


packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
```diff
@@ -5,12 +5,44 @@
...
+	return require(suiteCloudCliModulePackageJsonPath).sdkDownloadUrl;
+}
+
+function extensionConfigJsonFileExists() {
+	return fs.existsSync(path.resolve(__dirname, EXTENSION_CONFIG_JSON_FILENAME));
+}
+
+export function getSdkPath(): string {
+	return path.join(homedir(), SUITECLOUD_FOLDER, VSCODE_SDK_FOLDER, getSdkDownloadUrl());
+}
```
davidecorreu, line 35: I'm not sure about this one. Shouldn't end with `getSdkFileName()` instead of `getSdkDownloadUrl()` ?

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
```diff
@@ -5,12 +5,44 @@
...
+	if (extensionConfigJsonFileExists()) {
+		const extensionConfigJsonFile = require(EXTENSION_CONFIG_JSON_FILE);
+		return extensionConfigJsonFile.sdkDownloadUrl
+	}
+
+	const suiteCloudCliModulePackageJsonPath = require.resolve(SUITECLOUD_CLI_PACKAGE_JSON);
+	return require(suiteCloudCliModulePackageJsonPath).sdkDownloadUrl;
+}
+
+function extensionConfigJsonFileExists() {
```
davidecorreu, line 29: We can add `boolean` the return type of the function.

davidecorreu:
I have also noticed that the branch created for this story has a very nice descriptive name (which I like) but it doesn't contain a reference to our JIRA story PDPDEVTOOL-3786.

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
```diff
@@ -5,12 +5,44 @@
 'use strict';
 
 import { homedir } from 'os';
-import { join } from 'path';
-import { sdkDownloadUrl, sdkFilename } from './extension.config.json';
-
-export const SDK_DOWNLOAD_URL = sdkDownloadUrl;
-export const SDK_FILENAME = sdkFilename;
+import * as path from 'path';
+import * as fs from 'fs';
```
lfpvillegas, line 9: Could we only add required methods from the libraries?

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
```diff
@@ -5,12 +5,44 @@
...
+		const extensionConfigJsonFile = require(EXTENSION_CONFIG_JSON_FILE);
+		return extensionConfigJsonFile.sdkDownloadUrl
+	}
+
+	const suiteCloudCliModulePackageJsonPath = require.resolve(SUITECLOUD_CLI_PACKAGE_JSON);
+	return require(suiteCloudCliModulePackageJsonPath).sdkDownloadUrl;
+}
+
+function extensionConfigJsonFileExists() {
+	return fs.existsSync(path.resolve(__dirname, EXTENSION_CONFIG_JSON_FILENAME));
```
lfpvillegas, line 30: As __dirname is pointing to packages/vscode-extension/src/core/sdksetup, I'm wondering if this should be the location of the extension.config.json file. 
I know this is the original place stated in the story, just thinking that maybe should go to another place like "packages/vscode-extension/config/" or even in the main one "packages/vscode-extension/"

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
lfpvillegas: +1

lfpvillegas (changes requested):

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
lfpvillegas: As discussed in daily standup, for the moment we are not going to change this location. 
We may do a second thought in the future but not right now.

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
karl-anthony-ng: Both of you are correct here. Updated.

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
karl-anthony-ng: Added boolean as return type.

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
karl-anthony-ng: Added required methods from the libraries.

davidecorreu (approved):

lfpvillegas (approved):

PR #274, opened 2020-12-03
As a user, I want to see the output more clearly so i can distinguish…

karl-anthony-ng:
… between the outcome of each command

### DIFF
--- a/packages/vscode-extension/src/commands/BaseAction.ts
+++ b/packages/vscode-extension/src/commands/BaseAction.ts
@@ -7,74 +7,80 @@ import SuiteCloudRunner from '../core/SuiteCloudRunner';
 import MessageService from '../service/MessageService';
 import VSConsoleLogger from '../loggers/VSConsoleLogger';
 import { VSTranslationService } from '../service/VSTranslationService';
-import { CLIConfigurationService, ProjectInfoService, getRootProjectFolder } from '../util/ExtensionUtil';
+import { getRootProjectFolder } from '../util/ExtensionUtil';
 import { ERRORS } from '../service/TranslationKeys';
 import { getTimestamp } from '../util/DateUtils';
 
 export default abstract class BaseAction {
 
-	protected readonly translationService: VSTranslationService;
-	protected executionPath?: string;
-	protected readonly messageService: MessageService;
-	protected readonly commandName: string;
+    private vsConsoleLogger = new VSConsoleLogger();
 
-	protected abstract async execute(): Promise<void>;
+    protected readonly translationService: VSTranslationService;
+    protected executionPath?: string;
+    protected readonly messageService: MessageService;
+    protected readonly commandName: string;
 
-	constructor(commandName: string) {
-		this.commandName = commandName;
-		this.messageService = new MessageService(this.commandName);
-		this.translationService = new VSTranslationService();
-	}
+    protected abstract async execute(): Promise<void>;
+
+    constructor(commandName: string) {
+        this.commandName = commandName;
+        this.messageService = new MessageService(this.commandName);
+        this.translationService = new VSTranslationService();
+    }
 
     private logToOutputExecutionDetails(): void {
-        const vsConsoleLogger = new VSConsoleLogger()
         if (this.executionPath) {
-            const cliConfigurationService = new CLIConfigurationService();
-            cliConfigurationService.initialize(this.executionPath);
-            cliConfigurationService.getProjectFolder(this.commandName);
-    
-            const projectInfoService = new ProjectInfoService(cliConfigurationService.getProjectFolder(this.commandName));
-            vsConsoleLogger.info(getTimestamp() + " - " + projectInfoService.getProjectName());
+            this.vsConsoleLogger.info(getTimestamp() + " - " + BaseAction.getProjectFolderName(this.executionPath));
         } else {
-            vsConsoleLogger.info(getTimestamp());
+            this.vsConsoleLogger.info(getTimestamp());
         }
     }
 
-	protected init() {
-		this.executionPath = getRootProjectFolder();
-	}
+    private static getProjectFolderName(executionPath: string): string {
+        const executionPathParts = executionPath.replace(/\\/g, "/").split("/");
 
-	protected validate(): { valid: false; message: string } | { valid: true } {
-		if (!this.executionPath) {
-			return {
-				valid: false,
-				message: this.translationService.getMessage(ERRORS.NO_ACTIVE_FILE_OR_WORKSPACE),
-			};
-		} else {
-			return {
-				valid: true,
-			};
-		}
-	}
-
-	protected async runSuiteCloudCommand(args: { [key: string]: string } = {} ) {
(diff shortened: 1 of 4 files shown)


karl-anthony-ng:
I did an npm install on all the packages. package.lock.json was updated for vscode-extension package.

karl-anthony-ng:
Please also ignore the Pre-publish Build - master check, this is an error in our ci configuration.

packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -41,11 +58,13 @@ export default abstract class BaseAction {
 	}
 
 	protected async runSuiteCloudCommand(args: { [key: string]: string } = {} ) {
+        this.logToOutputExecutionDetails();
```
lfpvillegas, line 61: Identation seems different with following line. I think it's because there are parts of the file with tabs and others with spaces

lfpvillegas (changes requested):

packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -23,6 +26,20 @@ export default abstract class BaseAction {
...
 
+    private logToOutputExecutionDetails(): void {
+        const vsConsoleLogger = new VSConsoleLogger()
+        if (this.executionPath) {
+            const cliConfigurationService = new CLIConfigurationService();
+            cliConfigurationService.initialize(this.executionPath);
+            cliConfigurationService.getProjectFolder(this.commandName);
+    
+            const projectInfoService = new ProjectInfoService(cliConfigurationService.getProjectFolder(this.commandName));
+            vsConsoleLogger.info(getTimestamp() + " - " + projectInfoService.getProjectName());
```
davidecorreu, line 37: It's a very nice work, but I think it will be easier for the user to just see the name last directory of the execution path because it will be easier to identify the project from the tree view.
I know is not well specified in the story, so maybe we can ask about it.

packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -41,11 +58,13 @@ export default abstract class BaseAction {
 	}
 
 	protected async runSuiteCloudCommand(args: { [key: string]: string } = {} ) {
+        this.logToOutputExecutionDetails();
+
 		return new SuiteCloudRunner(this.executionPath).run({
 			commandName: this.commandName,
 			arguments: args,
 		});
```
davidecorreu, line 66: We should add a new line after this so in that way the next output doesn't get mixed with this one.

packages/vscode-extension/src/util/DateUtils.ts
```diff
@@ -0,0 +1,12 @@
...
+ ** Copyright (c) 2020 Oracle and/or its affiliates.  All rights reserved.
+ ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+ */
+
+export function getTimestamp(): string {
+    const dateNow = new Date();
+    const datePartsArray = [dateNow.getFullYear(), dateNow.getMonth() + 1, dateNow.getDate()].map(date => date.toString().length === 1 ? "0" + date : date);
+    const timePartsArray = [dateNow.getHours(), dateNow.getMinutes(), dateNow.getSeconds()].map(time => time.toString().length === 1 ? "0" + time : time);
+
+    return datePartsArray.join("/") + " " + timePartsArray.join(":");
```
davidecorreu, line 11: I like `/` but maybe we should use `-` to join the date parts as in story description.

packages/vscode-extension/src/commands/BaseAction.ts
davidecorreu: During the stand up we resolved to go for the same approach as in Webstorm, so using the folder name where the project is placed.

packages/vscode-extension/src/commands/BaseAction.ts
karl-anthony-ng: Updated the file. Replaced all tabs with spaces.

packages/vscode-extension/src/commands/BaseAction.ts
karl-anthony-ng: This makes sense. I updated the pull request to use the same approach as WebStorm.

packages/vscode-extension/src/commands/BaseAction.ts
karl-anthony-ng: Added the newline.

packages/vscode-extension/src/util/DateUtils.ts
karl-anthony-ng: Thanks for noticing this. I was so used to using "/" for dates that I didn't notice the story mentioned to make use of "-". I updated this.

packages/vscode-extension/src/commands/BaseAction.ts
davidecorreu: It should be all tabs instead of spaces according to the .pretierrc file in the root of our netsuite-suitecloud-sdk repo.
```
{
	"tabWidth": 4,
	"trailingComma": "es5",
	"singleQuote": true,
	"useTabs": true,
	"printWidth": 150
}
```

packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -5,56 +5,82 @@
...
+        } else {
+            this.vsConsoleLogger.info(getTimestamp());
+        }
+    }
+
+    private static getProjectFolderName(executionPath: string): string {
+        const executionPathParts = executionPath.replace(/\\/g, "/").split("/");
+
+        return executionPathParts[executionPathParts.length - 1];
+    }
```
davidecorreu, line 43: I don't see the value of making this method static.  Either we place in Util class, which I don't think is really needed, or we use "non static" method.
But maybe I'm missing something here. Let me know your thoughts Karl.

davidecorreu: Maybe you also want to user `path.sep`.
Check the end of: https://nodejs.org/api/path.html

karl-anthony-ng: It's a method that doesn't make use of any of the class attributes. So basically it's a utility method that's inside the class.  WebStorm was suggesting to make it static so I thought it should be fine. But since this method is not expected to be used outside of the class, maybe it's a good idea not to put static here. I updated this.

karl-anthony-ng: Regarding path.sep. It's not really needed here. What I wanted to do is to take the path (regardless if it's in Windows or Linux format) and then just transform all backslash ("\") to slash ("/") so that I don't have to worry about path separator when I split it to get the project folder name.

packages/vscode-extension/src/commands/BaseAction.ts
karl-anthony-ng: Updated to make use of tabs.

lfpvillegas (approved):

davidecorreu (approved):

PR #277, opened 2020-12-10
Feature/pdpdevtool 3782 preview flag for deploy

x8a:


packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -70,17 +75,51 @@ module.exports = class DeployAction extends BaseAction {
 				flags.push(COMMAND.FLAGS.APPLY_CONTENT_PROTECTION);
 			}
 
-			const sdkParams = CommandUtils.extractCommandOptions(params, this._commandMetadata);
+			if (params[COMMAND.FLAGS.PREVIEW]) {
+				try {
+					delete params[COMMAND.FLAGS.PREVIEW];
+					flags = flags.slice((0, 2));
```
lfpvillegas, line 81: This line is possibly deleting one of the flags (APPLY_CONTENT_PROTECTION)
Is that what we want? 

If yes, I would rather remove it by specifying it directly. If we add any other flag in the future this could be a source of problems.

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -70,17 +75,51 @@ module.exports = class DeployAction extends BaseAction {
...
+
+					if (flags.includes(COMMAND.FLAGS.VALIDATE)) {
```
lfpvillegas, line 83: You can pass from line 79 to end of the if condition to a  new method for better readibility. 
For example "execute_preview()"

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -31,12 +32,15 @@ const COMMAND = {
...
 	FLAGS: {
 		NO_PREVIEW: 'no_preview',
+		PREVIEW: 'dryrun',
 		SKIP_WARNING: 'skip_warning',
 		VALIDATE: 'validate',
 		APPLY_CONTENT_PROTECTION: 'applycontentprotection',
 	},
 };
 
+const previewCommand = 'preview';
```
karl-anthony-ng, line 42: Please change to PREVIEW_COMMAND. Use all caps and underscore for constants.

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -70,17 +75,51 @@ module.exports = class DeployAction extends BaseAction {
...
+
+					const sdkParams = CommandUtils.extractCommandOptions(params, this._commandMetadata);
+
+					const executionContextForDryrun = SdkExecutionContext.Builder.forCommand(previewCommand)
+						.integration()
+						.addParams(sdkParams)
+						.addFlags(flags)
+						.build();
+
+					const DryrunOperationResult = await executeWithSpinner({
```
karl-anthony-ng, line 95: Please use camelCase for variables. Change DryrunOperationresult to dryrunOperationResult.

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -70,17 +75,51 @@ module.exports = class DeployAction extends BaseAction {
...
+						: ActionResult.Builder.withErrors(DryrunOperationResult.errorMessages).build();
+				} catch (error) {
+					return ActionResult.Builder.withErrors([error]).build();
+				}
+			}
+
+			const sdkParams = CommandUtils.extractCommandOptions(params, this._commandMetadata);		
 
-			const executionContextForDeploy = SdkExecutionContext.Builder.forCommand(this._commandMetadata.sdkCommand)
+			const executionContextForDeploy = SdkExecutionContext.Builder.forCommand(this._commandMetadata.sdkCommand )
```
karl-anthony-ng, line 114: Please remove space before the closing parenthesis.

lfpvillegas, line 114: last space not needed before parenthesis

packages/node-cli/src/commands/project/deploy/DeployAction.js
karl-anthony-ng, line 81: There's extra parenthesis here.

packages/node-cli/src/commands/project/deploy/DeployAction.js
lfpvillegas, line 95: camelCase, dryRunOperationResult

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -70,17 +75,51 @@ module.exports = class DeployAction extends BaseAction {
...
+					delete params[COMMAND.FLAGS.PREVIEW];
+					flags = flags.slice((0, 2));
+
+					if (flags.includes(COMMAND.FLAGS.VALIDATE)) {
+						throw NodeTranslationService.getMessage(COMMAND_DEPLOY.ERRORS.VALIDATE_AND_DRYRUN_OPTIONS_PASSED);
+					}
+
+					const sdkParams = CommandUtils.extractCommandOptions(params, this._commandMetadata);
+
+					const executionContextForDryrun = SdkExecutionContext.Builder.forCommand(previewCommand)
```
lfpvillegas, line 89: I would use camelCase also in the "run" : executionContextForDryRun

packages/node-cli/.gitignore
```diff
@@ -1,2 +1,2 @@
 # config
-**/core/sdksetup/config*.json
\ No newline at end of file
+**/core/sdksetup/config.json
```
karl-anthony-ng, line 2: Was there any reason for changing this from config*.json to config.json?

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -70,17 +75,51 @@ module.exports = class DeployAction extends BaseAction {
 				flags.push(COMMAND.FLAGS.APPLY_CONTENT_PROTECTION);
 			}
 
-			const sdkParams = CommandUtils.extractCommandOptions(params, this._commandMetadata);
+			if (params[COMMAND.FLAGS.PREVIEW]) {
```
karl-anthony-ng, line 78: Since the execute command can now either call preview or deploy, maybe it would be a good idea to move the preview and deploy parts of the code to their own functions.

packages/node-cli/src/metadata/SdkCommandsMetadataPatch.json
```diff
@@ -42,5 +35,26 @@
...
+				"description": "Run a preview of your deploy process. Your project is not deployed.",
+				"mandatory": false,
+				"type": "FLAG",
+				"usage": "",
+				"defaultOption": false,
+				"disableInIntegrationMode": false
+			}
+		}
+	},
+	"object:import": {
```
lfpvillegas, line 53: Not very important comment but just wanted to say.
I know that  SdkCommandsMetadata is not ordered alphabetically because it is autogenerated from the Java code, but I think it's worth it that this file, SdkCommandsMetadataPatch, that we create manually, is filled in alphabetical order.

karl-anthony-ng (changes requested):

karl-anthony-ng:
I don't think there should be changes to package.lock.json since there was no changes to package.json. Can you check if updating package.lock.json is really necessary?

lfpvillegas (changes requested):

packages/node-cli/.gitignore
x8a: No, I'll change it back. I was doing some testing and forgot to change it back.

packages/node-cli/src/commands/project/deploy/DeployAction.js
x8a: I did this because I just wanted to delete the no_preview and skip_warning flags that are part of the constant. This only deletes those two. Any other additional flags that are passed, are kept and sent to the preview command in CLI for Java.

packages/node-cli/src/commands/project/deploy/DeployAction.js
x8a: I interpreted dry run as one word, so I didn't camel cased it.

packages/node-cli/src/commands/project/deploy/DeployAction.js
x8a: I need this extra parenthesis for slice to work properly.

karl-anthony-ng (approved):

packages/node-cli/src/commands/project/deploy/DeployAction.js
lfpvillegas: ok, I thought about it because the flag is -dry-run but it's ok

packages/node-cli/src/commands/project/deploy/DeployAction.js
lfpvillegas: ok, however I would still specify the removal of the two flags. I'm thinking about maintaining the code in the future. This numbers appear to me like ["magic numbers"](https://en.wikipedia.org/wiki/Magic_number_(programming)) 
Also, using double parenthesis could lead also to misinterpretations (as in the comment of Karl and myself, I didn't know the difference :S ). You could use instead `splice` 

I think it should be something like:
`flags.splice(flags.indexOf(COMMAND.FLAGS.NO_PREVIEW),1);
 flags.splice(flags.indexOf(COMMAND.FLAGS.SKIP_WARNING),1);
  `

lfpvillegas (approved):

PR #284, opened 2021-01-05
Added secret and center link options in object:import -i

lfpvillegas:


packages/node-cli/src/metadata/ObjectTypesMetadata.js
```diff
@@ -501,6 +510,16 @@ module.exports = [
 			relatedFiles: [{ type: 'blankscript' }],
 		},
 	},
+	{
+		name: 'Secret',
+		value: {
+			name: 'Secret',
+			type: 'secret',
+			prefix: 'custsecret',
```
davidecorreu, line 518: Probably an underscore is missing at the end of the prefix

lfpvillegas: As the way it is defined in metadata the prefix is `custsecret` without underscore.

antoniofruci (approved):

packages/node-cli/src/metadata/ObjectTypesMetadata.js
lfpvillegas: Fixed. It was already working but for the sake of consistency I added the underslash

antoniofruci (approved):

davidecorreu (approved):

PR #285, opened 2021-01-11
PDPDEVTOOL-3845: NEW SDF Objects fully supported

davidecorreu:


packages/node-cli/src/metadata/ObjectTypesMetadata.js
```diff
@@ -520,6 +520,16 @@ module.exports = [
...
 		},
 	},
+	{
+		name: 'Single Page Application',
+		value: {
+			name: 'Single Page Application',
+			type: 'singlepageapp',
+			prefix: 'custspa_',
+			hasRelatedFiles: true,
+			relatedFiles: [{ type: 'blankscript' }],
```
lfpvillegas, line 530: I think this filed should be false since we are not supporting creation of objects in NodeCLI (not supporting creation of SPA in other tools neither yet) 
Also, this is not a script but an object, if I'm not wrong

lfpvillegas (changes requested):

karl-anthony-ng (approved):

packages/node-cli/src/metadata/ObjectTypesMetadata.js
davidecorreu: From my understanding `hasRelatedFiles` means that the object will have some other files that associated to the object to make it work properly. The `type: 'blankscript'` it's just a kind of placeholder for maybe later change it to the appropriate one.
For instance a `usereventscritp` object record needs a `@NScriptType UserEventScript` .js file associated, but right now we just use `blankscript` as wildcart for all the objects.
The `Singe Page Appllication` needs to have specified at least 2 files the `clientscriptfile` and the `serverscriptfile`, that's why I have set `hasRelatedFiles: true`

lfpvillegas (approved):

PR #286, opened 2021-01-12
PDPDEVTOOL-3872

karl-anthony-ng:
Running multiple commands one after another makes the output messy

### DIFF
--- a/packages/vscode-extension/src/commands/ManageAccounts.ts
+++ b/packages/vscode-extension/src/commands/ManageAccounts.ts
@@ -4,6 +4,7 @@
  */
 import { actionResultStatus, AuthenticationUtils, InteractiveAnswersValidator, ApplicationConstants } from '../util/ExtensionUtil';
 import BaseAction from './BaseAction';
+import VSConsoleLogger from '../loggers/VSConsoleLogger';
 import { window, QuickPickItem, MessageItem } from 'vscode';
 import { AuthListData, ActionResult, AuthenticateActionResult } from '../types/ActionResult';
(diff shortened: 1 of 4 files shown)


lfpvillegas (approved):

packages/vscode-extension/src/commands/ManageAccounts.ts
```diff
@@ -57,6 +60,8 @@ export default class ManageAccounts extends BaseAction {
 	}
 
 	protected async execute() {
+		this.vsConsoleLogger.addExecutionDetailsToLog();
```
davidecorreu, line 63: Is this one really needed? I think you have already "initialized" the logger to show the execution details at line 52.

karl-anthony-ng: It looks like in the vscode-extension, the commands are being cached. So once ManageAccounts is initialized, it will not be initialized again on the subsequent calls to the command. If I don't add this line per execute, then the execution details will only be shown on the first call to the ManageAccount command.

davidecorreu: Oh! Thanks for the explanation. Maybe we should add a comment to remember in the future?

karl-anthony-ng: Yeah, I agree. I added the comment.

davidecorreu: Great!

davidecorreu (approved):

PR #287, opened 2021-01-13
Improve handling of VSConsoleLogger execution details logging.

karl-anthony-ng:


lfpvillegas (approved):

packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -16,7 +16,7 @@ export default abstract class BaseAction {
 	protected readonly messageService: MessageService;
 	protected readonly commandName: string;
 	protected executionPath?: string;
-	protected vsConsoleLogger: VSConsoleLogger = new VSConsoleLogger();
+	protected vsConsoleLogger!: VSConsoleLogger;
```
davidecorreu, line 19: Even if we set the vsConsoleLogger in the `init()` I think it will be safer to leave an initialization of the `vsConsoleLogger` either here or in the constructor to avoid further prpblems in the future if we somehow manage to skip the in `init()` (it shouldn't happen, but we never know) and then we can remove the `!` in this line.
Anyway its an improvement from before.

karl-anthony-ng: I initially coded it that way, but then I thought that it might be a source of confusion for someone reading that section of code for the first time. That person will see new VSConsoleLogger() and then in init() there's another intialization of new VSConsoleLogger(true, this.executionPath).

davidecorreu (approved):

PR #288, opened 2021-01-14
Feature/pdpdevtool 3812

antoniofruci:


packages/node-cli/messages.json
```diff
@@ -45,18 +45,18 @@
 	"COMMAND_CREATEPROJECT_OVERWRITE_PROJECT": "Do you want to overwrite the {0} project? All the files and folders from the former project will be deleted and replaced by the new project.",
 	"COMMAND_CREATEPROJECT_WRONG_PROJECT_TYPE": "Invalid \"type\" specified. The expected values are the following - \"ACCOUNTCUSTOMIZATION, SUITEAPP\".",
 
-	"COMMAND_DEPLOY_ERRORS_APPLY_CONTENT_PROTECTION_IN_ACP": "Content protection can not be applied in an Account Customziaton Project.",
+	"COMMAND_DEPLOY_ERRORS_APPLY_INSTALLATION_PREFERENCES_IN_ACP": "Installation preferences can not be applied in an Account Customizaton Project.",
```
davidecorreu, line 48: You really have good eyes for catching typos! 👏

packages/node-cli/messages.json
```diff
@@ -45,18 +45,18 @@
...
 	"COMMAND_DEPLOY_MESSAGES_LOCALLY_VALIDATED": "The \"{0}\" project has been locally validated.",
-	"COMMAND_DEPLOY_MESSAGES_NOT_APPLYING_CONTENT_PROTECTION": "The \"{0}\" project has been deployed without applying content protection.",
-	"COMMAND_DEPLOY_MESSAGES_NOT_ASKING_CONTENT_PROTECTION_REASON": "Content protection was not applied because the hiding.xml and locking.xml files were not found in the \"InstallationPreferences\" folder of the project. You need to create both files in order to apply content protection.\nTo learn how to create a hiding.xml file, see {0}.\nTo learn how to create a locking.xml file, see {1}.",
+	"COMMAND_DEPLOY_MESSAGES_NOT_APPLYING_INSTALLATION_PREFERENCES": "The \"{0}\" project has been deployed without applying installation preferences.",
+	"COMMAND_DEPLOY_MESSAGES_NOT_ASKING_INSTALLATION_PREFERENCES_REASON": "Installation preferences were not applied because the hiding.xml, locking.xml and overwriting.xml files were not found in the \"InstallationPreferences\" folder of the project. You need to create the three files in order to apply installation preferences.\nFor more information about installation preferences, see {0}",
```
davidecorreu, line 59: As far as I know, the ovewritting file is not a mandatory one. We did in this way to not break the deployment of the old projects without that file.

lfpvillegas: overwriting is definitively not mandatory.

packages/node-cli/messages.json
```diff
@@ -282,6 +282,10 @@
...
 	"UTILS_APPLY_CONTENT_PROTECTION_ARGUMENT_HANDLER_ERRORS_APPLY_CONTENT_PROTECTION_WITHOUT_HIDING_AND_LOCKING": "There was an error with the {0} command. Content protection was not applied because the hiding.xml and locking.xml files were not found in the \"InstallationPreferences\" folder of the project. You need to create both files in order to apply content protection.\nTo learn how to create a hiding.xml file, see {1}.\nTo learn how to create a locking.xml file, see {2}.",
+	"UTILS_APPLY_CONTENT_PROTECTION_ARGUMENT_HANDLER_ERRORS_APPLY_CONTENT_PROTECTION_WITH_APPLY_INSTALLATION_PREFERENCES": "The \"--applycontentprotection\" and \"--applyinstallprefs\" options are mutually exclusive. Use one of the two at a time.",
+	"UTILS_APPLY_CONTENT_PROTECTION_IS_DEPRECATED": "The \"--applycontentprotection\" option will be removed in NetSuite 2021.2. You should use the \"--applyinstallprefs\" option starting from NetSuite 2021.1.",
+	"UTILS_APPLY_INSTALLATION_PREFERENCES_ARGUMENT_HANDLER_ERRORS_APPLY_INSTALLATION_PREFERENCES_IN_ACP": "Installation preferences cannot be applied to account customization projects.",
```
lfpvillegas, line 287: This message is slight equal to one on top. Could we use the same?

packages/node-cli/messages.json
```diff
@@ -282,6 +282,10 @@
...
+	"UTILS_APPLY_INSTALLATION_PREFERENCES_ARGUMENT_HANDLER_ERRORS_APPLY_INSTALLATION_PREFERENCES_WITHOUT_HIDING_LOCKING_AND_OVERWRITING": "There was an error with the {0} command. Installation preferences were not applied because the hiding.xml, locking.xml, and overwriting.xml files were not found in the \\\"InstallationPreferences\\\" folder of the project. You need to create the three files to apply installation preferences.\\nFor more information about installation preferences, see {1}",
```
lfpvillegas, line 288: I see that this is a wrapping of another message (COMMAND_DEPLOY_MESSAGES_NOT_ASKING_INSTALLATION_PREFERENCES_REASO) . Could we use that one and do the wrapping outside in the code?

lfpvillegas (changes requested):

packages/node-cli/messages.json
antoniofruci: The problem is that COMMAND_DEPLOY_MESSAGES_NOT_ASKING_INSTALLATION_PREFERENCES_REASON has an embedded message itself, so I don't think it would work unless we structure the messages differently

packages/node-cli/src/utils/ApplyInstallationPreferencesUtils.js
```diff
@@ -0,0 +1,60 @@
...
+		APPLY_CONTENT_PROTECTION_ARGUMENT_HANDLER: { ERRORS, WARNINGS },
+		APPLY_INSTALLATION_PREFERENCES_ARGUMENT_HANDLER: { INSTALLATION_PREFERENCES_ERRORS },
+	},
+} = require('../services/TranslationKeys');
+
+function validate(args, projectFolder, commandName) {
+	const projectInfoService = new ProjectInfoService(projectFolder);
+
+	if (args[APPLY_CONTENT_PROTECTION]) {
+		NodeConsoleLogger.warning(NodeTranslationService.getMessage(WARNINGS.APPLY_CONTENT_PROTECTION_IS_DEPRECATED));
```
davidecorreu, line 27: We should not use directly the NodeConsoleLogger. Instead pass it from the commandAction that is calling  `ApplyInstallationPreferencesUtils.validate()`, the logger will properly injected to the specific commandAction from the BaseAction to its `this._log` property.

davidecorreu (approved):

packages/node-cli/messages.json
lfpvillegas: Well, I think it would work okay. If not, you could always concatenate strings.
Anyways it's not a big deal. Feel free to change it or just move it to ready for testing

lfpvillegas (approved):

PR #290, opened 2021-01-25
feature/pdpdevtool-3807

davidecorreu:


packages/vscode-extension/src/service/MessageService.ts
```diff
@@ -7,31 +7,47 @@ import { window } from 'vscode';
 import { Output } from '../suitecloud';
 import { COMMAND, SEE_DETAILS } from './TranslationKeys';
 import { VSTranslationService } from './VSTranslationService';
+import { EOL } from 'os';
```
karl-anthony-ng, line 10: Looks like this isn't being used. Can we remove it?

karl-anthony-ng (changes requested):

packages/vscode-extension/src/commands/Deploy.ts
```diff
@@ -8,6 +8,8 @@ import { actionResultStatus, ApplicationConstants, ProjectInfoServive, CLIConfig
 import BaseAction from './BaseAction';
 import { EOL } from 'os';
 
+const COMMAND_NAME = 'deploy';
+
 const DEPLOY_COMMAND = {
 	NAME: 'project:deploy',
 	OPTIONS: {
```
antoniofruci, line 15: Maybe we can already change the value of FLAGS.APPLY_INSTALLATION_PREFS since cli already supports applyinstallprefs flag

packages/vscode-extension/src/service/MessageService.ts
davidecorreu: Sure. Removed.

karl-anthony-ng (approved):

antoniofruci (approved):

packages/vscode-extension/src/commands/Deploy.ts
davidecorreu: Completely agree!

packages/vscode-extension/src/commands/Deploy.ts
```diff
@@ -8,6 +8,8 @@ import { actionResultStatus, ApplicationConstants, ProjectInfoServive, CLIConfig
 import BaseAction from './BaseAction';
 import { EOL } from 'os';
 
+const COMMAND_NAME = 'deploy';
+
 const DEPLOY_COMMAND = {
 	NAME: 'project:deploy',
```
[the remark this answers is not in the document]
davidecorreu: No needed anymore. Removed

PR #300, opened 2021-02-09
PDPDEVTOOL-3918: [VSCode] - Ensure all suitecloud commands are execut…

karl-anthony-ng:
…ed in proper context


packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -65,6 +66,17 @@ export default class UpdateObject extends BaseAction {
 				message: this.translationService.getMessage(ERRORS.NO_ACTIVE_WORKSPACE),
 			};
 		} else {
+			const projectInfoService = new ProjectInfoServive(this.executionPath);
+			if (projectInfoService.isAccountCustomizationProject() || projectInfoService.isSuiteAppProject) {
```
davidecorreu, line 70: The `new ProjectInfoService` expects a project folder as a parameter not the execution path, I'm unable to run the command properly in my vscode because I get an uncaught exception thrown by the `projectInfoService.isAccountCustomizationProject()` call.
When I did work on the vscode deploy, I had kind of the same problem, you can check there what I did, and probably we will need to extract to service that is accessible to for more commands (like deploy or more to come).

In the next line a couple of parenthesis is missing in `projectInfoService.isSuiteAppProject`.

davidecorreu (changes requested):

packages/vscode-extension/src/commands/UploadFile.ts
```diff
@@ -67,9 +67,21 @@ export default class UploadFile extends BaseAction {
 				message: this.translationService.getMessage(ERRORS.NO_ACTIVE_WORKSPACE),
 			};
 		} else {
+			const projectInfoService = new ProjectInfoServive(this.executionPath);
+			if (projectInfoService.isAccountCustomizationProject() || projectInfoService.isSuiteAppProject) {
```
x8a, line 71: I believe projectInfoService should be initialized, as David pointed out some lines above. That is, `projectInfoService.isSuiteAppProject()`.

packages/vscode-extension/messages.json
```diff
@@ -54,6 +54,8 @@
 	"ERRORS_NO_ACTIVE_FILE_OR_WORKSPACE": "There is no active file or the active file is not in a workspace.",
 	"ERRORS_SDK_JAVA_VERSION_NOT_COMPATIBLE": "Your Java version {0} is not compatible with SuiteCloud Extension for Visual Studio Code. Install Oracle JDK version {1}.",
 	"ERRORS_SDK_JAVA_VERSION_NOT_INSTALLED": "Java is not installed on this machine. Install Oracle JDK version {0} to run SuiteCloud Extension for Visual Studio Code.",
+	"ERRORS_SDF_OBJECT_MUST_BE_IN_OBJECTS_FOLDER": "The SDF object file should be in the Objects folder.",
+	"ERRORS_UPDATE_FILE_FOLDER_RESTRICTION": "The file should be in a supported folder inside the FileCabinet folder. For information about the supported folders, see https://system.netsuite.com/app/help/helpcenter.nl?fid=section_4724963992.html.",
```
x8a, line 58: 👌

packages/vscode-extension/src/commands/UploadFile.ts
karl-anthony-ng: Updated.

packages/vscode-extension/src/commands/UpdateObject.ts
karl-anthony-ng: Updated.

packages/vscode-extension/src/commands/UploadFile.ts
x8a: I still see `projectInfoService.isSuiteAppProject` and not `projectInfoService.isSuiteAppProject()`. Is that on purpose?

karl-anthony-ng: Sorry about that. Updated again.

x8a (approved):

packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -65,6 +66,18 @@ export default class UpdateObject extends BaseAction {
 				message: this.translationService.getMessage(ERRORS.NO_ACTIVE_WORKSPACE),
 			};
 		} else {
+			const projectFolderPath = this.getProjectFolderPath();
+			const projectInfoService = new ProjectInfoServive(projectFolderPath);
+			if (projectInfoService.isAccountCustomizationProject() || projectInfoService.isSuiteAppProject()) {
```
davidecorreu, line 71: I think we may be missing a try/catch block to avoid problems when for instance we have malformed manifest.xml

karl-anthony-ng: Added.

packages/vscode-extension/messages.json
```diff
@@ -54,6 +54,8 @@
 	"ERRORS_NO_ACTIVE_FILE_OR_WORKSPACE": "There is no active file or the active file is not in a workspace.",
 	"ERRORS_SDK_JAVA_VERSION_NOT_COMPATIBLE": "Your Java version {0} is not compatible with SuiteCloud Extension for Visual Studio Code. Install Oracle JDK version {1}.",
 	"ERRORS_SDK_JAVA_VERSION_NOT_INSTALLED": "Java is not installed on this machine. Install Oracle JDK version {0} to run SuiteCloud Extension for Visual Studio Code.",
+	"ERRORS_SDF_OBJECT_MUST_BE_IN_OBJECTS_FOLDER": "The SDF object file should be in the Objects folder.",
```
davidecorreu, line 57: Could we add a reference to the command?
I would like to see something similar to `The SDF object file to update should be in the Objects folder.` or add some kind of reference to the command triggered.

packages/vscode-extension/messages.json
davidecorreu, line 58: The key should be `ERRORS_UPLOAD_FILE_FOLDER_RESTRICTION` but even I think that this 2 new errors should be placed in their commands keys `UPLOAD_FILE` and `UPDATE_OBJECT` because they are not generic errors as the other ones.
And as the other message I think it should have some kind of reference to the upload file command.

davidecorreu (changes requested):

packages/vscode-extension/messages.json
x8a: I like it! Let's edit the message to what you proposed.

packages/vscode-extension/messages.json
karl-anthony-ng: Updated.

packages/vscode-extension/messages.json
karl-anthony-ng: Updated.

packages/vscode-extension/messages.json
```diff
@@ -21,9 +21,10 @@
 	"LIST_OBJECTS_LISTING": "Listing objects...",
 	"SEE_DETAILS": "See details",
 
-	"UPLOAD_FILE_UPLOADING": "Uploading files...",
+	"UPLOAD_FILE_ERROR_UPLOAD_FILE_FOLDER_RESTRICTION": "The file should be in a supported folder inside the FileCabinet folder. For information about the supported folders, see https://system.netsuite.com/app/help/helpcenter.nl?fid=section_4724963992.html.",
```
davidecorreu, line 24: Shouldn't we add a reference to upload file command?

karl-anthony-ng: Updated.

x8a: Could we change it to `The file to upload should be in a supported folder inside the FileCabinet folder. For more information...`?

packages/vscode-extension/messages.json
```diff
@@ -21,9 +21,10 @@
 	"LIST_OBJECTS_LISTING": "Listing objects...",
 	"SEE_DETAILS": "See details",
 
-	"UPLOAD_FILE_UPLOADING": "Uploading files...",
+	"UPLOAD_FILE_ERROR_UPLOAD_FILE_FOLDER_RESTRICTION": "The file being uploaded should be in a supported folder inside the FileCabinet folder. For information about the supported folders, see https://system.netsuite.com/app/help/helpcenter.nl?fid=section_4724963992.html.",
```
x8a, line 24: Could we change it to `The file to upload should be in a supported folder inside the FileCabinet folder. For more information...`?

karl-anthony-ng: Updated.

davidecorreu (approved):

PR #304, opened 2021-02-18
Feature/pdpdevtool 3992

lfpvillegas:


antoniofruci (approved):

packages/vscode-extension/package.json
```diff
@@ -71,7 +71,14 @@
 			{
 				"command": "suitecloud.uploadfile",
 				"key": "ctrl+alt+u",
-				"mac": "ctrl+alt+u"
+				"mac": "ctrl+alt+u",
+				"when": "resource =~ //FileCabinet/SuiteScripts/.*\\.js$|^file$/"
```
davidecorreu, line 75: I don't know how exactly this "regex" works, but it looks like we won't be able to upload files from SuiteApps, because they are normally located at `FileCabinet/SuiteApps/AppId/myfile.txt` .  I also see that are other folder created in the FileCabinet that we should also allow. I'm not sure if with this regex we are allowing all types of files to be uploaded or just `.js`. Please check it.

davidecorreu (changes requested):

packages/vscode-extension/package.json
alii-gator: FileCabinet may contain .pdf .html .template and images etc I suppose

karl-anthony-ng: FileCabinet can contain any kind of file.

antoniofruci (approved):

packages/vscode-extension/package.json
```diff
@@ -71,7 +71,14 @@
 			{
 				"command": "suitecloud.uploadfile",
 				"key": "ctrl+alt+u",
-				"mac": "ctrl+alt+u"
+				"mac": "ctrl+alt+u",
+				"when": "resource =~ //FileCabinet/SuiteScripts/.*$/ || resource =~ //FileCabinet/Templates/.*$/ || resource =~ //FileCabinet/Templates/E-mail Templates/.*$/ || resource =~ //FileCabinet/Templates/Marketing Templates.*$/ || resource =~ //FileCabinet/Web Site Hosting Files/.*$/ || resource =~ //FileCabinet/SuiteScripts/.*$/ || resource =~ //FileCabinet/Templates/.*$/ || resource =~ //FileCabinet/Templates/E-mail Templates/.*$/ || resource =~ //FileCabinet/Templates/Marketing Templates.*$/ || resource =~ //FileCabinet/SuiteApps/.*$/"
```
davidecorreu, line 75: This could be much shorter if you use the "or" `|` operator inside the regex , or even if create a regex that at least share the common `/FileCabinet/`.

davidecorreu (approved):

PR #309, opened 2021-02-26
PDPDEVTOOL-3952: [VSCode] Pass VSCode specific info to the core (SDK)…

karl-anthony-ng:
… for metrics collection - Platform, Platform Version

### DIFF
--- a/packages/node-cli/src/SdkExecutor.js
+++ b/packages/node-cli/src/SdkExecutor.js
@@ -26,11 +26,18 @@ const CLOSE_EVENT = 'close';
 const UTF8 = 'utf8';
 
 module.exports = class SdkExecutor {
-	constructor(sdkPath) {
+	constructor(sdkPath, executionEnvironmentContext) {
 		this._sdkPath = sdkPath;
 
 		this._CLISettingsService = new CLISettingsService();
+		this._environmentInformationService = new EnvironmentInformationService();
 		this.childProcess = null;
+
+		if (!executionEnvironmentContext) {
+			this._executionEnvironmentContext = new ExecutionEnvironmentContext();
+		} else {
+			this._executionEnvironmentContext = executionEnvironmentContext;
+		}
 	}
 
 	execute(executionContext, token) {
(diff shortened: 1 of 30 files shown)


packages/node-cli/src/SdkExecutor.js
```diff
@@ -29,7 +30,6 @@ module.exports = class SdkExecutor {
 		this._sdkPath = sdkPath;
 
 		this._CLISettingsService = new CLISettingsService();
-		this._environmentInformationService = new EnvironmentInformationService();
```
davidecorreu, line 32: The `this._environmentInformationService = new EnvironmentInformationService();` will still be needed by `_checkIfJavaVersionIssue()`

lfpvillegas (approved):

packages/node-cli/src/SdkExecutor.js
karl-anthony-ng: Added back.

lfpvillegas (approved):

packages/node-cli/src/commands/Command.js
```diff
@@ -38,9 +38,10 @@ class Command {
 		this._runInInteractiveMode = options.runInInteractiveMode;
 		this._interactiveSupport = options.interactiveSupport;
 		this._log = options.log;
+		this._executionEnvironmentContext = options.executionEnvironmentContext;
 
-		this._action = new action(options);
-		this._inputHandler = new inputHandler(options);
+		this._action = new action(options, this._executionEnvironmentContext);
+		this._inputHandler = new inputHandler(options, this._executionEnvironmentContext);
```
davidecorreu, line 44: First of all, great refactor Karl it looks much cleaner than the previous implementation.!!!
The executionEnvironmentContext is included in the options object, and we don't need to add an extra parameter when constructing the Action or InputHandler.
Actually in the BaseAction this second parameter is not used and the executionEnvironmentContet is take from the options. Same comment is valid for InputHandler. I also think that sdkPath can be added as a property in the BaseInputHandler
Snnipet from BaseAction:
```
module.exports = class BaseAction {
	constructor(options) {
		this._projectFolder = options.projectFolder;
		this._commandMetadata = options.commandMetadata;
		this._executionPath = options.executionPath;
		this._runInInteractiveMode = options.runInInteractiveMode;
		this._log = options.log;
		this._sdkPath = options.sdkPath;


		this._sdkExecutor = new SdkExecutor(this._sdkPath);
		this._executionEnvironmentContext = options.executionEnvironmentContext;
		this._sdkExecutor = new SdkExecutor(this._sdkPath, this._executionEnvironmentContext);
	}
```

packages/node-cli/src/commands/file/import/ImportFilesInputHandler.js
```diff
@@ -53,7 +53,7 @@ module.exports = class ImportFilesInputHandler extends BaseInputHandler {
 	constructor(options) {
 		super(options);
 		// TODO input handlers shouldn't execute actions. rework this
-		this._sdkExecutor = new SdkExecutor(options.sdkPath);
+		this._sdkExecutor = new SdkExecutor(options.sdkPath, this._executionEnvironmentContext);
```
davidecorreu, line 56: If we add the sdkPath as a property of the BaseInputHandler we can rewrite this line as:
`this._sdkExecutor = new SdkExecutor(this._sdkPath, this._executionEnvironmentContext);` .

packages/node-cli/src/commands/Command.js
karl-anthony-ng: I agree, it's looks much cleaner now.
Did some more updates based on your suggestion.

packages/node-cli/src/commands/file/import/ImportFilesInputHandler.js
karl-anthony-ng: Applied this change.

davidecorreu (approved):

PR #327, opened 2021-03-15
Bug/pdpdevtool 3873

lfpvillegas:
…s called

### DIFF
--- a/packages/node-cli/src/CLI.js
+++ b/packages/node-cli/src/CLI.js
@@ -22,6 +22,8 @@ const PACKAGE_FILE = `${path.dirname(require.main.filename)}/../package.json`;
 const configFile = require(PACKAGE_FILE);
 const CLI_VERSION = configFile ? configFile.version : 'unknown';
 const COMPATIBLE_NS_VERSION = '2020.2';
+const HELP_ALIAS = '-h';
+const HELP_OPTION = '--help';
 
 module.exports = class CLI {
 	constructor(dependencies) {
@@ -37,9 +39,12 @@ module.exports = class CLI {
 
 	start(process) {
 		try {
-			const runInInteractiveMode = this._isRunningInInteractiveMode();
-
 			const commandMetadataList = this._commandsMetadataService.getCommandsMetadata();
+			this._validateCommandExists(commandMetadataList, process.argv[2]);
+			
+			process.argv = this._takeOutUnnecesaryArgumentsWhenHelpIsCalled();
+
+			const runInInteractiveMode = this._isRunningInInteractiveMode();
 			this._initializeCommands(commandMetadataList, runInInteractiveMode);
 
 			// If there are no arguments, we print SuiteCloud version header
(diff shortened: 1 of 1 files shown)


packages/node-cli/src/CLI.js
```diff
@@ -37,9 +39,12 @@ module.exports = class CLI {
 
 	start(process) {
 		try {
-			const runInInteractiveMode = this._isRunningInInteractiveMode();
-
 			const commandMetadataList = this._commandsMetadataService.getCommandsMetadata();
+			this._validateCommandExists(commandMetadataList, process.argv[2]);
```
tourajvaziri, line 43: process.argv[2]);    will never be out of index? for example i call 'suitecloud'. isn't the args just 2? 0 and 1.

packages/node-cli/src/CLI.js
```diff
@@ -49,18 +54,49 @@ module.exports = class CLI {
 
 			program
 				.version(CLI_VERSION, '--version')
-				.option(`${INTERACTIVE_ALIAS}, ${INTERACTIVE_OPTION}`, NodeTranslationService.getMessage(INTERACTIVE_OPTION_DESCRIPTION))
-				.on('command:*', args => {
+				.option(
+					`${INTERACTIVE_ALIAS}, ${INTERACTIVE_OPTION}`,
+					NodeTranslationService.getMessage(INTERACTIVE_OPTION_DESCRIPTION),
+					this._validateInteractive
```
tourajvaziri, line 60: is this the right place to add this.validateinteractive?  I see from the definition that this parameter is for the default value ?
option(flags: string, description?: string, defaultValue?: string | boolean):

packages/node-cli/src/CLI.js
lfpvillegas: No, it would be "undefined" and then in the if of line 73 would not enter.

packages/node-cli/src/CLI.js
lfpvillegas: validateInteractive is a function so the definition that applies is the following one:

`option<T>(flags: string, description: string, fn: (value: string, previous: T) => T, defaultValue?: T): this;`

I have search the example in the documentation of [commander](https://github.com/tj/commander.js#custom-option-processing)

> You may specify a function to do custom processing of option-arguments. The callback function receives two parameters, the user specified option-argument and the previous value for the option. It returns the new value for the option.
> 
> This allows you to coerce the option-argument to the desired type, or accumulate values, or do entirely custom processing.
> 
> You can optionally specify the default/starting value for the option after the function parameter.

In this case, I am not passing any value to the function because I don't need to, because the variable that I'm using is already in the global scope of the class.

packages/node-cli/src/CLI.js
tourajvaziri: great.

packages/node-cli/src/CLI.js
tourajvaziri: got it. I forgot that the defaultvalue parameter is optional. looks good. thanks!

packages/node-cli/src/CLI.js
```diff
@@ -49,18 +54,49 @@ module.exports = class CLI {
...
 				})
 				.usage(NodeTranslationService.getMessage(USAGE))
 				.parse(process.argv);
-
 		} catch (exception) {
 			NodeConsoleLogger.error(unwrapExceptionMessage(exception));
 		}
 	}
 
+	_validateCommandExists(commandMetadataList, commandName) {
```
davidecorreu, line 72: I would name the second parameter `arg2` or `secondCommandLineArg` instead of `commadName`.

packages/node-cli/src/CLI.js
```diff
@@ -49,18 +54,49 @@ module.exports = class CLI {
...
+		if (commandName && !commandMetadataList.hasOwnProperty(commandName) && commandName !== HELP_ALIAS && commandName !== HELP_OPTION) {
+			throw NodeTranslationService.getMessage(ERRORS.COMMAND_DOES_NOT_EXIST, commandName);
+		}
+	}
+
+	_takeOutUnnecesaryArgumentsWhenHelpIsCalled() {
```
karl-anthony-ng, line 78: Type - Unnecesary should be Unnecessary (double s).

karl-anthony-ng (changes requested):

packages/node-cli/src/CLI.js
```diff
@@ -49,18 +54,49 @@ module.exports = class CLI {
...
+		if (process.argv.includes(HELP_ALIAS) || process.argv.includes(HELP_OPTION)) {
+			if (process.argv.length > 3) {
+				// In case there is a command
+				return process.argv.slice(0, 3).concat(HELP_ALIAS);
+			}
+			return process.argv.slice(0, 2);
```
davidecorreu, line 84: arg[0] = node.exe
arg[1] = suitecloud.js
arg[2] = existingCommandName
+ HELP_ALIAS or HELP_OPTINON
These ones already sum 4.
We are always going to get into the if statement.

packages/node-cli/src/CLI.js
```diff
@@ -37,9 +39,12 @@ module.exports = class CLI {
 
 	start(process) {
 		try {
-			const runInInteractiveMode = this._isRunningInInteractiveMode();
-
 			const commandMetadataList = this._commandsMetadataService.getCommandsMetadata();
+			this._validateCommandExists(commandMetadataList, process.argv[2]);
+			
+			process.argv = this._takeOutUnnecesaryArgumentsWhenHelpIsCalled();
```
davidecorreu, line 45: This line looks weird .... But I don't have a suggestion right now.

lfpvillegas: I was naming it checkHelp... but I thought it would be better to be more explicit.

packages/node-cli/src/CLI.js
lfpvillegas: But this is fine. Let me explain with an example:

if we have commandName --arg1 --arg2 --help, we want all arguments deleted except the --help one... 
if we have commandName --help --arg2 --arg3, the same
if we have commandName --help, we would still want help appearing.

So, this way, we ensure that, if there is a help argument and more than 3 arguments (i.e we write a command), we would show the help of the particular command.

If there is only two arguments, it's because it is the general help, without command, i.e suitecloud --help.

lfpvillegas: I'm changing it a little bit though so it would be simplified.

packages/node-cli/src/CLI.js
lfpvillegas: Well, I think it's more explicit to name it commandName, because we are looking if the command exists.
Of course there are the exceptions like the help or version flag.

packages/node-cli/src/CLI.js
```diff
@@ -49,18 +55,51 @@ module.exports = class CLI {
 
 			program
 				.version(CLI_VERSION, '--version')
```
davidecorreu, line 57: Please use the `VERSION_OPTION` constant you have created here.

packages/node-cli/src/CLI.js
davidecorreu: I don't see it that way, from my point of view we are checking if a given command line argument matches a command name, or some other options.

packages/node-cli/src/CLI.js
karl-anthony-ng: Maybe we can rename it to retainHelpArgumentsOnly?
But I don't have a strong opinion on this.

davidecorreu (approved):

karl-anthony-ng (approved):

PR #332, opened 2021-03-24
feature/pdpdevtool 4086

lfpvillegas:


antoniofruci (approved):

antoniofruci (approved):

packages/vscode-extension/src/suitecloud.ts
```diff
@@ -15,7 +15,7 @@ import { installIfNeeded } from './core/sdksetup/SdkServices';
 import BaseAction from './commands/BaseAction';
 import UpdateObject from './commands/UpdateObject';
 
-const SCLOUD_OUTPUT_CHANNEL_NAME = 'NetSuite SuiteCloud';
+const SCLOUD_OUTPUT_CHANNEL_NAME = 'SuiteCloud Extension';
```
davidecorreu, line 18: Other extension that have their own output console don't use the `Extension` word to name their one, like: GitLens, Git, GithubAuthentication.
Maybe we could name it just `SuiteCloud`

lfpvillegas: Ok I would do it then, because there was a questoin I made to David Flett

davidecorreu (approved):

PR #337, opened 2021-04-01
feature/pdpdevtool-4067: acccept f as non-production domain

davidecorreu:


packages/node-cli/src/commands/account/savetoken/AccountSaveTokenAction.js
```diff
@@ -7,7 +7,7 @@
 const path = require('path');
 const BaseAction = require('../../base/BaseAction');
 const { saveToken } = require('../../../utils/AuthenticationUtils');
-const { PROD_ENVIRONMENT_ADDRESS } = require('../../../ApplicationConstants');
+const { DOMAIN } = require('../../../ApplicationConstants');
```
lfpvillegas, line 10: Could we import only the message used? DOMAIN.PRODUCTION.GENERIC_NETSUITE_DOMAIN

lfpvillegas (changes requested):

packages/node-cli/src/commands/account/savetoken/AccountSaveTokenAction.js
davidecorreu: Sure!

lfpvillegas (approved):

antoniofruci (approved):

PR #342, opened 2021-04-21
Add List Files command to VsCode

lfpvillegas:


packages/node-cli/src/utils/FolderUtils.js
```diff
@@ -0,0 +1,63 @@
+/*
```
karl-anthony-ng, line 1: Since FolderUtils contain code mainly for File Cabinet related process, I was thinking if it would be a good idea to move contents of FolderUtils to FileCabinetService. What do you think?

packages/vscode-extension/src/commands/ListFiles.ts
```diff
@@ -0,0 +1,83 @@
...
+
+		if (selectedFolder === undefined) {
+			return;
+		}
+
+		await this._listFiles(selectedFolder);
+	}
+
+	protected async _getListFolders() {
+		const listFoldersPromise = FolderUtils.getFileCabinetFolders(this.executionPath, COMMAND_NAME);
```
karl-anthony-ng, line 42: If I understand this correctly, listFolderPromise contains the list of folder from the FolderUtils.getFileCabinetFolders call. I think we should rename this variable to remove the "Promise" suffix. Maybe something like folderList.

packages/vscode-extension/src/commands/ListFiles.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export default class ListFiles extends BaseAction {
+	constructor() {
+		super(COMMAND_NAME);
+	}
+
+	protected async execute(): Promise<void> {
+		const fileCabinetFolders = await this._getListFolders();
+		const filteredFolders = this.filterDisabledFolders(fileCabinetFolders);
+		if (filteredFolders === undefined) {
```
karl-anthony-ng, line 28: Will there be a case where filteredFolders can be null? If there is, it'll be better if the checking here is if (filteredFolders == null).

packages/vscode-extension/src/commands/ListFiles.ts
```diff
@@ -0,0 +1,83 @@
...
+			return;
+		}
+
+		const selectedFolder = await this._selectFolder(filteredFolders);
+
+		if (selectedFolder === undefined) {
```
karl-anthony-ng, line 34: Same as the previous comment.

packages/vscode-extension/src/commands/ListFiles.ts
```diff
@@ -0,0 +1,83 @@
...
+		this.messageService.showStatusBarMessage(statusBarMessage, listFoldersPromise);
+
+		return listFoldersPromise;
+	}
+
+	private filterDisabledFolders(fileCabinetFolders: any) {
+		return fileCabinetFolders.filter((folder: FolderType) => folder.disabled === '');
+	}
+
+	// protected async _selectFolder(filteredFolders: { map: (key: (folder: FolderType) => string | undefined) => string[] }): Promise<string | undefined> {
```
karl-anthony-ng, line 53: Please remove this if not needed.

packages/vscode-extension/src/commands/FolderType.ts
```diff
@@ -0,0 +1,5 @@
+export interface FolderType {
+	name: string | undefined;
```
karl-anthony-ng, line 2: Why do we need to specify undefined here? Can we use optional here instead "?"?

packages/vscode-extension/src/commands/ListFiles.ts
```diff
@@ -0,0 +1,83 @@
...
+	protected async _selectFolder(filteredFolders: {
+		map: (arg0: (folder: FolderType) => string | undefined) => string[];
```
karl-anthony-ng, line 55: Why the need for undefined here?

packages/vscode-extension/src/commands/ListFiles.ts
```diff
@@ -0,0 +1,83 @@
...
+	}): Promise<string | undefined> {
```
karl-anthony-ng, line 56: Why the need for undefined here?

karl-anthony-ng (changes requested):

packages/vscode-extension/src/commands/ListFiles.ts
lfpvillegas: I think it would be only undefined. However I could add the null case... eslint makes me use ===,

packages/vscode-extension/src/commands/ListFiles.ts
lfpvillegas: listFolderPromise is still a Promise... until we call the "await". This is why I´m putting "Promise" in its name

packages/vscode-extension/src/commands/ListFiles.ts
lfpvillegas: Typescript is making me adding it. I suppose you could decide to not select any option.  (like pressing Escape). That´s why it is needed to define the possible output of the promise as string or undefined

packages/node-cli/src/utils/FolderUtils.js
lfpvillegas: Yeah, I thought about that too. However, I have seen that FileCabinetService is more focused in the internal FileCabinet folder, and only used for UploadFile command. I would try to merge them though

lfpvillegas: Finally I have renamed it AccountFileCabinetService, to differentiate from FileCabinetService... because FleCabinetService make me create a constructor with the local path, when this is not needed when listing files from the server.

I´m not very happy with the name though so I could rename it differently if you think in a better option.

packages/vscode-extension/src/commands/FolderType.ts
lfpvillegas: I have deleted it and it´s working fine

packages/node-cli/src/services/AccountFileCabinetService.js
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		if (listFoldersResult.status === SdkOperationResultUtils.STATUS.ERROR) {
+			throw listFoldersResult.errorMessages;
+		}
+		return listFoldersResult;
+	}
+
+	async getFileCabinetFolders(executionPath, commandName) {
+		const sdkPath = SdkProperties.getSdkPath();
```
davidecorreu, line 51: This sdkPath is always going to be the node-cli one. This file is imported to vscode, so it will still use node-cli jar there.

packages/node-cli/src/services/AccountFileCabinetService.js
```diff
@@ -0,0 +1,61 @@
...
+		const sdkExecutor = new SdkExecutor(sdkPath, executionEnvironmentContext);
+		let listFoldersResult = await this._listFolders(sdkExecutor, executionPath, commandName);
+		return listFoldersResult.data.map((folder) => {
+			return {
+				name: folder.path,
+				value: folder.path,
+				disabled: folder.isRestricted ? NodeTranslationService.getMessage(RESTRICTED_FOLDER) : '',
+			};
+		});
+	}
```
davidecorreu, line 58: This transformation done to the data is reasonable for Inquirer but I don't think it should be part service.
I think it will be better to return just the data with no transformation and let the caller of the service decide what to do with it.
I think we do different things with it depending if call is from node-cli o vscode .

packages/vscode-extension/src/commands/ListFiles.ts
```diff
@@ -0,0 +1,86 @@
...
+			COMMAND_NAME
+		);
+		const statusBarMessage = this.translationService.getMessage(LIST_FILES.LOADING_FOLDERS);
+		this.messageService.showStatusBarMessage(statusBarMessage, listFoldersPromise);
+
+		return listFoldersPromise;
+	}
+
+	protected async _selectFolder(filteredFolders: {
+		map: (arg0: (folder: FolderType) => QuickPickItem) => QuickPickItem[];
```
davidecorreu, line 58: This part looks pretty weird to me. What do you accomplish with it?

davidecorreu: I refer to `filteredFolders: { map: (arg0: (folder: FolderType) => QuickPickItem) => QuickPickItem[]; }`
We have decided not to hide the restricted folders, then I think `filteredFolders` doesn't sound good.

packages/node-cli/src/services/AccountFileCabinetService.js
lfpvillegas: I´m already using this data in VsCode. I would leave it as it is.

packages/vscode-extension/src/commands/ListFiles.ts
lfpvillegas: Ok to the change of the name.
I need to specify QuickPickItem here because if not it would infer a string.
I need a QuickPickItem to set the "Restricted Folder" in the detail (it would appeear with special typing font)

lfpvillegas: Sorry, when I said "detail" I wanted to say "description", "detail" is another property inside QuickPickItem that is slightly different to "description"

packages/vscode-extension/src/commands/ListFiles.ts
karl-anthony-ng: I think if (filteredFolders == null) should cover both null and undefined.

lfpvillegas: Yes, I know... but eslint complaints about it. If you want we could discuss it in the daily

karl-anthony-ng: I see. Ok, if you're sure that its' just undefined, then I'm ok with it. But if you think null might happen, please add the null case and use === (so no eslint warning).

packages/vscode-extension/src/commands/ListFiles.ts
```diff
@@ -0,0 +1,86 @@
...
+			COMMAND_NAME
+		);
+		const statusBarMessage = this.translationService.getMessage(LIST_FILES.LOADING_FOLDERS);
+		this.messageService.showStatusBarMessage(statusBarMessage, listFoldersPromise);
+
+		return listFoldersPromise;
+	}
+
+	protected async _selectFolder(folders: {
+		map: (arg0: (folder: FolderType) => QuickPickItem) => QuickPickItem[];
```
davidecorreu, line 58: We could be using `_selectFolder(folders: FolderType[])` instead of the long parameter type definition.

davidecorreu (changes requested):

packages/node-cli/src/services/AccountFileCabinetService.js
```diff
@@ -0,0 +1,62 @@
...
+			throw listFoldersResult.errorMessages;
+		}
+		return listFoldersResult;
+	}
+
+	async getFileCabinetFolders(sdkPath, executionEnvironmentContext, executionPath, commandName) {
+		const sdkExecutor = new SdkExecutor(sdkPath, executionEnvironmentContext);
+		let listFoldersResult = await this._listFolders(sdkExecutor, executionPath, commandName);
+		return listFoldersResult.data;
+		// return listFoldersResult.data.map((folder) => {
```
karl-anthony-ng, line 52: Can we remove this commented code?

karl-anthony-ng (changes requested):

davidecorreu (approved):

karl-anthony-ng (approved):

PR #343, opened 2021-04-21
feature/pdpdevtool-3774: remove request-promise-native library

davidecorreu:
Using ProxyAgent.js to tunnel https/http GET requests via proxy.
Adding SUITECLOUD_PROXY environment variable as first option to get the proxy url.


packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+			})
+			.catch((err) => {
+				req.emit('error', err);
+			});
+	}
+
+	createConnection(options) {
+		return new Promise((resolve, reject) => {
+			const ssl = options.protocol ? options.protocol.toLowerCase() === 'https:' : false;
+			if (ssl && this.options.tunnel === true) {
```
antoniofruci, line 41: Can ssl be undefined here or why are you testing it?

packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+		this.createConnection(options)
+			.then((socket) => {
+				req.onSocket(socket);
+			})
+			.catch((err) => {
+				req.emit('error', err);
+			});
+	}
+
+	createConnection(options) {
```
antoniofruci, line 38: It's just a matter of taste, but I would rename options into parameters/params so it's not confused with the class property

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -41,52 +45,73 @@ class SdkDownloadService {
...
-			removeJarFilesFrom(sdkDirectory)
+		if (!/^http:$|^https:$/.test(downloadUrlProtocol)) {
+			throw new Error('Invalid SDK jar file download url protocol. Only http: or https: are allowed.');
+		}
 
-			const sdkDestinationFile = path.join(sdkDirectory, SdkProperties.getSdkFileName());
-			const file = fs.createWriteStream(sdkDestinationFile);
-			file.write(response.body, 'binary');
-			file.end();
+		const httpx = 'http:'.match(downloadUrlObject.protocol) ? http : https;
```
antoniofruci, line 77: httpClient would be a better name imho

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -41,52 +45,73 @@ class SdkDownloadService {
...
+		} catch (error) {
+			NodeConsoleLogger.error(NodeTranslationService.getMessage(DOWNLOADING_SUITECLOUD_SDK_ERROR, fullURL, unwrapExceptionMessage(error)));
 		}
 	}
 
-	_downloadFile(url, sdkDirectory) {
-		const proxy = process.env.npm_config_https_proxy || process.env.npm_config_proxy;
-		const isProxyRequired = proxy && !SdkProperties.configFileExists();
-		const removeJarFilesFrom = this._removeJarFilesFrom;
+	downloadJarFilePromise(downloadUrl, destinationFilePath, proxy, skipProxy) {
```
lfpvillegas, line 63: Could we maintain it as private?

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -41,52 +45,73 @@ class SdkDownloadService {
 		const sdkDirectory = this._fileSystemService.createFolder(sdkParentDirectory, FOLDERS.NODE_CLI);
 
 		const fullURL = `${SdkProperties.getDownloadURL()}/${SdkProperties.getSdkFileName()}`;
+		const destinationFilePath = path.join(sdkDirectory, SdkProperties.getSdkFileName());
```
lfpvillegas, line 48: I think this 3 variables are only used inside downloadJarFilePromise. 
Could we define them there? Or do you think it is better to define it here.

In special proxy and skipProxy, because if you don´t pass destinationFilePath you would have to pass sdkDirectory, so it´s fine with it.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -41,52 +45,73 @@ class SdkDownloadService {
...
 
-		return request(options).then(function(response) {
-			if (!VALID_JAR_CONTENT_TYPES.includes(response.headers['content-type'])) {
-				throw NodeTranslationService.getMessage(DOWNLOADING_SUITECLOUD_SDK_ERROR_FILE_NOT_AVAILABLE);
-			}
-
-			// remove all jar files before writing response to file
-			removeJarFilesFrom(sdkDirectory)
+		if (!/^http:$|^https:$/.test(downloadUrlProtocol)) {
+			throw new Error('Invalid SDK jar file download url protocol. Only http: or https: are allowed.');
```
lfpvillegas, line 74: Could we put this message in messages.json? Also, has it been reviewed by Technical writer?

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
lfpvillegas, line 77: What does it mean httpx? that it could be http or https?  could we named instead request?

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -41,52 +45,73 @@ class SdkDownloadService {
...
+				});
+			});
+
+			clientReq
+				.once('timeout', () => {
+					clientReq.destroy();
+					reject(new Error(`GET request timeout.`));
+				})
+				.once('error', (err) => reject(err))
+				.end();
```
lfpvillegas, line 106: What do you think about moving this whole Promise to a new method?

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -41,52 +45,73 @@ class SdkDownloadService {
...
-			const sdkDestinationFile = path.join(sdkDirectory, SdkProperties.getSdkFileName());
-			const file = fs.createWriteStream(sdkDestinationFile);
-			file.write(response.body, 'binary');
-			file.end();
+		const httpx = 'http:'.match(downloadUrlObject.protocol) ? http : https;
+
+		return new Promise((resolve, reject) => {
+			const clientReq = httpx.get(downloadUrlObject, requestOptions, (response) => {
+				const chunks = [];
+				response.on('data', (chunk) => chunks.push(Buffer.from(chunk, 'binary')));
```
lfpvillegas, line 82: We could move 'data' 'binary and ' content-type' to constants if possible.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -41,52 +45,73 @@ class SdkDownloadService {
...
+
+				response.on('error', (error) => {
+					reject(NodeTranslationService.getMessage(DOWNLOADING_SUITECLOUD_SDK_ERROR_FILE_NOT_AVAILABLE));
+				});
+			});
+
+			clientReq
+				.once('timeout', () => {
+					clientReq.destroy();
+					reject(new Error(`GET request timeout.`));
```
lfpvillegas, line 103: Move to message

packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+					req.destroy();
+					reject(new Error('ProxyAgent CONNECT request timeout.'));
+				});
+
+				req.once('error', (err) => {
+					reject(err);
+				});
+
+				req.once('close', () => {
+					reject(new Error('Tunnel failed. Socket closed prematurely'));
```
lfpvillegas, line 76: Move to message and get review from tech writer

packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+						host: options.hostname || options.host,
+						port: +options.port,
+						servername: options.servername || options.host,
+					});
+					resolve(tunnel);
+				});
+
+				req.once('timeout', () => {
+					req.destroy();
+					reject(new Error('ProxyAgent CONNECT request timeout.'));
```
lfpvillegas, line 68: Move to message and get review from tech writer

packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+				req.onSocket(socket);
+			})
+			.catch((err) => {
+				req.emit('error', err);
+			});
+	}
+
+	createConnection(options) {
+		return new Promise((resolve, reject) => {
+			const ssl = options.protocol ? options.protocol.toLowerCase() === 'https:' : false;
```
lfpvillegas, line 40: https move to constant

packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+			if (ssl && this.options.tunnel === true) {
+				if (options.port === 80) options.port = 443;
+				// CONNECT Method
+				const req = http.request({
+					host: this.proxyURL.hostname,
+					port: this.proxyURL.port,
+					auth: this.proxyURL.auth,
+					method: 'CONNECT',
+					path: (options.hostname || options.host) + ':' + options.port,
```
lfpvillegas, line 49: : Move to constant

packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+					headers: {
+						host: options.host,
+					},
+					timeout: this.options.timeout,
+				});
+
+				req.once('connect', (res, socket, head) => {
```
lfpvillegas, line 56: connect move to constant

packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+					const tunnel = tls.connect({
+						socket: socket,
+						host: options.hostname || options.host,
+						port: +options.port,
+						servername: options.servername || options.host,
+					});
+					resolve(tunnel);
+				});
+
+				req.once('timeout', () => {
```
lfpvillegas, line 66: timeout move to constant

packages/node-cli/src/core/sdksetup/ProxyAgent.js
```diff
@@ -0,0 +1,92 @@
...
+
+	addRequest(req, options) {
+		req.shouldKeepAlive = false;
+
+		this.createConnection(options)
+			.then((socket) => {
+				req.onSocket(socket);
+			})
+			.catch((err) => {
+				req.emit('error', err);
```
lfpvillegas, line 34: error move to constant

lfpvillegas (changes requested):

packages/node-cli/src/core/sdksetup/ProxyAgent.js
davidecorreu: We only want to use the tls/ssl when donwload url is on https: . That's what we check here. I'll rename the variable name to make it more readable.

packages/node-cli/src/core/sdksetup/ProxyAgent.js
davidecorreu: I like the suggestion :-) . I'm renaming the class options to agentOptions , which will accomplish the same goal.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
davidecorreu: I was making it public because I was thinking in reusing this method from vscode-extension. But that's not clear, so I'll prefix it again with the `_` .

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
davidecorreu: I want to make the `downloadJarFilePromise` method more "clean", and to take care only of downloading the jar file or throw/reject when it's not possible.
I think that the logic to get the info about the  proxy, destinationFile and downloadUrl shouldn't be part of the download function/service.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
davidecorreu: It's a kind of an internal message, because the end users shouldn't be touching the download urls, but I can move it to the messages. I'll add it to the story description so it could be reviewed.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
davidecorreu: Similar to Antonio's question/answer, this is not yet a client/server or a the request. It's just the appropriate "library" that could be `http` or `https`.
But clearly we'll need to find a better naming or strategy, because both of you have doubts about it 🤣

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
davidecorreu: Not sure about it, because this promise is using 4 variables of the context (httpx, downloadUrlObject, requestOptions and destinationFilePath) and we will need to forward them as arguments. But everything can be done.

packages/node-cli/src/core/sdksetup/ProxyAgent.js
davidecorreu: Reviewed.

lfpvillegas (approved):

antoniofruci (approved):

PR #347, opened 2021-05-06
Refactorization. Delete developmentMode 

lfpvillegas:


packages/node-cli/src/commands/account/savetoken/AccountSaveTokenAction.js
```diff
@@ -73,11 +72,8 @@ module.exports = class AccountSaveTokenAction extends BaseAction {
 		if (validationErrors.length > 0) {
 			throw new CLIException(ValidationErrorsFormatter.formatErrors(validationErrors));
 		}
-		// If url is system.netsuite.com, we must not pass it to the sdk. If it's anything else, we must add developmentMode flag
 		if (params[COMMAND.OPTIONS.URL] === GENERIC_NETSUITE_DOMAIN) {
 			delete params[COMMAND.OPTIONS.URL];
```
davidecorreu, line 76: Is it ok to pass an account specific url, will it work the same way?

packages/node-cli/src/commands/account/setup/SetupInputHandler.js
```diff
@@ -93,14 +95,17 @@ module.exports = class SetupInputHandler extends BaseInputHandler {
...
-				const isDevLabel = authentication.developmentMode
-					? NodeTranslationService.getMessage(QUESTIONS_CHOICES.SELECT_AUTHID.EXISTING_AUTH_ID_DEV_URL, authentication.urls.app)
-					: '';
-				const accountInfo = `${authentication.accountInfo.companyName} [${authentication.accountInfo.roleName}]`;
+				const accountCredentials = authIDActionResult.data[authID];
+				const isNotProductionLabel =
+					!accountCredentials.urls &&
+					!accountCredentials.urls.app.match(ApplicationConstants.DOMAIN.PRODUCTION.PRODUCTION_DOMAIN_REGEX) &&
+					!accountCredentials.urls.app.match(ApplicationConstants.DOMAIN.PRODUCTION.PRODUCTION_ACCOUNT_SPECIFIC_DOMAIN_REGEX)
+						? NodeTranslationService.getMessage(QUESTIONS_CHOICES.SELECT_AUTHID.EXISTING_AUTH_ID_URL_NOT_PRODUCTION, accountCredentials.urls.app)
```
davidecorreu, line 103: Can we extract this check for the ternary operator to a boolean `isNotProductionUrl` to improve readability. I would also rename the label to `notProductionLabel` because I think is better to use the `isVerbPrefix` for boolean values.

packages/node-cli/src/utils/AccountCredentialsFormatter.js
```diff
@@ -42,15 +42,18 @@ function _getAccountType(accountId) {
...
 function getListItemString(authID, accountCredential) {
-	const isDevLabel = accountCredential.developmentMode
-		? NodeTranslationService.getMessage(QUESTIONS_CHOICES.SELECT_AUTHID.EXISTING_AUTH_ID_DEV_URL, accountCredential.urls.app)
-		: '';
+	const isNotProductionLabel =
+		accountCredential.urls &&
+		!accountCredential.urls.app.match(ApplicationConstants.DOMAIN.PRODUCTION.PRODUCTION_DOMAIN_REGEX) &&
+		!accountCredential.urls.app.match(ApplicationConstants.DOMAIN.PRODUCTION.PRODUCTION_ACCOUNT_SPECIFIC_DOMAIN_REGEX)
+			? NodeTranslationService.getMessage(QUESTIONS_CHOICES.SELECT_AUTHID.EXISTING_AUTH_ID_URL_NOT_PRODUCTION, accountCredential.urls.app)
+			: '';
```
davidecorreu, line 50: I would apply here the same suggestions commented in SetupInputHandler.js

packages/node-cli/src/commands/account/savetoken/AccountSaveTokenAction.js
lfpvillegas: Yes, because the params dev was only used in AuthenticationUtils  for adding the dev flag to the Java CLI command.
Since we have deleted the need of this flag we could delete it also here.

antoniofruci (approved):

packages/node-cli/src/commands/account/savetoken/AccountSaveTokenAction.js
davidecorreu: Sorry I think I haven't been clear in my question, which I'll try to make it again. Will the account:savetoken work properly providing the tokenid/tokensecret and a production account specific url?
I guess yes, but just I wanted your confirmation.

davidecorreu (approved):

PR #371, opened 2021-06-15
Feature/pdpdevtool 4023

lfpvillegas:


packages/vscode-extension/package.json
```diff
@@ -44,6 +44,16 @@
...
 				"category": "SuiteCloud"
 			},
+			{
+				"command": "suitecloud.importfiles",
+				"title": "Import Files",
+				"category": "SuiteCloud"
+			},
+			{
+				"command": "suitecloud.importfile",
+				"title": "Import Single File to Account",
```
davidecorreu, line 54: This seems not right, I think it should be `Import Single File from Account` . Probably this one and plural form should be reviewed by @dforacle .

packages/vscode-extension/src/suitecloud.ts
```diff
@@ -32,6 +40,8 @@ export async function activate(context: vscode.ExtensionContext) {
 	context.subscriptions.push(
 		register('suitecloud.adddependencies', new AddDependencies()),
 		register('suitecloud.deploy', new Deploy()),
+		register('suitecloud.importfile', new ImportFiles()),
+		register('suitecloud.importfiles', new ImportFiles()),
```
davidecorreu, line 44: Is it intentional to use the same `new ImportFiles()` for both commands?

packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -72,8 +76,11 @@ export default abstract class BaseAction {
 		return cliConfigurationService.getProjectFolder(this.cliCommandName);
 	}
 
-	public async run() {
-		this.init();
+	public async run(fsPath?: string) {
+		this.init(fsPath);
+		if(fsPath) {
+			this.filePath = fsPath;
```
karl-anthony-ng, line 82: Will there be a case where filePath will be undefined?

packages/vscode-extension/src/commands/ImportFiles.ts
```diff
@@ -0,0 +1,141 @@
...
+import ListFilesService from '../service/ListFilesService';
+import { ANSWERS, COMMAND, ERRORS, IMPORT_FILES } from '../service/TranslationKeys';
+import { FolderItem } from '../types/FolderItem';
+import { actionResultStatus } from '../util/ExtensionUtil';
+import BaseAction from './BaseAction';
+
+const COMMAND_NAME = 'importfiles';
+
+export default class ImportFiles extends BaseAction {
+	protected filePath: string | undefined;
```
karl-anthony-ng, line 18: Why do we need filePath here when there's already a filePath in BaseAction?

packages/vscode-extension/src/suitecloud.ts
lfpvillegas: Yes, this was the refactorization I have made yesterday. There was code duplicated, because logic is the same.
The only thing is that for importfiles, we have a selection of the folders and the files, while in importfile we have already selected the single file to upload

packages/vscode-extension/src/commands/ImportFiles.ts
```diff
@@ -0,0 +1,141 @@
+/*
+ ** Copyright (c) 2020 Oracle and/or its affiliates.  All rights reserved.
+ ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+ */
+import * as path from 'path';
+import * as vscode from 'vscode';
```
karl-anthony-ng, line 6: I think we can remove this since the line below imports specific items from vscode.

packages/vscode-extension/src/service/ImportFileService.ts
```diff
@@ -0,0 +1,52 @@
+import { VSTranslationService } from './VSTranslationService';
```
karl-anthony-ng, line 1: Please add copywrite.

packages/vscode-extension/src/service/ListFilesService.ts
```diff
@@ -0,0 +1,127 @@
+import * as vscode from 'vscode';
```
karl-anthony-ng, line 1: Please add copywrite.

packages/vscode-extension/src/service/ListFilesService.ts
```diff
@@ -0,0 +1,127 @@
...
+import { VSTranslationService } from './VSTranslationService';
+
+const LIST_FILES_COMMAND = {
+	OPTIONS: {
+		FOLDER: 'folder',
+	},
+};
+
+export default class ListFilesService {
+	protected readonly translationService: VSTranslationService;
```
karl-anthony-ng, line 19: Can the class attributes be private instead of protected?

packages/vscode-extension/src/service/ImportFileService.ts
```diff
@@ -0,0 +1,52 @@
...
+import SuiteCloudRunner from '../core/SuiteCloudRunner';
+import MessageService from './MessageService';
+import VSConsoleLogger from '../loggers/VSConsoleLogger';
+
+const COMMAND_NAME = 'file:import';
+
+export default class ImportFileService {
+	protected readonly translationService: VSTranslationService;
```
karl-anthony-ng, line 9: Can the class attributes be private instead of protected?

packages/vscode-extension/src/service/ListFilesService.ts
```diff
@@ -0,0 +1,127 @@
...
+		if (actionResult.status === actionResultStatus.SUCCESS) {
+			return actionResult.data;
+		} else {
+			throw actionResult.errorMessages;
+		}
+	}
+
+	private async listFilesCommand(listFilesOption: { [key: string]: string }) {
+		if (!this.vsConsoleLogger) {
+			throw Error('Logger not initialized');
```
karl-anthony-ng, line 112: Should the message be in messages.properties or in a constant?

packages/vscode-extension/src/service/ImportFileService.ts
```diff
@@ -0,0 +1,52 @@
...
+		}
+
+		const commandActionPromise = this.runSuiteCloudCommand(commandArgs);
+		this.messageService.showStatusBarMessage(statusBarMessage, true, commandActionPromise);
+		return await commandActionPromise;
+	}
+
+	protected async runSuiteCloudCommand(args: { [key: string]: string } = {}) {
+		if (!this.vsConsoleLogger) {
+			throw Error('Logger not initialized');
```
karl-anthony-ng, line 41: Should this be in messages.properties or in a constant?

karl-anthony-ng (changes requested):

packages/vscode-extension/src/suitecloud.ts
davidecorreu: Thanks for removing the obsolete import of ImportFiles.

packages/vscode-extension/src/suitecloud.ts
```diff
@@ -19,10 +20,16 @@ import { installIfNeeded } from './core/sdksetup/SdkServices';
...
 
 function register<T extends BaseAction>(command: string, action: T) {
-	return vscode.commands.registerCommand(command, () => action.run());
+	return vscode.commands.registerCommand(command, (uri?: vscode.Uri) => {
+		if (uri && uri.fsPath) {
+			action.run(uri.fsPath);
+		} else {
+			action.run();
+		}
+	});
```
davidecorreu, line 29: This is a question more than a comment. Who is calling this callback? I mean who is the caller of function of the second argument of the reisterCommand, or when is it actually executed and what are the possible arguments passed there.
I guess you can get from there some info for ImportFiles but I'm a little bit lost here.

packages/vscode-extension/src/commands/BaseAction.ts
lfpvillegas: Yes, when we are calling importfiles from the command palette.
When calling from the context menu, fsPath would not be undefined.

Anyways, I'm deleting this and adding some logic to the init related to this to make it clear.

packages/vscode-extension/src/service/ListFilesService.ts
lfpvillegas: Yes, you have reason. I put this in a constant since it's an internal message that user is not going to see.
Also, I don't think it's needed revision from David Flett.
 What do you think?

packages/vscode-extension/src/suitecloud.ts
lfpvillegas: Yeah, I understand it is not clear. I'm adding a comment to it.
Basically, when you are calling an import from a context menu, it comes with an uri, with the info of the file you are right-clicking.
Meanwhile... if the command is called from the command palette.. it doesn't receive any uri. 
I put this if-else to avoid NullPointer excpetions with uri while only passing the fspath

packages/vscode-extension/package.json
lfpvillegas: I have fixed but let's wait for David Flett comment.

packages/vscode-extension/src/service/ListFilesService.ts
karl-anthony-ng: I agree, no need for David Flett to review it.

karl-anthony-ng (approved):

packages/vscode-extension/package.json
lfpvillegas: Ok, finally I have put the message from David. I would resolve this conversation then.

packages/vscode-extension/src/commands/ImportFiles.ts
```diff
@@ -0,0 +1,139 @@
...
+
+		if (!override || override === this.translationService.getMessage(ANSWERS.NO)) {
+			this.messageService.showInformationMessage(this.translationService.getMessage(IMPORT_FILES.PROCESS_CANCELED));
+			return;
+		}
+
+		const destinationFolder = this.executionPath
+			? path
+					.dirname(this.activeFile)
+					.split(this.executionPath + '\\src')[1]
```
davidecorreu, line 72: This is making the code not compatible with old structure projects. We can use `this.getProjectFolderPath()` from BaseAction to get the projectFolder.

packages/vscode-extension/src/commands/ImportFiles.ts
```diff
@@ -0,0 +1,139 @@
...
+
+			if (!selectedFiles) {
+				this.messageService.showInformationMessage(this.translationService.getMessage(IMPORT_FILES.PROCESS_CANCELED));
+				return;
+			}
+			return selectedFiles.map((file) => file.label.replace(/\\/g, '/'));
+		} else {
+			if (this.activeFile) {
+				const filePath = this.executionPath
+					? this.activeFile.split(this.executionPath + '\\src\\FileCabinet')[1].replace(/\\/g, '/')
```
davidecorreu, line 106: Similar to the previous comment use `this.getProjectFolderPath()`  `FileCabinet` is already on node-cli ApplicationConstants, which is also imported through ExtensionUtils.ts.

packages/vscode-extension/package.json
```diff
@@ -93,7 +103,16 @@
 				"key": "alt+shift+u",
 				"when": "resource =~ //FileCabinet/(SuiteScripts|Templates/(E-mail%20Templates|Marketing%20Templates)|Web%20Site%20Hosting%20Files|SuiteApps)/.+$/"
 			}
-		]
+		],
+		"menus": {
+			"explorer/context": [
+				{
+					"when": "resource =~ //FileCabinet/(SuiteScripts|Templates/(E-mail%20Templates|Marketing%20Templates)|Web%20Site%20Hosting%20Files|SuiteApps)/.+$/",
```
davidecorreu, line 110: This "regex" will match a child directory in `FileCabinet/SuiteScript/ChildDir` and by the name of the command I think it shouldn't.

lfpvillegas: Hey, David, I don't understand. Import files should allow only importing files from Suitescript Templates and WebSiteHostingFiles folders, and all of their childs... I don't see the problem in this regex.

packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -11,14 +11,17 @@ import { CLIConfigurationService, getRootProjectFolder } from '../util/Extension
 import { ERRORS } from '../service/TranslationKeys';
 import { commandsInfoMap } from '../commandsMap';
 import { assert } from 'console';
+import { window } from 'vscode';
 
 export default abstract class BaseAction {
 	protected readonly translationService: VSTranslationService;
+	protected isFileSelected: boolean | undefined;
```
davidecorreu, line 18: Can we use `protected isFileSelected?: boolean;` notation which seems equivalent and more compact, or is there a difference? I would say the same for `activeFile` property.
Or for `isFileSelected' we could set it initially to `false` so it will always be a boolean.

packages/vscode-extension/src/commands/ImportFiles.ts
lfpvillegas: Fixed

antoniofruci (approved):

karl-anthony-ng (approved):

PR #374, opened 2021-06-28
PDPDEVTOOL-4021: As a SuiteCloud Developer, I want to have the option to Create SuiteCloud Project in VsCode, so I can build from scratch a new Project using this Code Editor

karl-anthony-ng:


lfpvillegas:
Could you please change the name of the pull request to the story?

antoniofruci (approved):

packages/vscode-extension/src/commands/CreateProject.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    protected async execute(): Promise<void> {
+        const commandArgs = await this.getCommandArgs();
+        if (Object.keys(commandArgs).length === 0) {
+            return;
+        }
+
+        const actionResult = await this.runSuiteCloudCommand(commandArgs);
+        if (actionResult.isSuccess()) {
+            await CreateProject.openProjectInNewWindow(actionResult.projectDirectory);
```
lfpvillegas, line 40: The story says that there should be a question asking about atttaching the project to the current window or open a new one, but I think is not implemented.

packages/vscode-extension/src/commands/CreateProject.ts
```diff
@@ -0,0 +1,229 @@
...
+            if (!overwriteProject) {
+                return {};
+            }
+            commandArgs[COMMAND_ARGUMENTS.OVERWRITE] = overwriteProject === this.translationService.getMessage(ANSWERS.NO) ? 'false' : 'true'; 
+        }
+
+        return commandArgs;
+    }
+
+    private showOverwriteFilesQuestion(projectAbsolutePath: string): Thenable<string | undefined> {
```
lfpvillegas, line 127: In this method you are not only showing the questions but actually getting the result (more exactly the promise, that when using "awaits")
I would name it getOverwriteFiles directly or if you prefer to mantain the "Question" suffix then I would go for "promptOverwriteFilesQuestion" that have the extraction of the response implicit

lfpvillegas (changes requested):

packages/vscode-extension/src/commands/CreateProject.ts
karl-anthony-ng: You are correct there. I had to clarify first with the team before I implemented it. I will be adding that logic here and I'll let you know when I've added it.

packages/vscode-extension/src/commands/CreateProject.ts
karl-anthony-ng: I updated it and changed itto prompt... as you suggested.

packages/vscode-extension/src/commands/CreateProject.ts
karl-anthony-ng: Updated to include the logic I mentioned.

lfpvillegas (approved):

antoniofruci (approved):

PR #377, opened 2021-07-02
Feature/pdpdevtool 4291

lfpvillegas:


karl-anthony-ng (approved):

karl-anthony-ng (changes requested):
Please check if node-cli/src/templates/package.json.template also needs to be updated for jest 3rd party library.

karl-anthony-ng (approved):

antoniofruci (approved):

PR #379, opened 2021-07-06
Bug fixes for PDPDEVTOOL-4021: As a SuiteCloud Developer, I want to have the option to Create SuiteCloud Project in VsCode, so I can build from scratch a new Project using this Code Editor

karl-anthony-ng:
- PDPDVETOOL-4309: Project create requires an active file from current project
- PDPDEVTOOL-4310: setup account for new account

### DIFF
--- a/packages/vscode-extension/src/commands/CreateProject.ts
+++ b/packages/vscode-extension/src/commands/CreateProject.ts
@@ -42,6 +42,12 @@ export default class CreateProject extends BaseAction {
         }
     }
 
+    protected validate(): { valid: false; message: string } | { valid: true } {
+        return {
+            valid: true,
+        };
+	}
+
     private async openProjectInNewWindow(projectAbsolutePath: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/commands/CreateProject.ts
```diff
@@ -42,6 +42,12 @@ export default class CreateProject extends BaseAction {
         }
     }
 
+    protected validate(): { valid: false; message: string } | { valid: true } {
```
antoniofruci, line 45: Is it mandatory to keep this function since it always returns true?

karl-anthony-ng: Yes. I just overridden itfrom BaseAction so that there will not be a need to check if there's an active file opened to trigger the command.

lfpvillegas (approved):

antoniofruci (approved):

PR #380, opened 2021-07-06
bugfix: fixed string template reference in AccountCredentialsFormatte…

antoniofruci:


karl-anthony-ng (approved):

packages/node-cli/src/utils/AccountCredentialsFormatter.js
```diff
@@ -42,22 +42,21 @@ function _getAccountType(accountId) {
...
 }
 
-function getListItemString(authID, accountCredential) {
+function getListItemString(authID, accountCredentials) {
 	const isNotProductionUrl =
-		accountCredential.urls &&
-		!accountCredential.urls.app.match(DOMAIN.PRODUCTION.PRODUCTION_DOMAIN_REGEX) &&
-		!accountCredential.urls.app.match(DOMAIN.PRODUCTION.PRODUCTION_ACCOUNT_SPECIFIC_DOMAIN_REGEX);
+		accountCredentials.urls &&
+		!accountCredentials.urls.app.match(ApplicationConstants.DOMAIN.PRODUCTION.PRODUCTION_DOMAIN_REGEX) &&
```
lfpvillegas, line 48: I think you can delete the previous import DOMAIN related to AccountCredentials (line10)

lfpvillegas (changes requested):

packages/node-cli/src/utils/AccountCredentialsFormatter.js
antoniofruci: I can't. I am using it at line 30. That was the fix :)

lfpvillegas (approved):

PR #387, opened 2021-07-16
feature/pdpdevtool-4022: CreateFile

antoniofruci:


packages/vscode-extension/src/service/TranslationKeys.ts
```diff
@@ -173,3 +173,18 @@ export const UPDATE_OBJECT = {
 	PROCESS_CANCELED: 'UPDATE_OBJECT_PROCESS_CANCELED',
 	UPDATING: 'UPDATE_OBJECT_UPDATING',
 };
+
+export const CREATE_FILE = {
+	QUESTIONS: {
+		CHOOSE_SUITESCRIPT_TYPE: 'CREATE_FILE_CHOOSE_SUITESCRIPT_TYPE',
```
lfpvillegas, line 179: Not very important but we are usually following format CREATE_FILE_QUESTIONS_CHOOSE_SUITESCRIPT_TYPE. The same applies for MESSAGES

lfpvillegas (changes requested):

packages/vscode-extension/src/service/TranslationKeys.ts
davidecorreu: Fixed

packages/vscode-extension/messages.json
```diff
@@ -14,6 +14,15 @@
 	"COMMAND_SUCCESS": "'{0}' has been successfully executed.",
 	"COMMAND_TRIGGERED": "'{0}' has been triggered.",
 
+	"CREATE_FILE_ERRORS_MISSING_VALID_FOLDER_FOR_SUITECRIPT_FILE": "No valid folder found to create a SuiteScript file. Use '{0}' from a valid SuiteCloud project structure. For more information about the project structure, see {1}",
```
lfpvillegas, line 17: Maybe some of these sentences needs a final dot.

packages/vscode-extension/src/commands/CreateFile.ts
```diff
@@ -0,0 +1,209 @@
...
+		}
+
+		return {
+			valid: true,
+		};
+	}
+
+	protected async execute(): Promise<void> {
+		const commandArgs = await this.getCommandArgs();
+		if (commandArgs === undefined) {
```
lfpvillegas, line 54: Should we show a message warning about cancellation of the process (when pressing Esc) like it is done in other commands?

packages/vscode-extension/src/commands/CreateFile.ts
```diff
@@ -0,0 +1,209 @@
...
+		}
+
+		args.path = path.join(selectedFolder, fileName);
+
+		return args;
+	}
+
+	private promptScriptTypeQuestion(): Thenable<SuiteScriptTypeItem | undefined> {
+		return window.showQuickPick(
+			SUITESCRIPT_TYPES.map((el) => <SuiteScriptTypeItem>{ label: el.name, id: el.id, name: el.name }),
```
lfpvillegas, line 114: could we give e1 a more informative name?

packages/vscode-extension/src/commands/CreateFile.ts
```diff
@@ -0,0 +1,209 @@
...
+				return filteredFolderChoices[0];
+			}
+			return window.showQuickPick(filteredFolderChoices, {
+				placeHolder: this.translationService.getMessage(CREATE_FILE.QUESTIONS.SELECT_FOLDER),
+				canPickMany: false,
+			});
+		}
+
+		// action not originated from context menu
+		return window.showQuickPick(folderChoices, {
```
lfpvillegas, line 158: This lines are repeated from previous lines 151-154

lfpvillegas (changes requested):

packages/vscode-extension/messages.json
davidecorreu: The final argument is web link.
I have somehow reutilized the one from ```"MANAGE_ACCOUNTS_ERROR_MISSING_MANIFEST": "No manifest.xml file was found in {0}. Use '{1}' from an active file in a valid project folder. For more information about the project structure, see {2}",``` and it has not dot at the end neither.

packages/vscode-extension/src/commands/CreateFile.ts
davidecorreu: I don't think so. Probably we show a cancellation message when respond NO to the "The objects in your project will be overwritten, do you want to continue" type of question, but if we hit ESCAPE to cancel the "Setup Account" or the "Create Project" no message is shown.

packages/vscode-extension/src/commands/CreateFile.ts
davidecorreu: I think SUITESCRIPT_TYPES is descriptive enough, it could also be SUITESCRIPT_TYPES_METADATA but I'm not sure.
What would be your suggestion?

packages/vscode-extension/src/commands/CreateFile.ts
davidecorreu: They are not the same, one is using `filteredFolderChoices` and the other one `folderChoices`.

packages/vscode-extension/src/commands/CreateFile.ts
lfpvillegas: I'm refering to the el variable inside

packages/vscode-extension/src/commands/CreateFile.ts
lfpvillegas: It is implemented in other commands. I think we should be consistent.

karl-anthony-ng: I agree with David that there's no need to show a cancel message for this case.

I wasn't aware though that we implemented this in other commands.

packages/vscode-extension/src/commands/CreateFile.ts
```diff
@@ -0,0 +1,209 @@
...
+		);
+	}
+
+	private async promptFolderSelection(): Promise<string | undefined> {
+		const folderChoices = this.getFolderChoices();
+
+		let fileToCheck = this.activeFile;
+
+		// action orignated from context menu
+		if (this.isFileSelected && fileToCheck && fs.existsSync(fileToCheck)) {
```
karl-anthony-ng, line 138: Should we have something like a service for that can check if action originated from the context menu?

karl-anthony-ng (changes requested):
Please check my comments.

packages/vscode-extension/src/commands/CreateFile.ts
lfpvillegas: you could overwrite folderChoices instead of creating new variable filteredFolderChoices.
But anyways, if you think is more readable this way, it's ok

packages/vscode-extension/src/commands/CreateFile.ts
davidecorreu: I'm just using `this.isFileSelected` to know if the command was triggered from the context menu. The other conditions are just sanity checks.
Maybe we can rename `this.isFileSelected` -> `this.uriAvailable` or `this.fromContextMenu`. What do you think? I'm open to other suggestions.

karl-anthony-ng: I was looking at the comment and I thought the whole if statement was for checking if the file was selected from the context menu. So yes, maybe rename this.isFileSelected to this.fromContextMenu if isFileSelected is just being used for checking if the file has been selected from the context menu.

lfpvillegas (approved):

packages/vscode-extension/src/commands/CreateFile.ts
davidecorreu: I'll check its usages, and rename accordingly.

davidecorreu: Renamed to `this.isSelectedFromContextMenu`

karl-anthony-ng (approved):

PR #388, opened 2021-07-21
Bug Fix - PDPDEVTOOL-4321, PDPDEVTOOL-4323

karl-anthony-ng:
Bugs Fixed:
- PDPDEVTOOL-4321: Can't create project when no workspace is open
- PDPDEVTOOL-4323: Wrong Message

### DIFF
--- a/packages/vscode-extension/src/commands/BaseAction.ts
+++ b/packages/vscode-extension/src/commands/BaseAction.ts
@@ -54,8 +54,13 @@ export default abstract class BaseAction {
 		}
 	}
 
-	protected async runSuiteCloudCommand(args: { [key: string]: string | string[] } = {}) {
-		const suiteCloudRunnerRunResult = await new SuiteCloudRunner(this.vsConsoleLogger, this.executionPath).run({
+	protected async runSuiteCloudCommand(args: { [key: string]: string | string[] } = {}, executionPath?: string) {
+		let executionPathToUse = this.executionPath;
+		if (executionPath !== undefined) {
+			executionPathToUse = executionPath;
+		}
+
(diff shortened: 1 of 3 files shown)


packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -54,8 +54,13 @@ export default abstract class BaseAction {
 		}
 	}
 
-	protected async runSuiteCloudCommand(args: { [key: string]: string | string[] } = {}) {
-		const suiteCloudRunnerRunResult = await new SuiteCloudRunner(this.vsConsoleLogger, this.executionPath).run({
+	protected async runSuiteCloudCommand(args: { [key: string]: string | string[] } = {}, executionPath?: string) {
+		let executionPathToUse = this.executionPath;
+		if (executionPath !== undefined) {
+			executionPathToUse = executionPath;
+		}
```
davidecorreu, line 61: I think we can use here a ternary operator to decide `executionPahtToUse`, and I would like to rename the `executionPath` parameter to `externalExecutionPath` or `outerExecutionPath` to make clearer its propose and easier to differentiate from `this.executionPath`.

davidecorreu: Maybe `alternativeExecutionPath` could be an option but I'm not sure about it because it will be used if present, instead of an alternative if `this.excutionPath` is missing.

lfpvillegas (approved):

packages/vscode-extension/src/commands/BaseAction.ts
karl-anthony-ng: Updated to make use of otherExecutionPath.

davidecorreu (approved):

PR #390, opened 2021-07-21
Feature/pdpdevtool 4188

lfpvillegas:


packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+import CustomObjectService from '../service/ImportObjectService';
+import { ANSWERS, ERRORS, IMPORT_OBJECTS } from '../service/TranslationKeys';
+import { actionResultStatus, ApplicationConstants, InteractiveAnswersValidator, ProjectInfoService } from '../util/ExtensionUtil';
+import BaseAction from './BaseAction';
+
+const objectTypes: {
+	name: string;
+	value: { name: string; type: string; prefix: string; hasRelatedFiles: boolean; relatedFiles?: { type: string }[] };
+}[] = require('@oracle/suitecloud-cli/src/metadata/ObjectTypesMetadata');
+
```
davidecorreu, line 17: Can we move this one to ExtensionUtil.ts, I think should centralize there the code we import from node-cli (`@oracle/suitecloud-cli`). Then we should also decide to import as OBJECT_TYPES or in a different format.

packages/vscode-extension/src/commands/ImportFiles.ts
```diff
@@ -45,6 +45,7 @@ export default class ImportFiles extends BaseAction {
 		}
 
 		if (!selectedFilesPaths) {
+			this.messageService.showInformationMessage(this.translationService.getMessage(IMPORT_FILES.PROCESS_CANCELED));
```
davidecorreu, line 48: We will need to review this with the team, to know if we want to show a "cancelation info message" or nothing when the user hits ESCAPE while being asked from the command palette. 
Same for line 95.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+		const destinationFolder = this.getDestinationFolder(this.activeFile);
+		if (!destinationFolder.startsWith(ApplicationConstants.FOLDERS.OBJECTS)) {
+			this.messageService.showErrorMessage(this.translationService.getMessage(IMPORT_OBJECTS.ERROR.INCORRECT_FOLDER));
+			return;
+		}
+
+		let appId: string | undefined;
+		if (projectInfoService.isSuiteAppProject()) {
+			appId = await this.promptAppId(projectInfoService);
+			if (appId == undefined) {
```
karl-anthony-ng, line 49: We should use triple equal signs here (===) instead of just two equal signs (==).

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+				this.messageService.showInformationMessage(this.translationService.getMessage(IMPORT_OBJECTS.PROCESS_CANCELED));
+				return;
+			}
+			if (appId == '') {
```
karl-anthony-ng, line 53: We should use triple equal signs here (===) instead of just two equal signs (==).

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+		const selectedObjectTypes = await this.promptObjectTypes();
+
+		if (!selectedObjectTypes) {
+			this.messageService.showInformationMessage(this.translationService.getMessage(IMPORT_OBJECTS.PROCESS_CANCELED));
+			return;
+		}
+
+		const scriptId = await this.promptSelectedScriptId();
+
+		if (scriptId == undefined) {
```
karl-anthony-ng, line 67: We should use triple equal signs here (===) instead of just two equal signs (==).

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+			placeHolder: this.translationService.getMessage(IMPORT_OBJECTS.QUESTIONS.APP_ID, defaultAppId),
+			validateInput: (fieldValue) => {
+				if (fieldValue === '') {
+					return null; //In this case empty field is valid
+				}
+				let validationResult = InteractiveAnswersValidator.showValidationResults(fieldValue, InteractiveAnswersValidator.validateSuiteApp);
+				return typeof validationResult === 'string' ? validationResult : null;
+			},
+		});
+		if (appId == undefined) {
```
karl-anthony-ng, line 139: We should use triple equal signs here (===) instead of just two equal signs (==).

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+			return;
+		}
+
+		if (appId.length == 0) {
```
karl-anthony-ng, line 143: We should use triple equal signs here (===) instead of just two equal signs (==).

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+				placeHolder: this.translationService.getMessage(IMPORT_OBJECTS.QUESTIONS.FILTER_APP_ID),
+				canPickMany: false,
+			}
+		);
+
+		if (!filterAppId) {
+			return;
+		}
+
+		if (filterAppId && filterAppId == this.translationService.getMessage(ANSWERS.NO)) {
```
karl-anthony-ng, line 123: We should always use === unless usage of == is really needed.

packages/vscode-extension/src/service/ImportObjectService.ts
```diff
@@ -0,0 +1,86 @@
...
+		includeReferencedFiles: boolean,
+		executionPath: string | undefined
+	) {
+		this.executionPath = executionPath;
+		//We choose 'ALL' types because it is chosen before which scriptIds should be imported and not by type
+		let commandArgs: any = { type: 'ALL', destinationfolder: destinationFolder };
+
+		commandArgs.scriptid = scriptIds && scriptIds.length > 0 ? scriptIds : 'ALL';
+
+		if (appId == undefined || appId.length != 0) {
```
karl-anthony-ng, line 41: We should use === instead of == unless there is really a need to use ==. Please check all other == usage in here and change them to ===.

packages/vscode-extension/src/util/ExtensionUtil.ts
```diff
@@ -4,16 +4,24 @@
...
 
-// returns the root project folder of the active file in the editor
+// returns the root project folder of the active file in the editor if uri not defined
+// uri is present when action originated from a contextMenu of the treeView
 // works fine with workspace with multiple project folders opened
-export function getRootProjectFolder(): string | undefined {
-	const activeTextEditor = vscode.window.activeTextEditor;
-	const activeWorkspaceFolder = activeTextEditor ? vscode.workspace.getWorkspaceFolder(activeTextEditor.document.uri) : undefined;
-	return activeWorkspaceFolder ? activeWorkspaceFolder.uri.fsPath : undefined;
+export function getRootProjectFolder(uri?:vscode.Uri): string | undefined {
```
karl-anthony-ng, line 10: Maybe we should move this function to a service class? I think ExtensionUtil might have been originally a place for all utility functions that are for the vscode extension. But now it looks like it's mainly used to reference dependencies coming from NodeCli.

karl-anthony-ng (changes requested):
Please check comments.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	protected async execute() {
+		if (!this.activeFile) {
+			// Already checked in validate
+			return;
+		}
+		const projectInfoService = new ProjectInfoService(this.getProjectFolderPath());
+
+		const destinationFolder = this.getDestinationFolder(this.activeFile);
```
davidecorreu, line 40: Maybe the variable and the method could be referencing: `relativeDestinationFolder` to make it more explicit.

packages/vscode-extension/src/commands/ImportObjects.ts
davidecorreu, line 49: We should use `===`

packages/vscode-extension/src/util/ExtensionUtil.ts
lfpvillegas: I copied this solution from @davidecorreu  story. I would leave it for him to change or I would change after he merge the story, to not have any merge problem.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		const selectedObjectTypes = await this.promptObjectTypes();
+
+		if (!selectedObjectTypes) {
+			this.messageService.showInformationMessage(this.translationService.getMessage(IMPORT_OBJECTS.PROCESS_CANCELED));
+			return;
+		}
+
+		const scriptId = await this.promptSelectedScriptId();
```
davidecorreu, line 63: Variable name could be scriptIdToFilterBy or scriptIdFilter or filterByScriptId. Method name could be promptFilterByScripIdQuestion.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+
+		if (scriptId == undefined) {
+			this.messageService.showInformationMessage(this.translationService.getMessage(IMPORT_OBJECTS.PROCESS_CANCELED));
+			return;
+		}
+
+		const listObjectsResult = await this.listObjects(appId, selectedObjectTypes, scriptId);
```
davidecorreu, line 69: If  we change the reassignment of appId variable to undefined in line 54, the `listObjects` method should only allow string parameters (that could be empty strings) but never undefined, and the same in the service method (`ImportObjectsService.listobjects(...)`). 
Somehow I would like to follow a rule like, if the parameter was undefined the execution should stop, if you don't want to filter by that parameter let's make it's string value equals to an emptyString.
Following that approach we should not call the service with undefined values, and the method should not allow them to be undefined. I think it will make the code more robust.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+	private async promptObjects(objectList: any[]): Promise<string[] | undefined> {
+		const selectedObjects = await window.showQuickPick(
+			objectList.map((object) => object.scriptId),
+			{
+				placeHolder: this.translationService.getMessage(IMPORT_OBJECTS.QUESTIONS.SELECT_OBJECTS),
+				canPickMany: true,
+			}
+		);
+		return selectedObjects;
+	}
```
davidecorreu, line 177: I think we should add a validation that does not allow to continue without selecting at least one object.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+			this.showOutput(listObjectsResult);
+			return;
+		}
+
+		const selectedScriptIds = await this.promptObjects(listObjectsResult.data);
+
+		if (!selectedScriptIds || selectedScriptIds.length == 0) {
+			this.messageService.showCommandError(this.translationService.getMessage(IMPORT_OBJECTS.ERROR.EMPTY_LIST));
+			return;
+		}
```
davidecorreu: I think we should not allow to proceed the execution without selecting at least one object. I have added also added the comment the `this.promptObjects(...)` method.
This will make the unnecessary the `selectedScriptIds.length == 0` part of the condition.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,238 @@
...
+		const includeReferencedFiles = await this.promptIncludeReferencedFiles();
```
davidecorreu, line 81: After this line we should check if `includeReferencedFiles` is undefined and stop the execution if needed.

packages/vscode-extension/src/commands/ImportObjects.ts
lfpvillegas: appId could be undefined if the project is an ACP

packages/vscode-extension/src/commands/ImportObjects.ts
lfpvillegas: Quickpick doesn't allow validation, only Inputboxes.
That's why there is the validation in the other chunk of code.

karl-anthony-ng (approved):

packages/vscode-extension/src/commands/ImportObjects.ts
davidecorreu: If we are on ACP or we don't want to import the objects that belong to an specific suiteApp we can set it's value to an emptyString.

packages/vscode-extension/src/commands/ImportObjects.ts
davidecorreu: We could be asking again and again while the user is not selecting at least one object, that could be a way to force the user to select at least one, or hit ESCAPE to not continue with the process.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,225 @@
...
+					: this.translationService.getMessage(IMPORT_OBJECTS.QUESTIONS.OVERWRITE),
+			canPickMany: false,
+		});
+	}
+
+	private showOutput(actionResult: any) {
+		if (actionResult.status === actionResultStatus.SUCCESS && actionResult.data) {
+			if (actionResult.data.length === 0) {
+				this.messageService.showCommandError(this.translationService.getMessage(IMPORT_OBJECTS.ERROR.EMPTY_LIST_SEARCH));
+				return;
```
davidecorreu, line 200: When we reach this point of execution because there were no objects to list (with the criteria previously selected), we show the `see details` button, but there is nothing in the SuiteCloud output to be shown.

packages/vscode-extension/src/commands/ImportObjects.ts
lfpvillegas: Sorry David, I think this is adding some useless code (setting appid to emptystring) when listObjects could handle undefined appIds as it is doing right now.
I don't see so much benefits on this change.

lfpvillegas: More important I would have to look for a condition in ImportObjectService (Line 55) to ensure that appId length is 0 to not add it to the arguments.
It's more easy to ensure just that is not undefined

davidecorreu: The condition it's already there in the service.
`if (appId) {commandArgs.appid = appId;}`
My point here is that we should ensure we are not passing any parameter with undefined value by error. That's why I would change the listObjects() signature in the ImportObjectService to not accept undefined parameters.

packages/vscode-extension/src/service/ImportObjectService.ts
```diff
@@ -0,0 +1,86 @@
...
+			commandArgs.excludefiles = true;
+		}
+
+		const commandActionPromise = this.runSuiteCloudCommand(commandArgs, IMPORT_OBJECT_COMMAND_NAME);
+		const statusBarMessage = this.translationService.getMessage(IMPORT_OBJECTS.IMPORTING_OBJECTS);
+		this.messageService.showStatusBarMessage(statusBarMessage, true, commandActionPromise);
+		return await commandActionPromise;
+	}
+
+	async listObjects(appId: string | undefined, types: string[], scriptId: string | undefined, executionPath: string | undefined) {
```
davidecorreu, line 55: This actually the signature that I would like to change, to not accept undefined values. I think it will make our code more robust, because the typescript check would avoid undesired calls with parameters unset. Instead we will be at least forced to pass emptyStrings which are more difficult to pass by accident.

lfpvillegas: Again, I don't see any problem in having it undefined. Line 59 is waiting on that. Same for scriptid. 
Only one I see it's worthy to control is executionPath that is already checked in validate()

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,247 @@
...
+				canPickMany: true,
+			}
+		);
+	}
+
+	private async listObjects(appId: string, selectedObjectTypes: string[], scriptId: string) {
+		if (!this.executionPath) {
+			//already  checked in validate. Should not throw
+			throw 'Unexpected error at list objects';
+		}
```
davidecorreu, line 172: I would like to move this check to line 25 in the main execute method.
Then I would directly call this.customObjectService.listObjects() from line 63 and delete the whole listObjects() method.

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -0,0 +1,247 @@
...
+		if (overwrite === this.translationService.getMessage(ANSWERS.NO)) {
+			this.messageService.showInformationMessage(this.translationService.getMessage(IMPORT_OBJECTS.PROCESS_CANCELED));
+			return;
+		}
+
+		if (!this.executionPath) {
+			//already  checked in validate. Should not throw
+			throw 'Unexpected error at list objects';
+		}
```
davidecorreu, line 90: This can be deleted if the comment in lines 169/172 is applied.

davidecorreu (approved):

PR #404, opened 2021-08-09
feature/pdpdevtool-4205: reinstall the JAR if the file is corrupted in vscode

davidecorreu:
Prompt user to restart vscode instance when .jar file is corrupted.
Prompt user to setup an authid when ListFiles/ImportFiles is triggerd from vscode without it.
Refactor node-cli AccountFileCabinetService.js, and improve messages.
Refactor ListFilesService.ts, ListFiles.ts, ImportFiles.ts
Delete ImportFilesService
Improve ActionResult interface for vscode extension, and apply it on SuiteCloudRunner.ts and BaseAction.ts


packages/vscode-extension/messages.json
```diff
@@ -10,6 +10,9 @@
 	"ANSWERS_NO": "No",
 	"ANSWERS_YES": "Yes",
 
+	"BUTTONS_RESTART_NOW": "Restart Now",
```
antoniofruci, line 13: Why "BUTTONS" instead of "BUTTON" ?

packages/vscode-extension/messages.json
```diff
@@ -46,6 +49,7 @@
 
 	"DISMISS": "Dismiss",
 
+	"ERRORS_COURRUPTED_SDK_JAR_DEPENDENCY": "There was a problem with SuiteCloud Extension dependencies. Restart your Visual Studio Code instance.",
```
alii-gator, line 52: Was this reviewed by Montsant?

packages/vscode-extension/src/loggers/VSConsoleLogger.ts
```diff
@@ -35,7 +40,7 @@ export default class VSConsoleLogger extends ConsoleLogger {
 			output.appendLine(this.getExecutionDetails());
 			this._addExecutionDetailsToLog = false;
 		}
-
+		this.checkForCorruptedJar(message);
```
tourajvaziri, line 43: Why are we adding the checkForCorruptedJar in the logger? 
Cannot we move it to another class more suitable, and decouple the logic for logging and checking for corrupted jar? 
they can be call separately.

packages/vscode-extension/src/loggers/VSConsoleLogger.ts
```diff
@@ -54,4 +59,21 @@ export default class VSConsoleLogger extends ConsoleLogger {
 	error(message: string): void {
 		this.println(message);
 	}
+
+	private checkForCorruptedJar(message: string) {
```
tourajvaziri, line 63: Not sure this should be here (?)

packages/vscode-extension/src/service/ListFilesService.ts
```diff
@@ -130,38 +104,52 @@ export default class ListFilesService {
...
 			return actionResult.data;
 		} else {
 			throw actionResult.errorMessages;
 		}
 	}
 
-	private async listFilesCommand(listFilesOption: { [key: string]: string }) {
-		if (!this.vsConsoleLogger) {
-			throw Error(CONSOLE_LOGGER_ERROR);
+	private getDefaultAuthId(): string | undefined {
```
tourajvaziri, line 126: Assuming we will have this logic in different places, (when default id does not exist) can we move this code to a global place where others can use?

packages/vscode-extension/messages.json
davidecorreu: It comes from the TranslationKeys.ts were we have other "groups" of keys, like another one called ANSWERS.

packages/vscode-extension/messages.json
davidecorreu: Yes it was reviewed by Montsant. The capitalized words on the buttons section, was also reviewed by her.

packages/vscode-extension/src/loggers/VSConsoleLogger.ts
davidecorreu: I agree with you, but I didn't find a better way.

packages/vscode-extension/src/service/ListFilesService.ts
davidecorreu: I was initially thinking about it, and I forgot about it later. I'll move it to another place where it will could be easy reused.

antoniofruci (approved):

packages/vscode-extension/src/service/ListFilesService.ts
tourajvaziri: Thanks David.

tourajvaziri (approved):
Looks good.

alii-gator (approved):

PR #408, opened 2021-08-11
Fix the following bugs: PDPDEVTOOL-4362 and PDPDEVTOOL-4363

karl-anthony-ng:
Fixed the following...
- Bug Fixing - PDPDEVTOOL-4362: Creating project fails when a project exists with similar name
- Bug Fixing - PDPDEVTOOL-4363: Project is not opening in new window

In CreateProject.ts, please ignore whitespaces. There was an extra space on every line that I removed.


packages/node-cli/src/validation/InteractiveAnswersValidator.js
```diff
@@ -181,6 +181,14 @@ module.exports = {
 		return !fileSystemService.fileExists(path.join(parentFolderPath, filenameWithExtension))
 			? VALIDATION_RESULT_SUCCESS
 			: VALIDATION_RESULT_FAILURE(NodeTranslationService.getMessage(ANSWERS_VALIDATION_MESSAGES.FILE_ALREADY_EXISTS));
+	},
+
+	validateFolderAlreadyExists(path) {
```
davidecorreu, line 186: I think the name should be, `validateFolderDoesNotExist` .
I have seen also that the previous validate function `validateSuiteScriptFileAlreadyExists`  which I did create is also following the same wrong approach. 
I would say, to at least change the one you have created, and feel free to change the other one if you wish.

packages/node-cli/__test__/validation/InteractiveAnswersValidator.test.js
```diff
@@ -184,3 +185,22 @@ describe('validateXMLCharacters', () => {
 		expect(validateXMLCharacters("mixing_Some > of them < all 'in' the &same string \"")).toEqual(failureResponse);
 	});
 });
+
+describe('validateFolderAlreadyExists', () => {
+	it('should return true when folder does not exists', () => {
```
davidecorreu, line 190: Check comment on the InteractiveAnswersValidator

PR #410, opened 2021-08-12
Fix the following bugs: PDPDEVTOOL-4362 and PDPDEVTOOL-4363

karl-anthony-ng:
Fixed the following...

Bug Fixing - PDPDEVTOOL-4362: Creating project fails when a project exists with similar name
Bug Fixing - PDPDEVTOOL-4363: Project is not opening in new window

In CreateProject.ts and ExtensionUtils.ts, please ignore whitespaces. There was an extra space on every line that I removed.


packages/node-cli/src/validation/InteractiveAnswersValidator.js
```diff
@@ -181,6 +181,14 @@ module.exports = {
 		return !fileSystemService.fileExists(path.join(parentFolderPath, filenameWithExtension))
 			? VALIDATION_RESULT_SUCCESS
 			: VALIDATION_RESULT_FAILURE(NodeTranslationService.getMessage(ANSWERS_VALIDATION_MESSAGES.FILE_ALREADY_EXISTS));
+	},
+
+	validateFolderDoesNotExists(path) {
```
davidecorreu, line 186: Just a grammar question, shouldn't it be `validateFolderDoesNotExist` and `validateSuiteScriptFileDoesNotExist` without the final `s`?
Anyway it's ok for me.

packages/vscode-extension/src/commands/CreateProject.ts
```diff
@@ -3,255 +3,267 @@
...
+            this.translationService.getMessage(CREATE_PROJECT.BUTTONS.NEW_WINDOW),
+        );
+
+        if (openProjectInNewWindow !== undefined) {
+            commands.executeCommand(
+                VSCODE_OPEN_FOLDER_COMMAND,
+                Uri.file(projectAbsolutePath),
+                {
+                    forcenewwindow: openProjectInNewWindow ? this.translationService.getMessage(CREATE_PROJECT.BUTTONS.NEW_WINDOW) : false,
+                }
```
davidecorreu, line 76: I think this should be:
`forcenewwindow: openProjectInNewWindow === this.translationService.getMessage(CREATE_PROJECT.BUTTONS.NEW_WINDOW)`

packages/node-cli/src/validation/InteractiveAnswersValidator.js
karl-anthony-ng: You are correct. I updated it.

antoniofruci (approved):

PR #418, opened 2021-08-18
Bug Fixes - PDPDEVTOOL-4372, PDPDEVTOOL-4373, PDPDEVTOOL-4374

karl-anthony-ng:
Fixed the following bugs:
- PDPDEVTOOL-4372: Shouldn't ask about importing referenced SuiteScripts when in a SuiteApp
- PDPDEVTOOL-4373: The object import shouldn't require an active XML from Objects dir
- PDPDEVTOOL-4374: Ask users choice to select sub directories under Objects folder

### DIFF
--- a/packages/vscode-extension/src/commands/ImportObjects.ts
+++ b/packages/vscode-extension/src/commands/ImportObjects.ts
@@ -109,12 +109,11 @@ export default class ImportObjects extends BaseAction {
 	private getDestinationFolder(pathDir: string) {
 		const isDirectory = fs.lstatSync(pathDir).isDirectory();
 		const directoryName = isDirectory ? pathDir : path.dirname(pathDir);
-		const destinationFolder = directoryName.split(this.getProjectFolderPath())[1].replace(/\\/gi, '/');
-		return destinationFolder;
+		return directoryName.split(this.getProjectFolderPath())[1].replace(/\\/gi, '/');
 	}
 
 	private async promptFilterAppId() {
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -28,10 +28,9 @@ export default class ImportObjects extends BaseAction {
 		}
 		const projectInfoService = new ProjectInfoService(this.getProjectFolderPath());
 
-		const relativeDestinationFolder = this.getDestinationFolder(this.activeFile);
-		if (!relativeDestinationFolder.startsWith(ApplicationConstants.FOLDERS.OBJECTS)) {
-			this.messageService.showErrorMessage(this.translationService.getMessage(IMPORT_OBJECTS.ERROR.INCORRECT_FOLDER));
-			return;
+		let relativeDestinationFolder = ApplicationConstants.FOLDERS.OBJECTS;
+		if (this.isSelectedFromContextMenu) {
+			relativeDestinationFolder = this.getDestinationFolder(this.activeFile);
```
davidecorreu, line 33: Shouldn't we ask about the destination folder to the user as we do in node-cli when the `/Objects` folder or the selected one from the context menu contains subfolders?

davidecorreu (approved):

packages/vscode-extension/src/commands/ImportObjects.ts
davidecorreu: Let´s share this point with the team, and maybe add it as an enhancement.

antoniofruci (approved):

PR #423, opened 2021-08-23
fix shown output for CreateFileAction.js

davidecorreu:
Removes the extra "" added to the output message.

### DIFF
--- a/packages/node-cli/src/utils/CommandUtils.js
+++ b/packages/node-cli/src/utils/CommandUtils.js
@@ -1,7 +1,7 @@
 /*
-** Copyright (c) 2021 Oracle and/or its affiliates.  All rights reserved.
-** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
-*/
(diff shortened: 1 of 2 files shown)


packages/node-cli/src/utils/CommandUtils.js
```diff
@@ -38,6 +38,13 @@ class CommandUtils {
 	quoteString(string) {
 		return `"${string}"`;
 	}
+
+	unquoteString(stringValue) {
+		if (stringValue.length > 1 && stringValue.startsWith('"') && stringValue.endsWith('"')) {
```
antoniofruci, line 43: Why >1 ?

davidecorreu: I was thinking on `"`.

antoniofruci: I don't know if that can happen and if it does the result would be weird because the function will return the double quotes.... Anyway it would be a very rare case so I wouldn't mind too much :)

antoniofruci (approved):

karl-anthony-ng (approved):

PR #424, opened 2021-08-23
PDPDEVTOOL-4387 showing error in Setup Account command

antoniofruci:
showing error in Setup Account command when jar is corrupted

### DIFF
--- a/packages/vscode-extension/src/commands/SetupAccount.ts
+++ b/packages/vscode-extension/src/commands/SetupAccount.ts
@@ -19,6 +19,7 @@ enum UiOption {
 	cancel_process,
 	dismiss,
 }
+
 interface NewAuthIdItem extends QuickPickItem {
 	option: UiOption.new_authid;
 }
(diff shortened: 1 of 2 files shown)


packages/vscode-extension/src/commands/SetupAccount.ts
```diff
@@ -47,28 +47,28 @@ export default class SetupAccount extends BaseAction {
...
-
-		if (actionResult.status === actionResultStatus.SUCCESS) {
-			const selected = await this.getAuthListOption(actionResult.data);
-			if (!selected) {
-				return;
-			} else if (selected.option === UiOption.new_authid) {
-				this.handleNewAuth(actionResult.data);
-			} else if (selected.option === UiOption.select_authid) {
-				this.handleSelectedAuth(selected.authId);
+			const actionResult: ActionResult<AuthListData> = await accountsPromise;
```
lfpvillegas, line 53: Identation seems strange. Please check.

lfpvillegas (changes requested):

packages/vscode-extension/src/commands/SetupAccount.ts
karl-anthony-ng, line 53: Can you check the indentation? It looks like there's one extra tab.

karl-anthony-ng (changes requested):

packages/vscode-extension/src/commands/SetupAccount.ts
```diff
@@ -47,28 +47,28 @@ export default class SetupAccount extends BaseAction {
...
+				if (!selected) {
+					return;
+				} else if (selected.option === UiOption.new_authid) {
+					await this.handleNewAuth(actionResult.data);
+				} else if (selected.option === UiOption.select_authid) {
+					this.handleSelectedAuth(selected.authId);
+				}
+			} else {
+				this.messageService.showCommandError(actionResult.errorMessages[0]);
+				this.vsConsoleLogger.error(actionResult.errorMessages[0]);
```
davidecorreu, line 65: Maybe we just want to `this.messageService.showErrorMessage(actionResult.errorMessages[0]);`  ?
Because in these 2 lines we are populating commandError message and then consoleLogger as well with the same info. 
Or do we want to intentionally leave the message in the SuiteCloud Output window. I'm fine with it, I just want to know if it's done on propose.

antoniofruci: You are right, now it´s consistent with the other commands when such error occurs. I removed the message from the function and swapped the order

karl-anthony-ng (approved):

lfpvillegas (approved):

davidecorreu (approved):

packages/vscode-extension/src/commands/SetupAccount.ts
davidecorreu: Great. Thanks

PR #438, opened 2021-09-03
Add LICENSE to vscode-extension package

karl-anthony-ng:


packages/vscode-extension/LICENSE
```diff
@@ -0,0 +1,34 @@
+Copyright (c) 2019, Oracle and/or its affiliates.  All rights reserved.
```
davidecorreu: I think we could use 2021 as year.

davidecorreu (approved):

lfpvillegas (approved):

PR #439, opened 2021-09-03
PDPDEVTOOL-4444 removing trailing spaces in ACP projects folder name

antoniofruci:
removing trailing spaces in ACP projects folder name creation

### DIFF
--- a/packages/vscode-extension/src/commands/CreateProject.ts
+++ b/packages/vscode-extension/src/commands/CreateProject.ts
@@ -101,7 +101,8 @@ export default class CreateProject extends BaseAction {
 		if (!projectName) {
 			return {};
 		}
-		commandArgs[COMMAND_ARGUMENTS.PROJECT_NAME] = projectName;
+		commandArgs[COMMAND_ARGUMENTS.PROJECT_NAME] = projectName.trim();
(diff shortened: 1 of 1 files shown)


karl-anthony-ng (approved):

packages/vscode-extension/src/commands/CreateProject.ts
```diff
@@ -124,9 +125,8 @@ export default class CreateProject extends BaseAction {
 			commandArgs[COMMAND_ARGUMENTS.PROJECT_VERSION] = projectVersion;
 
 			commandArgs[COMMAND_ARGUMENTS.PROJECT_FOLDER_NAME] = publisherId + '.' + projectId;
```
davidecorreu, line 127: Should we also trim publisherId and projectId ?

antoniofruci: It looks like a good idea

antoniofruci: But it's not needed, since into the validator we already check the fields have no spaces with the validator: validateFieldHasNoSpaces

davidecorreu (approved):

PR #516, opened 2022-01-26
Fix vscode engine and add types to exceptions.

lfpvillegas:


packages/vscode-extension/package.json
```diff
@@ -23,7 +23,7 @@
 		"**/*.vsix"
 	],
 	"engines": {
-		"vscode": "^1.57.0"
+		"vscode": "^1.63.1"
```
davidecorreu, line 26: If we change this line we should also edit the `/packages/vscode-extension/README.md` that contains a related reference.
`VS Code version 1.57 or higher`.

lfpvillegas: Done. Good catch

davidecorreu (approved):

karl-anthony-ng (approved):

PR #518, opened 2022-01-27
PDPDEVTOOL-4610 show license before installing cli

antoniofruci:


packages/node-cli/preinstall.js
```diff
@@ -0,0 +1,50 @@
...
+		input: process.stdin,
+		output: process.stdout,
+	});
+
+	rl.on('close', () => {
+		console.log(`${os.EOL}Installation aborted by user.`);
+		process.exit(REJECT_EXIT_CODE);
+	});
+
+	console.log(RED_COLOR);
```
lfpvillegas, line 24: I suppose we cannot use chalk because in here. Could it be possible to put RED_COLOR, message and COLOR_RESET in the same console.log?

Not a big deal if it's not possible.

davidecorreu (commented):
So, far looks good. I'll wait to approve it, to see how we can deal with CI environments.

packages/node-cli/preinstall.js
```diff
@@ -0,0 +1,104 @@
...
+(() => {
+	if (process.env.npm_config_supressSuiteCloudSDKLicensePrompt) {
+		return showLicenseWithoutPrompt();
+	}
+	return showLicenseAndPrompt();
+})();
+
+function showLicenseWithoutPrompt() {
+
+		try {
```
lfpvillegas, line 36: I think there are more spaces here. Please format.

davidecorreu (approved):

PR #527, opened 2022-02-03
feature/PDPDEVTOOL-4131

lfpvillegas:


packages/node-cli/src/commands/object/update/UpdateAction.js
```diff
@@ -47,26 +53,86 @@ module.exports = class UpdateAction extends BaseAction {
...
+
+	async _executeCommandUpdateCustomRecordWithInstances(params, scriptId) {
+		params[COMMAND_OPTIONS.SCRIPT_ID] = scriptId;
+		const executionContextForUpdate = SdkExecutionContext.Builder.forCommand(COMMAND_UPDATE_CUSTOM_RECORD_WITH_INSTANCES)
+			.integration()
+			.addParam(COMMAND_OPTIONS.AUTH_ID, params[COMMAND_OPTIONS.AUTH_ID])
+			.addParam(COMMAND_OPTIONS.PROJECT, params[COMMAND_OPTIONS.PROJECT])
+			.addParam(COMMAND_OPTIONS.SCRIPT_ID, scriptId)
+			.build();
+		const operationResult = await executeWithSpinner({
```
karl-anthony-ng, line 132: You can return here instead of assigning it first to an operationResult variable.

karl-anthony-ng (changes requested):

packages/node-cli/src/commands/object/update/UpdateInputHandler.js
```diff
@@ -41,7 +48,48 @@ module.exports = class UpdateInputHandler extends BaseInputHandler {
 		this._fileSystemService = new FileSystemService();
 	}
 
-	async getParameters(params) {
+	async getParameters(_params) {
```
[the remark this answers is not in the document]
lfpvillegas: This was a VsCode suggestion, because params is not used.
Normally in Javascript, you put the varaibles you don't use (but you need to put because of inheritance) with an underscore.
Anyways, I could put it params normally

packages/node-cli/src/metadata/CommandGenerators.json
```diff
@@ -49,6 +49,11 @@
 		"supportsInteractiveMode": true,
 		"generator": "./commands/object/update/UpdateCommand"
 	},
+	{
+		"commandName": "object:updatecustomrecordwithinstances",
+		"supportsInteractiveMode": false,
+		"generator": "./commands/object/update/UpdateCommand"
+	},
```
davidecorreu, line 56: If we haven't created any new command, should this be here?

packages/node-cli/src/metadata/SdkCommandsMetadata.json
```diff
@@ -589,7 +589,7 @@
 			"scriptid": {
 				"name": "scriptid",
 				"option": "scriptid",
-				"description": "Specify the script ID of the objects you want to overwrite.",
+				"description": "Specifies the script ID of the objects you want to overwrite.",
```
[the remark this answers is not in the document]
lfpvillegas: Is in SdkCommandsMetadataPatch.json in object:update-> options -> includeinstances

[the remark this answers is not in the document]
davidecorreu: Sorry I didn't have your branch updated.

packages/node-cli/src/commands/object/update/UpdateInputHandler.js
```diff
@@ -42,6 +49,47 @@ module.exports = class UpdateInputHandler extends BaseInputHandler {
 	}
 
 	async getParameters(params) {
+		const foundXMLFiles = this.searchFilesFromObjectsFolder();
+		let filteredObjects = await this.getObjectsToSelect(foundXMLFiles);
+		const selectedScriptIds = await this.selectObjectsScriptIds(foundXMLFiles, filteredObjects);
```
davidecorreu, line 54: It think we can name these like:
filteredObjects -> filteredObjectsList
selectObjectsScriptsIds(foundXMLFiles, filteredObjects) method -> getSelectedScriptsIds(filteredObjects)
Actually the method should only use the filteredObjects/filteredObjectsList.

packages/node-cli/src/commands/object/update/UpdateInputHandler.js
```diff
@@ -54,68 +102,69 @@ module.exports = class UpdateInputHandler extends BaseInputHandler {
...
-			{
-				when: foundXMLFiles.length > 1,
-				type: CommandUtils.INQUIRER_TYPES.CHECKBOX,
-				name: ANSWERS_NAMES.SCRIPT_ID_LIST,
-				message: NodeTranslationService.getMessage(QUESTIONS.SCRIPT_ID),
-				default: 1,
-				choices: filteredObjects,
-				validate: (fieldValue) => showValidationResults(fieldValue, validateArrayIsNotEmpty),
-			},
+		return filteredObjects;
```
davidecorreu, line 121: I think we shouldn't add the new Separator() to the filteredObjects. And later in the next method selectObjectsScriptIds we could check if the length of the filteredObjects is bigger 1 to actually ask about which objects the user want to update.
Actually after some tryout I have found that with the current implementation we are not skipping the question even when there is only one object after the filtering, which may be good so the user will see the object that is being updated, but if we go in this direction I think we should remove the when clause on line 126. Actually I think that if there is only one object in the project the code will fail because the `return answers[ANSWERS_NAMES.SCRIPT_ID_LIST];` will be undefined.

packages/node-cli/src/commands/object/update/UpdateInputHandler.js
```diff
@@ -42,6 +49,47 @@ module.exports = class UpdateInputHandler extends BaseInputHandler {
...
 	async getParameters(params) {
+		const foundXMLFiles = this.searchFilesFromObjectsFolder();
+		let filteredObjects = await this.getObjectsToSelect(foundXMLFiles);
+		const selectedScriptIds = await this.selectObjectsScriptIds(foundXMLFiles, filteredObjects);
+		const customRecordsAndSegments = selectedScriptIds.filter(
+			(scriptid) => scriptid.startsWith(SCRIPT_ID_PREFIXES.CUSTOM_RECORD) || scriptid.startsWith(SCRIPT_ID_PREFIXES.CUSTOM_SEGMENT)
+		);
+
+		const includeCustomInstancesQuestions = {
+			when: customRecordsAndSegments.length >= 1, // && customRecordsAndSegments.length == selectedScriptIds.length,
```
davidecorreu, line 60: We can remove the comment.

packages/node-cli/src/commands/object/update/UpdateInputHandler.js
```diff
@@ -54,68 +102,69 @@ module.exports = class UpdateInputHandler extends BaseInputHandler {
...
+			default: 1,
+			choices: filteredObjects,
+			validate: (fieldValue) => showValidationResults(fieldValue, validateArrayIsNotEmpty),
+		};
+		const answers = await prompt([selectObjectsToUpdateQuestion]);
+
+		return answers[ANSWERS_NAMES.SCRIPT_ID_LIST];
+	}
+
+	async filterObjectsByScriptId(filteredObjects, foundXMLFiles) {
```
davidecorreu, line 139: There is no need to pass here `filteredObjects` as argument. Please remove it from the method signature.

packages/node-cli/src/commands/object/update/UpdateAction.js
```diff
@@ -47,26 +53,85 @@ module.exports = class UpdateAction extends BaseAction {
...
+
+				delete params[COMMAND_OPTIONS.INCLUDE_INSTANCES];
+			}
+
+			if (params[COMMAND_OPTIONS.SCRIPT_ID] === '') {
+				// If there are no more objects to update (all were custom records) we already return with only previous messages
+				return ActionResult.Builder.withData(totalData).build();
+			}
+
+			return await this._updateObjects(params, totalData);
```
davidecorreu, line 76: Not sure if we should be returning the result of the "normal" `this._updateObjects` and merge the results here to have more clarity on what we are doing. In this way we would not need to pass the "totalData" to `this._updateObjects`

packages/node-cli/src/commands/object/update/UpdateAction.js
```diff
@@ -47,26 +53,85 @@ module.exports = class UpdateAction extends BaseAction {
...
+	}
+
+	_isCustomRecordOrSegment(scriptid) {
+		return scriptid.startsWith(SCRIPT_ID_PREFIXES.CUSTOM_RECORD) || scriptid.startsWith(SCRIPT_ID_PREFIXES.CUSTOM_SEGMENT);
+	}
+
+	async _updateCustomRecordWithInstances(params, customRecordScriptIds) {
+		this._log.result(NodeTranslationService.getMessage(OUTPUT.UPDATED_CUSTOM_RECORDS));
+		const copiedParams = { ...params };
+		const dataResults = [];
```
davidecorreu, line 108: What about `updateWithInstancesResults` ?

packages/node-cli/src/commands/object/update/UpdateInputHandler.js
```diff
@@ -54,68 +102,65 @@ module.exports = class UpdateInputHandler extends BaseInputHandler {
...
 				throw NodeTranslationService.getMessage(MESSAGES.NO_OBJECTS_WITH_SCRIPT_ID_FILTER);
 			}
+			return filteredObjects;
 		} else {
-			filteredObjects = foundXMLFiles;
+			return foundXMLFiles;
 		}
+	}
+
+	async _filterObjectsByScriptId(filteredObjects, foundXMLFiles) {
```
davidecorreu, line 120: We also don't need here the `fileteredObjects` passed as a parameter, please remove it from the method signature. Later we directly can return the ternary operator result.

packages/node-cli/src/commands/object/update/UpdateOutputHandler.js
```diff
@@ -20,8 +21,18 @@ module.exports = class UpdateOutputHandler extends BaseOutputHandler {
 	}
 
 	parse(actionResult) {
-		const updatedObjects = actionResult.data.filter((element) => element.type === UPDATED_OBJECT_TYPE.SUCCESS);
-		const noUpdatedObjects = actionResult.data.filter((element) => element.type !== UPDATED_OBJECT_TYPE.SUCCESS);
+		this._parseUpdateObjects(actionResult);
+		this._parseUpdateCustomRecordsWithInstancesNonInteractiveExecution(actionResult);
+		return actionResult;
```
davidecorreu, line 26: I think this is the only command that outputs messages during the action itself which is a "kind of hack" for the CustomRecordWithInstances and it will be not obvious to someone looking at this code.
My suggestion is to at least add a comment here before line 24, something like: // updated CustomRecordsWithInstances had been logged during the action execution.

Then I also think that we could rename `_parseUpdateCustomRecordsWithInstancesNonInteractiveExecution` to `_showNonInteractiveExecForCustomRecordWithInstances` or `_logNonInteractiveExecForCustomRecordWithInstances` because we are not parsing/outputting any part of the actionResult.

The rest of the file looks good to me.

packages/node-cli/src/commands/object/update/UpdateAction.js
```diff
@@ -47,26 +53,90 @@ module.exports = class UpdateAction extends BaseAction {
...
+				return ActionResult.Builder.withData(updateCustomRecordsWithInstancesResult).build();
+			}
 
-			return operationResult.status === SdkOperationResultUtils.STATUS.SUCCESS
-				? ActionResult.Builder.withData(operationResult.data)
-					.withResultMessage(operationResult.resultMessage)
-					.withCommandParameters(sdkParams)
-					.build()
-				: ActionResult.Builder.withErrors(operationResult.errorMessages).withCommandParameters(sdkParams).build();
+			const sdkParams = CommandUtils.extractCommandOptions(params, this._commandMetadata);
```
davidecorreu, line 75: This line could be inside `this._updateObjects` and just pass the `params` as parameters.

packages/node-cli/src/commands/object/update/UpdateAction.js
```diff
@@ -47,26 +53,90 @@ module.exports = class UpdateAction extends BaseAction {
...
+			const updateObjectsResult = await this._updateObjects(sdkParams);
+			updateCustomRecordsWithInstancesResult = updateCustomRecordsWithInstancesResult.concat(updateObjectsResult.data);
```
davidecorreu, line 77: I personally don't like re-assigning the overall results to the `updateCustomRecordsWithInstancesResult` variable, we could create a new variable named `allResults` or `totalResults` or `aggregatedResults` and assign it to it.
`const aggregatedResults= updateCustomRecordsWithInstancesResult.concat(updateObjectsResult.data)`.

It will also make more clear the Builder call in line 80 that will make use of this variable.

I know it looks I'm going back to the previous implementation, but it's because the context has changed, sorry 😅 .

packages/node-cli/src/commands/object/update/UpdateAction.js
lfpvillegas: Hey David, I have put this out, because sdkParams in needed below. I don't want to call two times to CommandUtils.extractCommandOptions

This is because you asked to get out the following code from  _updateObjects.

So either we go to the previous implementation, or we stay with sdkParams here

davidecorreu: Oh, I didn't realize about it. Then I think we can leave them here.

karl-anthony-ng (approved):

PR #528, opened 2022-02-03
Updata package-lock.json

lfpvillegas:


packages/vscode-extension/package-lock.json
```diff
@@ -72,19 +72,6 @@
 				"fastq": "^1.6.0"
 			}
 		},
-		"@oracle/suitecloud-cli": {
```
davidecorreu, line 75: I think we have completely removed this dependency. I don't think this should happen.

davidecorreu (approved):

antoniofruci (approved):

PR #534, opened 2022-02-11
feature/pdpdevtool-4539: validate for vscode

davidecorreu:
validiate server side for vscode
reuse common logic for deploy and validate on DeployValidateCommon.ts


lfpvillegas (approved):

packages/vscode-extension/src/commands/DeployValidateCommon.ts
```diff
@@ -0,0 +1,126 @@
...
+const ACCOUNT_SPECIFIC_VALUES = {
+	ERROR: 'ERROR',
+	WARNING: 'WARNING',
+};
+
+interface commandOption extends QuickPickItem {
+	value: string;
+}
+
+export default abstract class DeployValidateCommon extends BaseAction {
```
karl-anthony-ng, line 29: What do you think of renaming this to something like BaseDeployValidateAction or DeployValidateCommonAction?

karl-anthony-ng (changes requested):

packages/vscode-extension/src/commands/DeployValidateCommon.ts
karl-anthony-ng: As discussed, I'm good with DeployValidateCommon. No changes needed.

karl-anthony-ng (approved):

PR #545, opened 2022-02-24
feature/pdpdevtool-4540: manage auth vscode

davidecorreu:


packages/vscode-extension/src/loggers/VSConsoleLogger.ts
```diff
@@ -14,6 +14,7 @@ const INVALID_JAR_FILE_MESSAGE = 'Invalid or corrupt jarfile';
 export default class VSConsoleLogger extends ConsoleLogger {
 	private _executionPath?: string;
 	private _addExecutionDetailsToLog: boolean = false;
+	private _hiddeProjectFolerName: boolean = false;
```
karl-anthony-ng, line 17: Typo in "_hiddeProjectFolerName". Should this be "_hideProjectFolderName"?

And since this is TypeScript, I don't think we need to prefix private variables with underscores ("_") anymore? What do you think?

karl-anthony-ng (changes requested):

packages/vscode-extension/src/commands/ImportObjects.ts
```diff
@@ -131,14 +131,13 @@ export default class ImportObjects extends BaseAction {
 
 		if (objectsFolderAbsolutePaths.length > 0) {
 			const srcFolderAbsolutePath = path.join(rootWorkspaceFolder, SRC_FOLDER_NAME);
-			const transformAbsolutePathsToRelativePathsFunc = (absolutePath: string) => (
-				absolutePath.replace(srcFolderAbsolutePath, '').replace(/\\/g, '/')
-			);
-			const objectsFolderRelativePaths = objectsFolderAbsolutePaths.map(transformAbsolutePathsToRelativePathsFunc)
+			const transformAbsolutePathsToRelativePathsFunc = (absolutePath: string) =>
```
antoniofruci, line 134: I am not sure, but maybe this function could be replaced with path.relative native function: https://nodejs.org/docs/latest-v14.x/api/path.html#path_path_relative_from_to

antoniofruci (approved):

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
```diff
@@ -65,7 +65,7 @@ async function install() {
 		await downloadFilePromise;
 		messageService.showInformationMessage(translationService.getMessage(EXTENSION_INSTALLATION.SUCCESS.SDK_DOWNLOADED));
 	} catch (error) {
-		messageService.showErrorMessage(translationService.getMessage(EXTENSION_INSTALLATION.ERROR.GENERAL_ERROR, fullUrl, error));
+		messageService.showErrorMessage(translationService.getMessage(EXTENSION_INSTALLATION.ERROR.GENERAL_ERROR, fullUrl, `${error}`));
```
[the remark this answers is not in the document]
davidecorreu: As I have added:
 ```
 getMessage(key: string, ...params: string[]): string {
		return super.getMessage(key, ...params);
	}
```
to VSTranslationService class to have "code completion" for the `getMessage` method, now typescript is expecting a strings as `params` after `key`. So the typecheck is preventing to directly pass the `error` and we need to pass a string, and that's why I'm using `'${error}'` but we could also use `error as string`.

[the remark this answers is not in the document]
karl-anthony-ng: Understood. I'm ok with what you did. No need to change it to error as string.

packages/vscode-extension/src/commands/ImportObjects.ts
davidecorreu: I needed to change line 95 for a change related again with `TranslationService.getMessage` returning a string type . This line has changed due to the formatting but I haven't actually worked in any logical change in this file.
I would prefer to not risk any possible logical change without proper QA testing. I say so, because we are removing some relative path, but also replacing the `\` character in the function.

karl-anthony-ng (approved):

PR #546, opened 2022-02-24
Removing csegs from updating with custom isntances

lfpvillegas:


antoniofruci (approved):

packages/node-cli/src/commands/object/update/UpdateAction.js
```diff
@@ -34,7 +34,6 @@ const COMMAND_OPTIONS = {
 const COMMAND_UPDATE_CUSTOM_RECORD_WITH_INSTANCES = 'updatecustomrecordwithinstances';
 const SCRIPT_ID_PREFIXES = {
 	CUSTOM_RECORD: 'customrecord',
```
antoniofruci, line 36: This value could be extracted as CUSTOM_RECORD_PREFIX directly

karl-anthony-ng (approved):

PR #553, opened 2022-03-11
Feature/pdpdevtool 4541 vscode update instances

antoniofruci:


packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -41,9 +43,31 @@ export default class UpdateObject extends BaseAction {
...
+		const yes = this.translationService.getMessage(ANSWERS.YES);
+		const no = this.translationService.getMessage(ANSWERS.NO);
+		const includeInstancesAnswer = await window.showQuickPick([yes, no], {
+			placeHolder: this.translationService.getMessage(UPDATE_OBJECT.QUESTIONS.INCLUDE_INSTANCES),
+			canPickMany: false,
+		});
+
+
+		let includeInstancesFlag = '';
+		if (includeInstancesAnswer === this.translationService.getMessage(ANSWERS.YES)) {
```
lfpvillegas, line 55: You could use previously defined constant.

packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -41,9 +43,31 @@ export default class UpdateObject extends BaseAction {
...
+			const overwriteInstances = await window.showQuickPick([continueMessage, cancelMessage], {
+				placeHolder: this.translationService.getMessage(UPDATE_OBJECT.OVERWRITE_INSTANCES),
+				canPickMany: false,
+			});
+	
+			if (!overwriteInstances || overwriteInstances === this.translationService.getMessage(ANSWERS.CANCEL)) {
```
lfpvillegas, line 61: you could use previously defined constant cancel (also in line 41)

lfpvillegas (changes requested):

packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -31,19 +33,42 @@ export default class UpdateObject extends BaseAction {
...
-			this.messageService.showInformationMessage(this.translationService.getMessage(UPDATE_OBJECT.PROCESS_CANCELED));
+		if (!overwriteObject || overwriteObject === cancelMessage) {
+			this.messageService.showInformationMessage(processCanceledMessage);
 			return;
 		}
 
+		const includeInstancesAnswer = await window.showQuickPick([yes, no], {
+			placeHolder: this.translationService.getMessage(UPDATE_OBJECT.QUESTIONS.INCLUDE_INSTANCES),
+			canPickMany: false,
+		});
```
davidecorreu, line 53: This should only be asked if the selected object is a custom record. There is some logic missing here.

packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -31,19 +33,42 @@ export default class UpdateObject extends BaseAction {
...
+
+		let includeInstancesFlag = '';
+		if (includeInstancesAnswer === yes) {
+			const overwriteInstances = await window.showQuickPick([continueMessage, cancelMessage], {
```
davidecorreu, line 57: I'm not in favor of asking twice about overwriting, once for regular objects, and again for the instances of custom records. I think is a bad user experience for the developer. I would prefer to ask only once, either overwrite objects or overwrite objects and its instances.
I'm actually not sure that we could update more than one object at a time, but I would like to be only be asked about overwriting only once in any case.

packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -103,4 +99,8 @@ export default class UpdateObject extends BaseAction {
 			return this.unsuccessfulValidation(e.getErrorMessage());
 		}
 	}
+
+	private static isCustomRecord(scriptid: string): boolean {
```
lfpvillegas, line 103: we don't really need that this method is static, it is not called for anywhere else.

packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -31,19 +34,36 @@ export default class UpdateObject extends BaseAction {
...
+
+		if (UpdateObject.isCustomRecord(scriptId)) {
+			const includeInstancesAnswer = await window.showQuickPick([yes, no], {
+				placeHolder: this.translationService.getMessage(UPDATE_OBJECT.QUESTIONS.INCLUDE_INSTANCES),
+				canPickMany: false,
+			});
+			if (includeInstancesAnswer === yes) {
+				suiteCloudCommand.includeinstances = INCLUDE_INSTANCES;
+				overwriteMessage = this.translationService.getMessage(UPDATE_OBJECT.OVERWRITE_INSTANCES, scriptId);
+			}
```
davidecorreu, line 51: I think we are missing the case were the user doesn't respond to the question or just hit ESC , in that case I think we should stop the command flow. Something like:
```
if (!includeInstancesAnswer) {
  return;
}
```

packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -31,19 +34,36 @@ export default class UpdateObject extends BaseAction {
 
 		const continueMessage = this.translationService.getMessage(ANSWERS.CONTINUE);
 		const cancelMessage = this.translationService.getMessage(ANSWERS.CANCEL);
-		const override = await window.showQuickPick([continueMessage, cancelMessage], {
-			placeHolder: this.translationService.getMessage(UPDATE_OBJECT.OVERRIDE, scriptId),
+		const yes = this.translationService.getMessage(ANSWERS.YES);
+		const no = this.translationService.getMessage(ANSWERS.NO);
+		const processCanceledMessage = this.translationService.getMessage(UPDATE_OBJECT.PROCESS_CANCELED);
+		const suiteCloudCommand = { scriptid: [scriptId], includeinstances: '' };
```
davidecorreu, line 40: I would prefer to name this one `suiteCloudCommandOptions` / `commandOptions` or similar one with "options" in the name.

packages/vscode-extension/src/commands/UpdateObject.ts
```diff
@@ -31,19 +34,39 @@ export default class UpdateObject extends BaseAction {
...
+				placeHolder: this.translationService.getMessage(UPDATE_OBJECT.QUESTIONS.INCLUDE_INSTANCES),
+				canPickMany: false,
+			});
+			if (includeInstancesAnswer === yes) {
+				suiteCloudCommandOptions.includeinstances = INCLUDE_INSTANCES;
+				overwriteMessage = this.translationService.getMessage(UPDATE_OBJECT.OVERWRITE_INSTANCES, scriptId);
+			} else if(!includeInstancesAnswer) {
+				this.messageService.showInformationMessage(processCanceledMessage);
+				return;
+			}
```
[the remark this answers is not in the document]
antoniofruci: True, I'll move it

lfpvillegas (approved):

PR #564, opened 2022-04-08
Bugfix/pdpdevtool 4753 sdk download error

antoniofruci:
Copy license in tmp folder before opening it.

### DIFF
--- a/packages/node-cli/src/core/sdksetup/SdkLicense.js
+++ b/packages/node-cli/src/core/sdksetup/SdkLicense.js
@@ -28,7 +28,10 @@ class SdkLicense {
 			return;
 		}
 		const tmpFilePath = os.tmpdir() + path.sep + LICENSE_FILENAME;
-		fs.cpSync(LICENSE_PATH, tmpFilePath);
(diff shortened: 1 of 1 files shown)


lfpvillegas (approved):

packages/node-cli/src/core/sdksetup/SdkLicense.js
```diff
@@ -26,9 +27,14 @@ class SdkLicense {
 		if (process.env.npm_config_acceptsuitecloudsdklicense || process.env.npm_config_acceptSuiteCloudSDKLicense) {
 			return;
 		}
+		const tmpFilePath = os.tmpdir() + path.sep + LICENSE_FILENAME;
+		const fileExists = fs.lstatSync(tmpFilePath);
```
davidecorreu, line 31: From nodejs documentation seems that fs.lstatSync() throws if the entry is not found by default. Don't we need to change that behavior here or surround this call in try/catch block?

antoniofruci: You're right. I'll change it

davidecorreu (approved):

PR #586, opened 2022-06-28
Feature/pdpdevtool 4784 vscode file compare

karl-anthony-ng:


tourajvaziri (approved):

packages/vscode-extension/src/commandsMap.ts
```diff
@@ -35,6 +36,11 @@ export const commandsInfoMap: CommandsInfoMapType = {
 		vscodeCommandName: 'Add Dependency References to the Manifest',
 		vscodeShortName: 'Add Dependencies',
 	},
+	comparefile: {
+		cliCommandName: 'file:import',
```
tourajvaziri, line 40: why the command is file:import?

packages/vscode-extension/src/commands/CompareFile.ts
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		return this.successfulValidation();
+	}
+
+	protected async execute() {
+		const activeFilePath = this.activeFile!;
+		const tempFolderPath = fs.mkdtempSync(path.join(os.tmpdir(), CompareFile.TEMP_FOLDER_PREFIX));
+
+		this.copyManifestFileToTempFolder(tempFolderPath);
```
tourajvaziri, line 40: why do we need to copy this?

tourajvaziri, line 40: why do we need to copy this?

karl-anthony-ng: There are validations in the manifest file that is being performed before the action is done. So manifest file is also copied to the temporary directory. The temporary directory becomes a "clone" of the project. That's where the import files is actually being performed.

karl-anthony-ng: The temporary directory becomes a "clone" of the project. That's where the import files is actually being performed. So that means project.json is also needed to do the actual import.

packages/vscode-extension/src/commandsMap.ts
karl-anthony-ng: There is no cli command for compare. We are actually calling file:import to import the file and then call vscode compare to compare the imported file to the file in the project.

packages/vscode-extension/src/commands/CompareFile.ts
```diff
@@ -0,0 +1,96 @@
...
+import * as os from 'os';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { UNRESTRICTED_FOLDERS } from '../ApplicationConstants';
+import { COMPARE_FILE, IMPORT_FILES } from '../service/TranslationKeys';
+import { actionResultStatus, ApplicationConstants } from '../util/ExtensionUtil';
+import FileImportCommon from './FileImportCommon';
+
+export default class CompareFile extends FileImportCommon {
+	private static readonly COMMAND_NAME = 'comparefile';
```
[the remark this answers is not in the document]
karl-anthony-ng: Removed spaces, changed to tabs.

packages/vscode-extension/src/commands/CompareFile.ts
```diff
@@ -0,0 +1,96 @@
...
+		if (!this.activeFileIsUnderUnrestrictedFolder()) {
+			return this.unsuccessfulValidation(this.translationService.getMessage(IMPORT_FILES.ERROR.NOT_ALLOWED_FOLDER));
+		}
+
+		return this.successfulValidation();
+	}
+
+	protected async execute() {
+		const activeFilePath = this.activeFile!;
+		const tempFolderPath = fs.mkdtempSync(path.join(os.tmpdir(), CompareFile.TEMP_FOLDER_PREFIX));
```
lfpvillegas, line 38: I'm wondering if the temp files are cleaning ok... in [this page](https://stackoverflow.com/questions/43607942/cleaning-temporary-files-after-file-upload-with-restify) it implies that in Unix it could not be cleaned until system is reboot.
I would say to put a last step (if possible) for cleaning these folder with fs.unlink()

Also, while searching about this, I found this[ other page](https://advancedweb.hu/secure-tempfiles-in-nodejs-without-dependencies/) explain also some possible problems with macOS and os.tmpdir() method but it may not apply to us, just sending it to you in case you find it useful.

packages/vscode-extension/src/commandsMap.ts
tourajvaziri: ok seems a bit strange but sure :)

packages/vscode-extension/src/commands/CompareFile.ts
tourajvaziri: same as above.

lfpvillegas: Hey Touraj, the problem is that the cli (and sdf) needs to import the file to specific location, if not there is a validation error. 
To do so, we would have to change how file:import cli command is working, which is not part of the story (which only applies to vscode), Moreover, I don't know if we want this to work this way and give all users this option (even if it's hidden)

Anyways, I also see some possible problems with current solution (what to do if the command stops in the middle of the execution and the file is not copied back to the original place) So this it's a possible solution that we may want to discuss.

packages/vscode-extension/src/commands/CompareFile.ts
karl-anthony-ng: Unfortunately, we can't do cleanup since the diff window still needs to access the file. If we do cleanup, the diff window will not be able to show the difference.

packages/vscode-extension/src/commands/CompareFile.ts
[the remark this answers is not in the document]
karl-anthony-ng: The original files are not touched so there is no need to copy from the temporary folder to the actual project.

packages/vscode-extension/src/commands/CompareFile.ts
lfpvillegas: Ok, and what about detecting when the diff window is closed and then removing the file? Maybe it is not possible, so in that case forget my comment.

packages/vscode-extension/src/commands/CompareFile.ts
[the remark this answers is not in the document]
lfpvillegas: Ah ok, I see it now. Then it's fine.

packages/vscode-extension/src/commands/CompareFile.ts
```diff
@@ -0,0 +1,97 @@
...
+
+	private getFileCabinetFolderPath(): string {
+		return path.join(this.getProjectFolderPath(), ApplicationConstants.FOLDERS.FILE_CABINET);
+	}
+
+	private copyActiveFileToTempFolder(tempFolderPath: string, activeFilePath: string, activeFileRelativePath: string): string {
+		const importFileParentFolderPath = path.join(tempFolderPath, ApplicationConstants.FOLDERS.FILE_CABINET, path.dirname(activeFileRelativePath));
+		fs.mkdirSync(importFileParentFolderPath, { recursive: true });
+		const importFilePath = path.join(importFileParentFolderPath, path.basename(activeFilePath));
+		fs.copyFileSync(activeFilePath, importFilePath);
```
lfpvillegas, line 83: Final question. Again, I may have not understood the full behavior.

Do we need to copy the original File to the temporary directory? I think it's no needed since the file:import command it's going to import it to the temp directory anyways.

packages/vscode-extension/src/commands/CompareFile.ts
tourajvaziri: I think we could leave it there. no?

antoniofruci: I don't think it's a big deal if we leave some small file in the temp folder :)

packages/vscode-extension/src/commands/CompareFile.ts
karl-anthony-ng: Good catch. No we don't. I pushed changes to remove copying of the original file.

packages/vscode-extension/src/commands/CompareFile.ts
karl-anthony-ng: There's no mechanism to capture the closing of the window.

antoniofruci (approved):

tourajvaziri (approved):

Montsant19 (approved):

PR #588, opened 2022-07-05
PDPDEVTOOL-4784: VSCODE - Ability to compare file with account version

karl-anthony-ng:
Bug fixes for the following...
- PDPDEVTOOL-4849: Compare File triggered from a wrong place shows misleading error
- PDPDEVTOOL-4850: The logs printed in the console after comparing are incorrect
- PDPDEVTOOL-4851: Compare File triggered from a suiteapp shows misleading error
- PDPDEVTOOL-4852: compare a file that does not exist handled incorrectly


antoniofruci (approved):

tourajvaziri (approved):

packages/node-cli/src/commands/file/import/ImportFilesAction.js
```diff
@@ -31,6 +32,7 @@ module.exports = class ImportFilesAction extends BaseAction {
 		super(options);
 
 		this._projectInfoService = new ProjectInfoService(this._projectFolder);
+		this._calledFromCompareFiles = true;
```
tourajvaziri, line 35: why is the default set to true. Should not be false?

karl-anthony-ng: Should be false. Updated.

Montsant19 (approved):

PR #596, opened 2022-08-04
PDPDEVTOOL-4862: Bug fixes for vscode-extension Compare Files

karl-anthony-ng:


packages/vscode-extension/messages.json
```diff
@@ -18,8 +18,8 @@
 	"COMMAND_TRIGGERED": "'{0}' has been triggered.",
 	"COMMAND_WARNING": "'{0}' has finished.",
 
-	"COMPARE_FILE_COMPARE_FILE_WITH_ACCOUNT_FILE": "Compare with Account File",
 	"COMPARE_FILE_COMPARING_FILE": "Comparing file...",
+	"COMPARE_FILE_EDITOR_LABEL": "Account Copy ⟷ Local Copy",
```
tourajvaziri, line 22: I see it here. Yes, i think we can change it to an arrow only on the Local Copy side.

antoniofruci: It looks good to me as it is

antoniofruci (approved):

packages/vscode-extension/messages.json
antoniofruci: Maybe we could have the file name into the tab title instead of this string? Is there any reason for not doing it?

tourajvaziri: It was suggested to go with this. What do you thin Karl?
filename.js (Account Copy  ⟷ Local Copy)

karl-anthony-ng: The final decision is to go with this format...
`filename.js (Account <-> Local)`

antoniofruci (approved):

packages/vscode-extension/messages.json
```diff
@@ -18,8 +18,8 @@
 	"COMMAND_TRIGGERED": "'{0}' has been triggered.",
 	"COMMAND_WARNING": "'{0}' has finished.",
 
-	"COMPARE_FILE_COMPARE_FILE_WITH_ACCOUNT_FILE": "Compare with Account File",
 	"COMPARE_FILE_COMPARING_FILE": "Comparing file...",
+	"COMPARE_FILE_EDITOR_LABEL": "{0} (Account ⟷  Local)",
```
tourajvaziri, line 22: Seems to be an extra space between  ⟷  Local

karl-anthony-ng: Intentional because without that extra space, it looks like the right there is no space between the right arrow and Local Copy.

tourajvaziri: oh great. then all good!:)

PR #601, opened 2022-08-19
Feature/pdpdevtool 4908 prepare for 2022.2.0 release

karl-anthony-ng:


packages/vscode-extension/package.json
```diff
@@ -12,7 +12,7 @@
 	"publisher": "Oracle",
 	"license": "UPL-1.0",
 	"private": true,
-	"version": "1.2.0",
+	"version": "1.3.0",
```
davidecorreu, line 15: On line 198 we have `"@oracle/suitecloud-cli": "^1.5.0"` , I think we should refer to the new version 1.6.0.

And there is also a java version check performed in vscode that I'm not sure if needs to be reviewed.

karl-anthony-ng: Updated to reference 1.6.0.

Regarding the Java version check. Aren't we just riding on the node-cli checking?

davidecorreu (approved):

antoniofruci (approved):

PR #604, opened 2022-09-01
Fix vscode-extension build problems by updating vscode engine version

karl-anthony-ng:


packages/vscode-extension/package.json
```diff
@@ -201,7 +201,7 @@
 		"@types/glob": "7.2.0",
 		"@types/mocha": "9.1.1",
 		"@types/node": "17.0.10",
-		"@types/vscode": "1.70.0",
+		"@types/vscode": "^1.70.0",
```
davidecorreu, line 204: We were not using the `^` in this line of the previous package.json

karl-anthony-ng: The reason I did this was so that it'll match with the vscode engine version. What do you think?

antoniofruci: I think it's indeed a good idea.

antoniofruci (approved):

packages/vscode-extension/package.json
davidecorreu: As I see it, the `engines.vscode="^1.70.0"` on top sets the minimum vscode application version that should use our plugin, and in consequence we shouldn't be using a higher version of `@types/vscode` to ensure compatibility with the 1.70.0 vscode minimal allowed version.

karl-anthony-ng: Yes that's correct. My assumption here is that there is always an equivalent @vscode/types to engines.vscode. And that they are always equivalent when using "^".

But maybe what we should do here is that for engines.vscode, it should be 1.70.0 instead of ^1.70.0 and then @vscode/types should be 1.70.0.

davidecorreu (approved):
Both solutions seems to be good. Approved.

packages/vscode-extension/package.json
karl-anthony-ng: We decided to remove the "^" from @types/vscode.

PR #606, opened 2022-09-06
PDPDEVTOOL-4863: Allow File Compare with SuiteApps

karl-anthony-ng:


packages/vscode-extension/src/commands/CompareFile.ts
```diff
@@ -96,9 +105,22 @@ export default class CompareFile extends FileImportCommon {
...
+		const suiteAppFileCabinetPath = path.join(
+			this.getFileCabinetFolderPath(),
+			ApplicationConstants.FOLDERS.SUITEAPPS,
+			projectInfoService.getApplicationId()
+		);
+		const suiteAppFileCabinetRelativePath = suiteAppFileCabinetPath
+			.replace(this.getFileCabinetFolderPath(), '')
+			.replace(/\\/g, '/');
+		return activeFileRelativePath!.startsWith(suiteAppFileCabinetRelativePath);
+	}
```
davidecorreu, line 119: I think there is no need to create 2 relative paths, maybe we could directly check `this.activeFile!.startsWith(suiteAppFileCabinetPath)` or have you already tried and there is some inconvenient?

I would also try to name the method something like `activeFileIsUnderSuiteAppSlashAppIdFolder` but not 100% convinced about it, so you leave it as it is.

karl-anthony-ng: You are correct. Using the relative path is not necessary. Updated to just make use of the absolute path.

I also updated the function name to activeFileIsUndersuiteAppsAppIdFolder. I didn't add the slash anymore because I think it's understandable without the slash.

davidecorreu (approved):

PR #632, opened 2022-10-14
PDPDevtool 4922 - Improve versioning management in SuiteCloud SDK

alii-gator:


packages/node-cli/src/core/sdksetup/SdkProperties.js
```diff
@@ -28,7 +28,9 @@ class SdkProperties {
 	}
 
 	getSdkFileName() {
-		return this.configFileExists() ? CONFIG_FILE_CACHE.sdkFilename : SDK_FILENAME;
+		// read config.js file if exists or use package.json
+		const configFile = this.configFileExists() ? CONFIG_FILE_CACHE : require(PACKAGE_FILE);
+		return configFile.sdkFileName;
```
karl-anthony-ng, line 33: Should we do something like this instead?
`const sdkFileName = this.configFileExists() ? CONFIG_FILE_CACHE.sdkFileName : sdkFileName`

antoniofruci (approved):

karl-anthony-ng (changes requested):

packages/node-cli/src/core/sdksetup/SdkProperties.js
alii-gator: This will require a refactor of TC build configs also right? because we are generating config files on the fly, with `sdkFilename`, so if we change that to `sdkFileName` it will cause issues I guess.

Previous I did something like that, 

`this.configFileExists() ? CONFIG_FILE_CACHE.sdkFilename : sdkFileName`  but it was looking weird

Either we make both `sdkFileName` or `sdkFilename`
and refactor TC configs also accordingly. What do you think? @karl-anthony-ng

karl-anthony-ng: No, what I meant here is that you already imported sdkFilename from package.json so there's no need to require again. You can just reference it directly. Just a mistake on the casing I made in the comment.
`const sdkFilename = this.configFileExists() ? CONFIG_FILE_CACHE.sdkFilename : sdkFileName`

And maybe we should also use sdkFilename in package.json instead of sdkFileName to make it consistent?

packages/node-cli/package.json
```diff
@@ -19,6 +19,8 @@
 	"repository": "https://github.com/oracle/netsuite-suitecloud-sdk/tree/master/packages/node-cli",
 	"main": "./src/suitecloud.js",
 	"preferGlobal": true,
+	"sdkFileName": "cli-2022.2.1.jar",
```
karl-anthony-ng, line 22: Maybe change this to sdkFilename to be consistent with what we have in the config file?

karl-anthony-ng (changes requested):

karl-anthony-ng (approved):

PR #643, opened 2022-10-19
PDPDEVTOOL-4853: Modify display messages for update file in vsCode

lfpvillegas:


packages/node-cli/src/commands/file/import/ImportFilesAction.js
```diff
@@ -80,20 +86,28 @@ module.exports = class ImportFilesAction extends BaseAction {
 				.addParams(params)
 				.build();
 
+			let statusMessage = NodeTranslationService.getMessage(MESSAGES.IMPORTING_FILES);
+			if (this._calledFromUpdate) {
+				statusMessage = NodeTranslationService.getMessage(MESSAGES.UPDATING_FILE);
```
davidecorreu, line 91: This message is not the one showed in vscode while importing (updating or comparing) this is the message showed only in node-cli where we only have the file:import command.
I think we this is not necessary please remove it.

packages/node-cli/src/services/TranslationKeys.js
```diff
@@ -140,11 +140,14 @@ module.exports = {
 			IMPORTING_FILES: 'COMMAND_IMPORTFILES_MESSAGES_IMPORTING_FILES',
 			LOADING_FILES: 'COMMAND_IMPORTFILES_MESSAGES_LOADING_FILES',
 			LOADING_FOLDERS: 'COMMAND_IMPORTFILES_MESSAGES_LOADING_FOLDERS',
+			UPDATING_FILE: 'COMMAND_IMPORTFILES_MESSAGES_UPDATING_FILE',
```
davidecorreu, line 143: Following the previous comment from `ImportFilesAction.js`  this message probably also needs to be deleted.

packages/vscode-extension/messages.json
```diff
@@ -83,6 +83,7 @@
 	"IMPORT_FILES_QUESTIONS_SELECT_FOLDER": "Select the path from where you want to import the files.",
 	"IMPORT_FILES_QUESTIONS_OVERRIDE": "The selected files in the project will be overwritten.",
 	"IMPORT_FILES_QUESTIONS_OVERRIDE_SINGLE": "The selected file in the project will be overwritten: {0}",
+	"IMPORT_FILES_UPDATING_FILE": "Updating file...",
```
davidecorreu, line 86: Following the previous comment from ImportFilesAction.js this message probably also needs to be deleted.

packages/vscode-extension/src/commands/CompareFile.ts
```diff
@@ -33,17 +33,12 @@ export default class CompareFile extends FileImportCommon {
 		if (projectInfoService.isSuiteAppProject()) {
 			if (!this.activeFileIsUnderSuiteAppsAppIdFolder(projectInfoService)) {
 				return this.unsuccessfulValidation(
-					this.translationService.getMessage(
-						COMPARE_FILE.ERROR.SUITEAPP_NOT_ALLOWED_FOLDER,
-						projectInfoService.getApplicationId()
-					)
+					this.translationService.getMessage(COMPARE_FILE.ERROR.SUITEAPP_NOT_ALLOWED_FOLDER, projectInfoService.getApplicationId())
```
davidecorreu, line 36: No action is needed for this comment. Changes in this file are only about formatting, if possible please avoid committing files with no logic changes in the code.

packages/node-cli/src/metadata/SdkCommandsMetadata.json
```diff
@@ -330,6 +330,15 @@
...
+			"calledfromupdate": {
+				"name": "calledfromupdate",
+				"option": "calledfromupdate",
+				"description": "Message displayed should be different if called from Update File.",
+				"mandatory": false,
+				"type": "FLAG",
+				"defaultOption": false,
+				"disableInIntegrationMode": false
 			}
 		}
```
davidecorreu, line 343: calledfromupdatea and calledfromcomparefiles should be in `SdkCommandsMetadataPatch.json` instead of `SdkCommandsMetadata.json`. This last one should be generated from java sdk.

packages/vscode-extension/src/commands/FileImportCommon.ts
```diff
@@ -68,18 +68,29 @@ export default abstract class FileImportCommon extends BaseAction {
...
+	getCommandArgs(
+		selectedFilesPaths: string[],
+		excludeProperties: string
+	): { excludeproperties?: string | undefined; paths: string[]; calledfromupdate?: string } {
+		return {
+			paths: selectedFilesPaths,
+			...(excludeProperties === this.translationService.getMessage(ANSWERS.YES) && { excludeproperties: 'true' }),
+		};
+	}
+
```
davidecorreu, line 93: This could be enhanced a little bit. 
I think it should be a protected method that returns `{ excludeproperties?: boolean; paths: string[]; calledfromupdate?: string }`
Then we ovewrite it from UpdateFile.ts as you already do.

And I think we should use the same approach for the statusMessage. We could create a protected getStatusMessage() that returns the `importing files...` and ovewrite it from UpdateFile.ts to return `updating file...`

lfpvillegas: excludeproperties should be a string, since this argument is passed as argument of the command.

For the rest, I would do it as you say.

antoniofruci (approved):

davidecorreu (approved):

PR #647, opened 2022-10-31
PDPDEVTOOL-4972 fix action result error value when exception occurs

antoniofruci:


packages/node-cli/src/commands/file/import/ImportFilesAction.js
```diff
@@ -105,7 +105,7 @@ module.exports = class ImportFilesAction extends BaseAction {
 						.build()
 				: ActionResult.Builder.withErrors(operationResult.errorMessages).withCommandParameters(params).build();
 		} catch (error) {
-			return ActionResult.Builder.withErrors([error]).build();
+			return ActionResult.Builder.withErrors([error.getErrorMessage()]).build();
```
vic-nik, line 108: I am wondering if there are cases when error is still the message. Should we handle both maybe?

antoniofruci: You are right. It would be nice to ensure that we are always consistent with the kind of exception that we throw, but for safety and simplicity I am going to check the type before calling the getErrorMessageType function.

antoniofruci: Pushed fix.

davidecorreu (approved):

vic-nik (approved):

PR #648, opened 2022-11-02
feature/pdpdevtool-4785 vscode show active suitecloud project and authID

davidecorreu:


packages/vscode-extension/src/startup/StatusBars.ts
```diff
@@ -0,0 +1,135 @@
...
+ * @param {vscode.TextEditor|undefined} textEditor
+ * @param {vscode.StatusBarItem} suitecloudProjectStatusBar
+ * @param {vscode.StatusBarItem} authIDStatusBar
+ */
+export async function updateStatusBars(
+	textEditor: vscode.TextEditor | undefined,
+	suitecloudProjectStatusBar: vscode.StatusBarItem,
+	authIDStatusBar: vscode.StatusBarItem
+) {
+	const activeTextEditor = textEditor;
```
antoniofruci, line 48: Maybe we can just rename the function parameter instead of re-assigning it

packages/vscode-extension/src/startup/StatusBars.ts
```diff
@@ -0,0 +1,135 @@
...
+	return suitecloudProjectStatusBar;
+}
+
+/**
+ * Create the status bar item used to display the current authid used in the active suitecloud project
+ *
+ * @returns {vscode.StatusBarItem} The authID status bar item
+ */
+export function createAuthIDStatusBar(): vscode.StatusBarItem {
+	const authIDStatusBar = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, 10);
```
antoniofruci, line 28: We should avoid 'magic numbers'

packages/vscode-extension/src/startup/StatusBars.ts
```diff
@@ -0,0 +1,135 @@
+import * as vscode from 'vscode';
```
antoniofruci, line 1: I don't have any smart suggestion, but maybe we should rethink how this file is called since the current name is not very descriptive about its usage.

packages/vscode-extension/src/startup/StatusBars.ts
davidecorreu: Renamed some of the variables to only use param `textEditor` and not re-assign it.

packages/vscode-extension/src/startup/StatusBars.ts
davidecorreu: Added new constant.

packages/vscode-extension/src/startup/StatusBars.ts
davidecorreu: Renamed to `StatusBarItemsFunctions.ts`

antoniofruci (approved):

karl-anthony-ng (approved):

PR #657, opened 2022-12-28
PDPDEVTOOL-5016 - Unit testing stubs 2

vic-nik:


packages/unit-testing/stubs/ui/dialog/dialog.js
```diff
@@ -0,0 +1,75 @@
...
+     *
+     */
+    dialog.protptype.create = function (options) { };
+
+    /**
+     * @typedef {Object} ButtonOption
+     * @property {*} label The label of the button
+     * @property {*} value The return value of the dialog if the button was clicked
+     */
+    function ButtonOptionTypedef() {
```
davidecorreu, line 67: Not in docs and not exposed but we can leave it I guess.

packages/unit-testing/stubs/ui/serverWidget/AssistantStep.js
```diff
@@ -0,0 +1,143 @@
...
+         * @restriction Server SuiteScript only
+         * @governance none
+         * @param {Object} options
+         * @param {string} options.group  The internal id of the sublist
+         * @return {Array<string>}
+         * @throws {SuiteScriptError} SSS_MISSING_REQD_ARGUMENT when group parameter is missing
+         *
+         * @since 2015.2
+         */
+        this.getSublistFieldIds = function (options) { };
```
davidecorreu, line 59: Not in table view of docs, but present in tree view. Let's keep it.

packages/unit-testing/stubs/ui/serverWidget/Field.js
```diff
@@ -0,0 +1,242 @@
...
+         * @governance none
+         * @param {Object} options
+         * @param {string} options.help The help text for the field
+         * @param {boolean} [options.showInlineForAssistant] This means that field help will appear only in a field help popup box when the field label is clicked. The default value is false, which means the field help appears in a popup when the field label is clicked and does not appear inline.
+         * @return {void}
+         * @throws {SuiteScriptError} SSS_MISSING_REQD_ARGUMENT when help parameter is missing
+         * @throws {SuiteScriptError} SSS_INVALID_TYPE_ARG when help parameter is not string
+         * @since 2015.2
+         */
+        this.setHelpText = function (options) { };
```
davidecorreu, line 194: By the docs seems that this function should return itself, so maybe the return type should be {Field}. Not sure about it.

packages/unit-testing/stubs/ui/serverWidget/Form.js
```diff
@@ -0,0 +1,403 @@
...
+        this.title = undefined;
+        /**
+         * This method is called during a beforeLoad UE or a suitelet and the message is later displayed on the client side,
+         * once the pageInit script is completed. The method takes either an already created Message object or the options
+         * object that would be used for creating the message.
+         * User event context can be used to control whether the message is shown on records in view, create, or edit mode (not applicable for Suitelets).
+         * @restriction Server SuiteScript only
+         * @governance none
+         * @param {Object|Message} options config object OR the message options object to be displayed in browser, described in N/ui/message: create()
+         * @param {Message} options.message the message object to be displayed in browser, described in N/ui/message: create()
```
davidecorreu, line 29: This Message is refers to ui/message/MessageInstance, we could inject it.

packages/unit-testing/stubs/ui/serverWidget/List.js
```diff
@@ -0,0 +1,163 @@
+define(['./ListColumn', './Button'], function (ListColumn, Button) {
```
davidecorreu, line 1: Missing to inject List.

packages/unit-testing/stubs/ui/ui.js
```diff
@@ -0,0 +1,20 @@
...
+
+    ui.prototype.dialog = dialog;
+    ui.prototype.message = message;
+    ui.prototype.serverWidget = serverWidget;
+
+    /**
+     * @exports N/ui
+     * @namespace ui
+     */
+    return new ui();
```
davidecorreu, line 19: Adding this `ui` moudule makes me think if we should also add the `N` module at the root of our stubs. It's just an open question to consider later with the team.

packages/unit-testing/stubs/ui/dialog/dialog.js
vic-nik: I will remove it, looks like an internal definition

packages/unit-testing/stubs/ui/serverWidget/Field.js
vic-nik: Not sure either, could be inconsistency in the doc as well. Looks like the functions that return Field object are mostly related to the placement of the field in the UI and this one is different. I am inclined to leave it as is for now.

packages/unit-testing/stubs/ui/serverWidget/Form.js
vic-nik: Yes

packages/unit-testing/stubs/ui/ui.js
vic-nik: It made sense to include N/ui becuase of this Help Center section:
https://nlcorp.app.netsuite.com/app/help/helpcenter.nl?fid=section_156939471908.html
Which says:
Use [‘N/ui’] as the first argument of the define function as a shortcut to load the three modules (N/ui/dialog, N/ui/message, and N/ui/serverWidget) on server scripts, or the N/ui/dialog and N/ui/message on client scripts.

I would avoid creating N module as its usage is not recommended in SuiteScript. However, if we find similar examples in the doc for other sub-modules, we should expose them. E.g. N/crypto and N/crypto/certificate

alii-gator (approved):

PR #660, opened 2022-12-29
feature/PDPDEVTOOL5021 remove redundant occurrences of auth ID

Montsant19:


packages/node-cli/src/metadata/SdkCommandsMetadataPatch.json
```diff
@@ -2,22 +2,22 @@
...
-				"description": "Prints a list of all the configured authentication IDs (authID). Usage: account:manageauth --list."
+				"description": "Prints a list of all the configured authentication IDs. Usage: account:manageauth --list."
 			},
 			"remove": {
-				"description": "Removes an authentication ID (authID). Usage: account:manageauth --remove {authID}."
+				"description": "Removes an authentication ID. Usage: account:manageauth --remove {authID}."
 			},
 			"rename": {
-				"description": "Renames an authentication ID (authID). You must specify it together with the renameto option. Usage: account:manageauth --rename {authID} --renameto {newauthID}."
+				"description": "Renames an authentication ID. You must specify it together with the renameto option. Usage: account:manageauth --rename {authID} --renameto {newAuthID}."
```
davidecorreu, line 14: We should go for the long versions to be as similar as possible with java-cli.

- `{auhtID}` should be `{authenticationID}`
- `{newAuthID}` should be `{newAuthenticationID}`

Same applies to line 17.

davidecorreu:
Should we also change this type of messages that are shown with the spinner while performing some action?
`"COMMAND_DEPLOY_MESSAGES_DEPLOYING": "Deploying project \"{0}\" with auth ID \"{1}\"..."`

packages/node-cli/src/metadata/SdkCommandsMetadataPatch.json
Montsant19: Yes, this is true. Changing it.

Montsant19: But generally speaking, I think it is fine to use either short or long form. Remember that we mention that the docs use both short and long forms. So I wouldn't change the spinner message below. Or if we want to go for the long form in all cases, I would address it in a new iteration (but I really don't think it is necessary).

davidecorreu (approved):

vic-nik (approved):

PR #742, opened 2023-07-24
update libraries

davidecorreu:


packages/vscode-extension/package.json
```diff
@@ -139,12 +139,12 @@
 				"command": "suitecloud.updateobject",
 				"key": "ctrl+alt+u",
 				"mac": "cmd+alt+u",
-				"when": "resource =~ //Objects/.+\\.xml$/"
+				"when": "resource =~ /\\/Objects\\/.+\\.xml$/"
```
karl-anthony-ng, line 142: Are the changes here intentional?

davidecorreu: Those changes where in my workspace, I forgot to remove them.
They are the solution to not show the specific command in the context menu in any place as those `when` clauses are kind of broken.
I can leave them or remove them, as you wish. I did test them in my Windows machine and worked fine, but maybe is better to leave as it was.

karl-anthony-ng: I think we already have a separate issue to work on that part. Maybe better to move these changes to that issue?

davidecorreu: I'll remove those changes then. 👍

karl-anthony-ng (approved):

antoniofruci (approved):

PR #805, opened 2024-04-25
PDPDEVTOOL-5637 reauth flow node-cli

davidecorreu:
add refresh authorization functionallity to node-cli


packages/node-cli/src/utils/AuthenticationUtils.js
```diff
@@ -160,5 +174,43 @@ async function authenticateCi(params, sdkPath, projectFolder, executionEnvironme
...
 		.build();
 }
+/**
+ * 
+ * @param {String} authid 
+ * @param {String} sdkPath 
+ * @param {ExecutionEnvironmentContext} executionEnvironmentContext 
+ * @returns {SdkOperationResult}
+ */
+async function checkIfReauthzorizationIsNeeded(authid, sdkPath, executionEnvironmentContext) {
```
karl-anthony-ng, line 184: Was it intentional to have the "z" in authzorization?

packages/node-cli/src/utils/AuthenticationUtils.js
```diff
@@ -160,5 +174,43 @@ async function authenticateCi(params, sdkPath, projectFolder, executionEnvironme
...
+	const sdkExecutor = new SdkExecutor(sdkPath, executionEnvironmentContext);
+	const inspectAuthContext = SdkExecutionContext.Builder.
+		forCommand(COMMANDS.INSPECT_AUTHORIZATION.SDK_COMMAND)
```
karl-anthony-ng, line 187: Please put the dot "." from the previous line here so make it uniform with the other method calls below this.

packages/node-cli/src/utils/SdkOperationResult.js
```diff
@@ -0,0 +1,68 @@
+/*
```
karl-anthony-ng, line 1: May I know the reason why there is a need to create SdkOperationResult instead of just using OperationResult?

packages/node-cli/src/utils/AuthenticationUtils.js
davidecorreu: It's a mistake. I'll fix the typo.

packages/node-cli/src/utils/AuthenticationUtils.js
davidecorreu: Sure. I didn't realize about it. I'll change it.

packages/node-cli/src/utils/SdkOperationResult.js
davidecorreu: It's a convinient way to kind of "type" via jsdocs the return type of an executed commad via sdkExecutor.
If you open the code in an editor you'll see that there is some "code completion" available when using the recently created methods on AuthenticationUtils.js
Also we can add some useful functionality as the .isSuccess() method, instead of comparing `operationResult.status === SdkOperationResultUtils.STATUS.SUCCESS`. There is a comment about it on the `src\utils\SdkOperationResultUtils.js`

karl-anthony-ng: Understood.

karl-anthony-ng (approved):

antoniofruci (approved):

PR #813, opened 2024-07-03
PDPDEVTOOL-5593: [NSN] vscode - oauth2.0 - re-authorize flow

karl-anthony-ng:


packages/vscode-extension/src/commands/BaseAction.ts
```diff
@@ -150,6 +161,10 @@ export default abstract class BaseAction {
 	}
 
 	protected async runSuiteCloudCommand(args: { [key: string]: string | string[] } = {}, otherExecutionPath?: string) {
+		if (this.refreshAuthorizationPeformed) {
+			args.skipAuthorizationCheck = BaseAction.SKIP_AUTHORIZATION_CHECK_VALUE;
+		}
+
```
davidecorreu, line 167: We always want to have `args.skipAuthorizationCheck` so node-cli code never executes the check, it doesn't matter if the refresh authz has been performed or was not needed.

When working on node-cli story I thought we will be adding the `skipAuthorizationCheck` directly in the `src/core/SuiteCloudRunner.ts`

antoniofruci (approved):

packages/vscode-extension/src/commands/BaseAction.ts
karl-anthony-ng: You are correct. Please check changes.

davidecorreu (approved):

PR #817, opened 2024-07-11
PDPDEVTOOL-5587: [NSN] vscode - oauth2.0 - add the new command authenticate ci

karl-anthony-ng:


antoniofruci (approved):

packages/vscode-extension/src/commands/SetupAccount.ts
```diff
@@ -233,37 +236,28 @@ export default class SetupAccount extends BaseAction {
...
-			password: true,
-			validateInput: (fieldValue) => {
-				let validationResult = InteractiveAnswersValidator.showValidationResults(
-					fieldValue,
-					InteractiveAnswersValidator.validateFieldIsNotEmpty
-				);
-				return typeof validationResult === 'string' ? validationResult : null;
-			},
+	private async getPrivateKeyFilePath() {
+		return window.showOpenDialog({
```
davidecorreu, line 255: Let's try to explicitly use some of the `window.showOpenDialog`  options so it will be clearer what can be returned.
I would like to make explicit use of:
- `OpenDialogOptions.canSelectMany=true`
- `OpenDialogOptions.canSelectFolders=false`
- `OpenDialogOptions.canSelectMany=false`

packages/vscode-extension/src/commands/SetupAccount.ts
```diff
@@ -279,45 +273,53 @@ export default class SetupAccount extends BaseAction {
...
 
-		const saveTokenPromise = AuthenticationUtils.saveToken(commandParams, getSdkPath(), this.rootWorkspaceFolder);
+		const authenticateCiPromise = AuthenticationUtils.authenticateCi(
+			commandParams,
+			getSdkPath(),
+			this.rootWorkspaceFolder,
+			new ExecutionEnvironmentContext({
+				platform: VSCODE_PLATFORM,
+				platformVersion: vscode.version,
+			}),
```
davidecorreu, line 303: Shouldn't we add this execEnv the browser based authentication?

packages/vscode-extension/src/commands/SetupAccount.ts
karl-anthony-ng: I made the changes as you requested.

But just to note, I intentionally did not add those as they were optional and defaulted to the values we needed.

packages/vscode-extension/src/commands/SetupAccount.ts
karl-anthony-ng: I think we should. It's looks like a bug from the original implementation.

davidecorreu (approved):

PR #821, opened 2024-07-25
node-cli dependencies version bump

antoniofruci:


packages/node-cli/package.json
```diff
@@ -32,7 +32,7 @@
 		"chalk": "4.1.2",
 		"cli-spinner": "0.2.10",
 		"commander": "11.1.0",
-		"inquirer": "8.2.6",
+		"inquirer": "10.1.2",
```
davidecorreu, line 35: Is this correct? Does the interactive questions work ok with this update?

karl-anthony-ng (approved):

packages/node-cli/package.json
antoniofruci: You're right. Reverted

PR #826, opened 2024-07-26
Bug Fixes for PDPDEVTOOL-5587: [NSN] vscode - oauth2.0 - add the new command authenticate ci

karl-anthony-ng:
Fixes the following...
- PDPDEVTOOL-5736: [NSN] vscode-extension fails to authenticate via m2m against scrumboxes
- Improve Command Paletter behavior for selecting the Private Key File.


antoniofruci (approved):

packages/vscode-extension/src/commands/SetupAccount.ts
```diff
@@ -256,6 +256,14 @@ export default class SetupAccount extends BaseAction {
 	}
 
 	private async getPrivateKeyFilePath() {
+		if (!await window.showQuickPick(['Select private key file'], {
```
davidecorreu, line 259: Please use translation service for string values.

davidecorreu: Maybe we could reuse a modified version of `SETUP_ACCOUNT_CREATE_M2M_SELECT_PRIVATE_KEY_FILE`

karl-anthony-ng: I forgot about. Pushed changes.

davidecorreu (approved):

PR #871, opened 2025-01-13
PDPDEVTOOL-5944 - Fix execution context for browser based auth

carolnavarropereira:


packages/node-cli/src/services/ExecutionContextService.js
```diff
@@ -0,0 +1,92 @@
...
+
+const ExecutionMode = {
+	CI: 'CI',
+	AUTH_CI_SETUP: 'AUTH_CI_SETUP',
+	DEV_MACHINE: 'DEV_MACHINE',
+	DEV_MACHINE_FALLBACK_PASSKEY: 'DEV_MACHINE_FALLBACK_PASSKEY',
+};
+
+const validateMachineToMachineAuthIsAllowed = () => {
+	let executionMode = getExecutionMode();
```
karl-anthony-ng, line 31: Use const whenever the value is not expected to be changed. Please do the same for same cases below.

packages/vscode-extension/src/commands/SetupAccount.ts
```diff
@@ -101,10 +101,24 @@ export default class SetupAccount extends BaseAction {
...
+				await this.handleM2m(accountCredentialsList);
+			}
+		}
+	}
+
+	private validateSupportedMode(validatorFunction: () => void): boolean {
+		try {
+			validatorFunction();
+		} catch (err: any) {
+				this.messageService.showCommandError(err);
```
karl-anthony-ng, line 118: I think there's an extra indentation here.

karl-anthony-ng (changes requested):

packages/vscode-extension/src/commands/SetupAccount.ts
carolnavarropereira: Thanks, changed

packages/node-cli/src/services/ExecutionContextService.js
carolnavarropereira: Thank you. I will do it.

packages/vscode-extension/src/commands/SetupAccount.ts
```diff
@@ -193,8 +203,9 @@ export default class SetupAccount extends BaseAction {
 		const actionResult = await authenticatePromise;
 		this.handleAuthenticateActionResult(actionResult);
 
-		if (actionResult.status === actionResultStatus.SUCCESS && process.env.SUITECLOUD_FALLBACK_PASSKEY) {
-			this.messageService.showCommandWarning(this.translationService.getMessage(MANAGE_ACCOUNTS.WARNING.ROTATE_PASSWORD_WARNING));
+		let warning = ExecutionContextService.getBrowserBasedWarningMessages();
+		if (warning) {
```
karl-anthony-ng, line 207: Can probably be const here as well.

karl-anthony-ng (approved):

packages/node-cli/src/services/ExecutionContextService.js
```diff
@@ -0,0 +1,92 @@
...
+	const executionMode = getExecutionMode();
+	if (executionMode === ExecutionMode.DEV_MACHINE_FALLBACK_PASSKEY) {
+		return NodeTranslationService.getMessage(ROTATE_PASSWORD_WARNING);
+	}
+};
+
+const getExecutionMode = () => {
+	const envFallBackPassKey = process.env.SUITECLOUD_FALLBACK_PASSKEY;
+	const envCiPassKey = process.env.SUITECLOUD_CI_PASSKEY;
+	const envCi = process.env.SUITECLOUD_CI;
```
davidecorreu, line 53: I think we should have all the env variables name in constants, probably into 'ApplicationConstant.js`

```
{
   ...
   ENV_VARS: {
      SUITECLOUD_CI: 'SUITECLOUD_CI',
      SUITECLOUD_CI_PASSKEY: 'SUITECLOUD_CI_PASSKEY',
      SUITECLOUD_FALLBACK_PASSKEY: 'SUITECLOUD_FALLBACK_PASSKEY
   }
   ...
}
```

Like this or in similar way like `SUITECLOUD_ENV: {CI, CI_PASSKEY, FALLBACK_PASSKEY}`

packages/vscode-extension/src/commands/SetupAccount.ts
```diff
@@ -101,10 +101,24 @@ export default class SetupAccount extends BaseAction {
 		if (!selected) {
 			return;
 		} else if (selected.option === UiOption.new_authid_browser) {
-			await this.handleBrowserAuth(accountCredentialsList);
+			if (this.validateSupportedMode(ExecutionContextService.validateBrowserBasedAuthIsAllowed)) {
```
davidecorreu, line 104: Is the `validateSupportedMode` wrapping function really needed? I guess yes, but can you explain why?
If not needed I would just call `ExecutionContextService.validateBrowserBasedAuthIsAllowed` a let throw error propagated.
If any case I think we could reduce one nested level by using:
`else if (selected.option === UiOption.new_authid_browser && this.validateSupportedMode(ExecutionContextService.validateBrowserBasedAuthIsAllowed))`

carolnavarropereira: The error message does not show it

packages/node-cli/src/services/ExecutionContextService.js
carolnavarropereira: Done

packages/vscode-extension/src/commands/SetupAccount.ts
davidecorreu: You have missed the second part of the comment:
- we could reduce one nested level by combining 2 conditions in the `else if` statement 
`else if (selected.option === UiOption.new_authid_browser && this.validateSupportedMode(ExecutionContextService.validateBrowserBasedAuthIsAllowed))`

The same could be applied for the next`else if` on line 107.

carolnavarropereira: Technically it is not the best way, because if 
selected.option === UiOption.new_authid_browser is true and 
this.validateSupportedMode(ExecutionContextService.validateBrowserBasedAuthIsAllowed) is false it will go to the second else if, of course it would fail, since selected.option === UiOption.new_authid_m2m is false, however technically it is more correct if you check the type and inside the block related to the type i.e. browser, you check the validation.

davidecorreu (approved):

PR #873, opened 2025-01-15
PDPDEVTOOL-5872 Release JS Tools

carolnavarropereira:
Upgrade library versions

### DIFF
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -20,7 +20,7 @@
 		"directory": "packages/vscode-extension"
 	},
 	"engines": {
-		"vscode": "^1.91.1"
+		"vscode": "^1.96.0"
 	},
 	"categories": [
 		"Other"
@@ -196,7 +196,7 @@
 	},
 	"devDependencies": {
 		"@types/mocha": "10.0.10",
-		"@types/node": "22.10.6",
+		"@types/node": "20.17.14",
 		"@types/vscode": "1.96.0",
 		"@typescript-eslint/eslint-plugin": "8.20.0",
 		"@typescript-eslint/parser": "8.20.0",
--- a/packages/vscode-extension/package.json
+++ b/packages/vscode-extension/package.json
@@ -195,15 +195,15 @@
 		"@oracle/suitecloud-cli": "^2.1.0"
 	},
 	"devDependencies": {
-		"@types/mocha": "10.0.7",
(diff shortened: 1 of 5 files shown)


davidecorreu:
@carolnavarropereira Don't we have anything to update on `packages/node-cli/package.json` ?

packages/vscode-extension/package.json
```diff
@@ -195,15 +195,15 @@
 		"@oracle/suitecloud-cli": "^2.1.0"
 	},
 	"devDependencies": {
-		"@types/mocha": "10.0.7",
-		"@types/node": "20.14.12",
-		"@types/vscode": "1.91.0",
-		"@typescript-eslint/eslint-plugin": "7.17.0",
-		"@typescript-eslint/parser": "7.17.0",
+		"@types/mocha": "10.0.10",
+		"@types/node": "22.10.6",
```
davidecorreu, line 199: I think we should go for `@types/node` in sync with vscode node version which currently is 20.
Screenshot 2025-01-16 at 09 27 39

packages/vscode-extension/package.json
```diff
@@ -195,15 +195,15 @@
...
+		"@types/vscode": "1.96.0",
```
davidecorreu, line 200: I think this is the right version for it, but I think we should also update line 23 to match the same version. I think we won't be able to build the extension file (.vsix) if these values are not in sync.
```
	"engines": {
		"vscode": "^1.91.1"
	},
```

carolnavarropereira:
> @carolnavarropereira Don't we have anything to update on `packages/node-cli/package.json` ?
chalk, commander and inquirer have newer versions but we cannot update them because it would required to configuration the setup of the application for all the packages

Error [ERR_REQUIRE_ESM]: require() of ES Module (...) is not supported. Instead change the require of inquirer.js in (...)  to a dynamic import() which is available in all CommonJS modules.

packages/vscode-extension/package.json
carolnavarropereira: so, should we update the line 23? (I think it is the best option)

packages/vscode-extension/package.json
carolnavarropereira: Updated to 20.18.1 to match exactly the version

karl-anthony-ng (approved):

davidecorreu (approved):

PR #877, opened 2025-01-24
Feature/pdpdevtool 5988 update readmes

Montsant19:


packages/vscode-extension/README.md
```diff
@@ -13,7 +13,7 @@ Read more about SuiteCloud SDK and SDF in <a href="https://docs.oracle.com/en/cl
 
 ## Installation Prerequisites
 The following software is required to work with SuiteCloud Extension for Visual Studio Code:
-- Oracle JDK version 17
+- Oracle JDK version 17 and 21
 - VS Code version 1.91.1 or higher
```
karl-anthony-ng, line 17: Please change this to 1.96.0

packages/vscode-extension/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## 3.0.0
+- Secure Credentials Storage for SuiteCloud SDK
+- Removal of OAuth 1.0 Token-Based Authentication support in SuiteCloud SDK
```
davidecorreu, line 5: This is not correct, lets remove the second line.
From past release we only work with OAuht2.0 in vscode-extension and in suitecloud-sdk tools in general.

packages/vscode-extension/README.md
TheJorgeBorras: If we are talking about prerequisits we only require either 17 OR 21.
`Oracle JDK version 17 and 21` -> `Oracle JDK version 17 or 21`

packages/node-cli/README.md
```diff
@@ -13,8 +13,8 @@ CLI for Node.js is an interactive tool that guides you through all the steps of
 
 ## Prerequisites
 The following software is required to work with SuiteCloud CLI for Node.js:
-- Node.js version 20 LTS
-- Oracle JDK version 17
+- Node.js version 22 LTS
+- Oracle JDK version 17 and 21
```
TheJorgeBorras, line 17: Same issue as before, I think should be **or** instead of **and**.

davidecorreu (approved):

PR #885, opened 2025-01-31
-PDPDEVTOOL-5947 vscode-extension - Remove the See Details button on warnings and errors

carolnavarropereira:
Fallback warning now uses a "dismiss" instead of a see details button
M2M/BWB not allowed removed the see details button


packages/vscode-extension/src/service/MessageService.ts
```diff
@@ -52,6 +52,12 @@ export default class MessageService {
 			);
 	}
 
+	showWarningMessageDismiss(warningMessage: string, includeProjectName = true) {
```
karl-anthony-ng, line 55: Suggest to rename it to showWarningMessageWithDismiss.

karl-anthony-ng (changes requested):

packages/vscode-extension/src/service/MessageService.ts
carolnavarropereira: Now is going to be Ok at the end, so showWarningMessageWithOk will do

davidecorreu (approved):

karl-anthony-ng (approved):

PR #912, opened 2025-05-06
PDPDEVTOOL-5724: [Node-cli] New non interactive Switch Account Command

TheJorgeBorras:
New command to change `defaultAccountId` for current project (changes value in `project.json`).
* Non interactive to allow users specify which account to use in CI/CD environments.
* Doesn't interact with specific Environment Variables (works for CI/CD, Browser-Based and Fallback).
* Doesn't check if AuthID exists or not (just checks for no special chars, spaces, max length, etc.).


packages/node-cli/messages.json
```diff
@@ -195,6 +195,8 @@
 
 	"COMMAND_SETUPACCOUNTCI_OUTPUT_NEW_OAUTH": "The authentication was successful with the following account and role: {0} [{1}]. This project will use the authentication ID \"{2}\" as default.",
 
+	"COMMAND_SWITCHACCOUNT_OUTPUT_NEW_DEFAULT_ACCOUNT": "[Placeholder Message 1a] Successfully completed Authentication Switch action \"{0}\".",
```
antoniofruci, line 198: What is `[Placeholder Message 1a]` ?

packages/node-cli/src/commands/account/switch/SwitchAction.js
antoniofruci, line 1: We should remove all the TODOs and unneeded comments from the file.
Also consider if we should update the copyright year to 2025 to the newly created files.

packages/node-cli/src/commands/account/switch/SwitchAction.js
```diff
@@ -0,0 +1,82 @@
...
+const SwitchAccountActionResult = require('../../../services/actionresult/SwitchAccountActionResult');
+const { setDefaultAuthentication } = require('../../../utils/AuthenticationUtils');
+const ProjectInfoService = require('../../../services/ProjectInfoService');
+const {
+	validateFieldHasNoSpaces,
+	validateFieldIsNotEmpty,
+	validateAlphanumericHyphenUnderscore,
+	validateMaximumLength,
+	showValidationResults,
+} = require('../../../validation/InteractiveAnswersValidator');
```
antoniofruci, line 17: Are these the correct validators? I see we import from `InteractiveAnswersValidator`. If there aren't validators for non interactive commands ignore this comment.

packages/node-cli/src/commands/account/switch/SwitchOutputHandler.js
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+		parse(actionResult) {
+			const resultMessage = NodeTranslationService.getMessage(
+				COMMAND_SWITCHACCOUNT.OUTPUT.NEW_DEFAULT_ACCOUNT,
+				actionResult.authId
+			);
+
+			this._log.result(resultMessage);
+			this._log.result("[Placeholder Message 2b] Successful account switch.");
```
antoniofruci, line 22: What is `[Placeholder Message 2b]` ?

packages/node-cli/src/metadata/NodeCommandsMetadata.json
```diff
@@ -15,5 +15,23 @@
 				"disableInIntegrationMode": false
 			}
 		}
+	},
+	"account:switch": {
+		"name": "account:switch",
+		"alias": "sw",
+		"description": "[Verify With Montsant] Sets up the default account to use with SuiteCloud SDK on this SuiteCloud project.",
```
antoniofruci, line 22: Please add @Montsant19 to the code review.

antoniofruci (changes requested):

packages/node-cli/src/commands/account/switch/SwitchAction.js
TheJorgeBorras: My bad on the copyright. Pushed the fix.

For the //TODOs, you are right and I won't be merged into the Dev branch (I will erase them right before).

I like to set them up as notes for whoever is reading the code and myself to not forget to ask specific questions.
To ask for confirmation and such.
VSCode/WebStorm highlights them, so they stand out and are hard to miss.

packages/node-cli/src/commands/account/switch/SwitchAction.js
TheJorgeBorras: There are not specific validators for non-interactive commands (as far as I can tell). I believe Java handles that.

I'm kinda doing something dirty with the output of showValidationResults(...) so I'm not against creating a specific validator class for non-interactive mode (it might fall outside of the scope of the story ).
@davidecorreu

packages/node-cli/src/commands/account/switch/SwitchOutputHandler.js
TheJorgeBorras: Same as `Placeholder Message 2b`.
Removed it in the newest commit.

packages/node-cli/src/commands/account/switch/SwitchAction.js
```diff
@@ -0,0 +1,85 @@
...
+
+	//TODO: [STEP 2] Suitecloud calls Command._validateActionParameters(preExec) and validates command options
+
+	async execute(params) {
+		try {
+
+			//TODO: [STEP 3] Validate AuthId (Do we want to do it)?
+			let validationErrors= this._validateParams(params);
+			if (validationErrors.length > 0) {
+				validationErrors = validationErrors.map(error => `Invalid auth Id: ${error}`); //TODO: (* __ *)
```
davidecorreu, line 50: I would go for a message like "There are validation errors:\n   authid: {returned error messager from showValidationResults}"

TheJorgeBorras: Fixed in last commits.
Agreed on: 
```
if (validateResult !== true) {
	throwValidationException(['authid: ' + validateResult], false, this._commandMetadata);
}
```


Which would return an error message like:
```
There are validation errors:
authid: This field contains forbidden characters. Use only alphanumerical characters, hyphens, or underscores.
```

packages/node-cli/src/commands/account/switch/SwitchAction.js
TheJorgeBorras: After discussing it further, it seems that **there is not any sort of non-interactive specific validation in place** (at the Suitecloud-CLI level). Whatever validation we do, we do it on Interactive commands, using the interactive commands validators or in the JAVA CLI layer.

There seems to be a larger issue with this. 
Multiple Bugs have been detected due to missing validation for non interactive commands. For example `account:setup:ci` allows for authids with "special characters", "spaces" and "more than 40 characters".

Fixing and testing this falls outside of the scoope of the story. We reused the interactive validation logic, documented these 'bugs' and we're keeping track of it.

packages/node-cli/src/commands/account/switch/SwitchAction.js
TheJorgeBorras: Addressed and deleted these //TODO's.
Code should be cleaner now :+1:

packages/node-cli/messages.json
TheJorgeBorras: Addressed and deleted this Placeholder.
Went with: 
```
"The auth ID was successfully changed. This project will use the authentication ID \"{0}\" as default."
```

packages/node-cli/src/services/actionresult/SwitchAccountActionResult.js
```diff
@@ -0,0 +1,55 @@
...
+class SwitchAccountActionResult extends ActionResult {
+	constructor(parameters) {
+		super(parameters);
+		this._authId = parameters.authId;
+	}
+
+	validateParameters(parameters) {
+		super.validateParameters(parameters);
+		if (parameters.status === STATUS.SUCCESS) {
+			assert(parameters.authId, '_authId is required when ActionResult is a success.');
```
davidecorreu, line 18: Can we remove the underscore from the assert message (`_authId`) ?

packages/node-cli/src/commands/account/switch/SwitchAction.js
```diff
@@ -0,0 +1,70 @@
...
+		const validateResult = showValidationResults(
+				authId,
+				validateFieldIsNotEmpty,
+				validateFieldHasNoSpaces,
+				validateAlphanumericHyphenUnderscore,
+				validateMaximumLength
+		);
+
+		if (validateResult !== true) {
+			throwValidationException(['authid: ' + validateResult], false, this._commandMetadata);
```
davidecorreu, line 67: Althoug the mesage is short, we should get it from NodeTranslationService.

davidecorreu (approved):

antoniofruci (approved):

PR #917, opened 2025-06-30
DPDEVTOOL-6104: Update chalk and inquire libraries to the last version

carolnavarropereira:


packages/node-cli/src/commands/object/create/CreateObjectInputHandler.js
```diff
@@ -27,53 +27,53 @@ module.exports = class CreateObjectInputHandler extends BaseInputHandler {
...
+		// 		{
+		// 			type: 'list',
+		// 			name: 'relatedfiledestinationfolder',
+		// 			message: (answers_9) => `Where would you like to store the ${chalk.green.bold(answers_9.relatedfilename)}.js file?`,
+		// 			choices: fileCabinetDirectoryChoices,
+		// 		},
+		// 	]);
+		// 	return { ...answers_10, ...mainAnswers };
+		// }
+		// return mainAnswers;
```
TheJorgeBorras, line 80: If I'm not mistaken, we've commented this class because we deprecated the associated command.

I'm fine with this decision, but can we add a comment on top of the class to Document this decision.

davidecorreu: I would either add the comment or remove the whole file.

packages/node-cli/src/commands/file/upload/UploadFilesInputHandler.js
```diff
@@ -4,7 +4,7 @@
  */
 'use strict';
 
-const { prompt } = require('inquirer');
+const { default: { prompt} } = require('inquirer');
```
TheJorgeBorras: Missed a space here between prompt and the closing bracket.

`const { default: { prompt} } = require('inquirer');` => const { default: { prompt } } = require('inquirer');

package-lock.json
```diff
@@ -17,6 +17,7 @@
 			"version": "2.3.0",
 			"resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
 			"integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
+			"license": "Apache-2.0",
```
TheJorgeBorras, line 20: Not a very big deal, but we are now adding the '**license**' field for each of the dependencies listed in the `package-lock.json`.

Is this something we **need** to do? Is it something we **want** to do?
@davidecorreu  what do you think?

packages/node-cli/src/commands/object/create/CreateObjectInputHandler.js
TheJorgeBorras: yes, same here

package-lock.json
davidecorreu: I would say that this something that has been added when using a newer version of npm. I would keep it.

TheJorgeBorras: Seems to be something that **NPM 10.3.0 and onward** does by default.

It's the first time we see it in our repo, but it's not something we should concern ourselves with.

packages/node-cli/src/loggers/NodeConsoleLogger.js
```diff
@@ -4,35 +4,50 @@
...
+			this._println(message, ERROR);
+		});
 	}
 
 	_println(message, color, isBold) {
-		console.log(this._formatString(message, { color: color, bold: isBold }));
+		let formatterString = this._formatString(message, { color: color, bold: isBold });
+		formatterString.then(chalkFormatter => {
+			console.log(chalkFormatter);
+		});
```
davidecorreu: I guess there is some reason to avoid using `async/await` syntax. But can we rename `chalkFormatter` to `formattedString` or `chalkFormattedString` to make it a little bit more readable.

Another part I'm not sure how it's used is the `isBold` do you know why we should keep it?

packages/node-cli/src/commands/file/upload/UploadFilesInputHandler.js
carolnavarropereira: Applied formatter

packages/node-cli/src/loggers/NodeConsoleLogger.js
carolnavarropereira: return bold(color(str));  to me more than a formatted string it looks like a function or a wrapper, anyway, I can use formattedString, but I will try to avoid the name of the library and use the functionality.

carolnavarropereira: isBold is used in _formatString
	async _formatString(str, options) {
		const color = options.color || fontFormatterPromise.DEFAULT;
		//if bold, apply BOLD
                const bold = options.bold ? fontFormatterPromise.BOLD : str => str;
		return bold(color(str));
	}
if it is bold, BOLD it is applied.

TheJorgeBorras (approved):

packages/node-cli/src/loggers/NodeConsoleLogger.js
```diff
@@ -4,35 +4,50 @@
...
+			this._println(message, ERROR);
+		});
 	}
 
 	_println(message, color, isBold) {
-		console.log(this._formatString(message, { color: color, bold: isBold }));
+		let formatterString = this._formatString(message, { color: color, bold: isBold });
+		formatterString.then(formattedString => {
+			console.log(formattedString);
+		});
```
davidecorreu: I guess there is some reason to avoid using async/await syntax. But can we rename `chalkFormatter` to `formattedString` or `chalkFormattedString` to make it a little bit more readable.

Another part I'm not sure how it's used is the isBold do you know why we should keep it?

packages/node-cli/src/loggers/NodeConsoleLogger.js
```diff
@@ -4,35 +4,50 @@
...
 	}
 
-	_formatString(str, options) {
-		const color = options.color || COLORS.DEFAULT;
-		const bold = options.bold ? chalk.bold : str => str;
+	async _formatString(str, options) {
+		const color = options.color || fontFormatterPromise.DEFAULT;
+		const bold = options.bold ? fontFormatterPromise.BOLD : str => str;
```
davidecorreu: why in this case we are not chaining the`fontFormatterPromise` with `then`?

`fontFormatterPromise.DEFAULT` doesn't seem to be valid, shoudn't it be `fontFormatterPromise.COLORS.DEFAULT`

davidecorreu: And connected to a line 42, is `options.bold` ever really used?

packages/node-cli/src/commands/object/create/CreateObjectInputHandler.js
carolnavarropereira: Comment added

packages/node-cli/src/loggers/NodeConsoleLogger.js
carolnavarropereira: I removed the unnecessary code, currently only calls _println and _formatString from inside NodeConsleLogger using only the color, I guess this is leftovers from previous code. I removed the default color and the bold.

packages/node-cli/src/loggers/NodeConsoleLogger.js
carolnavarropereira: use the word chalk is tied to the library, I would rather leave formattedString since it is more functional, I mean, what is doing, not the library used for doing so, so I renaved as formattedString

packages/node-cli/src/loggers/NodeConsoleLogger.js
carolnavarropereira: It is not used currently so I removed it.

PR #920, opened 2025-07-16
PDPDEVTOOL-6172 Expand account:setup:ci to change the default auth id in ci mode

carolnavarropereira:


packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
```diff
@@ -23,18 +25,22 @@ const COMMAND = {
...
 
 	preExecute(params) {
 		this._projectInfoService.checkWorkingDirectoryContainsValidProject(this._commandMetadata.name);
-		
+
 		if (params[COMMAND.OPTIONS.ACCOUNT]) {
 			params[COMMAND.OPTIONS.ACCOUNT] = params[COMMAND.OPTIONS.ACCOUNT].toUpperCase();
 		}
+
+		const validator = new AccountSetupCiValidation(this._commandMetadata, this._runInInteractiveMode);
```
davidecorreu: Could create this one in the constructor?
Although we are not using injection now I think we could create the "service" in the constructor similar to ProjectInfoService.

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
```diff
@@ -0,0 +1,123 @@
...
+const NOT_MANDATORY_STANDARD_MODE_PARAMS = [COMMAND.OPTIONS.DOMAIN, URL_PARAM];
+const { COMMAND_OPTIONS : { IS_MANDATORY, IS_MANDATORY_SELECT_MODE, IS_NOT_ALLOWED_SELECT_MODE} } = require('../../../services/TranslationKeys');
+
+class AccountSetupCiValidation {
+	constructor(commandMetadata, runInInteractiveMode) {
+		this._commandMetadata = commandMetadata;
+		this._runInInteractiveMode = runInInteractiveMode;
+	}
+
+	validateAuthID(authId) {
```
davidecorreu: Can we rename method to `validateAuthIDFormat` / `isValidAuthIDFormat` or similar?

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
```diff
@@ -0,0 +1,123 @@
...
+		AUTHID: 'authid',
+		CERTIFCATEID: 'certificateid',
+		PRIVATEKEYPATH: 'privatekeypath',
+		DOMAIN: 'domain',
+
+	},
+	SDK_COMMAND: 'authenticateci',
+};
+//url is a not mandatory parameter for setup mode
+const URL_PARAM = 'url';
```
davidecorreu: This is not a node-cli option, please remove.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
```diff
@@ -43,6 +49,15 @@ module.exports = class AccountSetupCiAction extends BaseAction {
 			delete params[COMMAND.OPTIONS.DOMAIN];
 		}
 		validateMachineToMachineAuthIsAllowed();
```
TheJorgeBorras, line 38: Is this flow correct?

I would expect us to first check the Execution Context, then validate if the provided `authId` is valid or not (like we do for `account:setup` browser_base mode).
In other words, ensuring the proper CI/CD env variables are set and only then execute the authid Validations.


As a side note I would prefer the validation of `authId` to take place inside the `execute()` call, not on the `preExecute()`

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
```diff
@@ -0,0 +1,123 @@
...
+//Select indicator, the action behaves in two different ways depending on select parameter is selected or not
+const SELECT_PARAM = 'select';
```
davidecorreu: Add `SELECT_PARAM` to `COMMAND.OPTIONS` as `SELECT` to follow same logic.

```
const COMMAND = {
	OPTIONS: {
		ACCOUNT: 'account',
		AUTHID: 'authid',
		CERTIFCATEID: 'certificateid',
		PRIVATEKEYPATH: 'privatekeypath',
		DOMAIN: 'domain',
                SELECT: 'select'

	},
	SDK_COMMAND: 'authenticateci',
};

File: packages/node-cli/src/metadata/SdkCommandsMetadata.json [hunk 6]
```diff
@@ -109,11 +109,21 @@
...
+			},
+			"select" : {
+				"name": "select",
+				"option": "select",
+				"description": "Selects authorization id",
+				"mandatory": false,
+				"type": "FLAG",
+				"usage": "",
+				"defaultOption": false,
+				"disableInIntegrationMode": false
```
davidecorreu: This should go to another metadata file `packages/node-cli/src/metadata/SdkCommandsMetadataPatch.json`

And same for the `"mandatory": false` changes

packages/node-cli/src/services/TranslationKeys.js
```diff
@@ -255,6 +255,8 @@ module.exports = {
 		HELP: 'COMMAND_OPTIONS_HELP',
 		INTERACTIVE_HELP: 'COMMAND_OPTIONS_INTERACTIVE_HELP',
 		IS_MANDATORY: 'COMMAND_OPTIONS_IS_MANDATORY',
+		IS_MANDATORY_SELECT_MODE: 'COMMAND_OPTIONS_IS_MANDATORY_SELECT_MODE',
+		IS_NOT_ALLOWED_SELECT_MODE: 'COMMAND_OPTIONS_IS_NOT_ALLOWED_SELECT_MODE',
```
davidecorreu: This are general command options, this specific one for `account:setup:ci` should be part of `COMMAND_SETUPACCOUNTCI`

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
```diff
@@ -0,0 +1,123 @@
...
+		return validationErrors;
+	}
+
+	_validateActionParametersSelectMode(params) {
+		//authId parameter, it is mandatory both in default and select mode
+		const AUTH_ID_PARAM = COMMAND.OPTIONS.AUTHID;
+		const validationErrors = [];
+		//it is going to be rejected anyway by the standard CommandOptionsValidator since it is mandatory
+		//it is validated also here for consistency reasons and to be able to customize the message
+		if (params[AUTH_ID_PARAM] === null || params[AUTH_ID_PARAM] === undefined) {
```
TheJorgeBorras: I think here it's better to use `if (!params[AUTH_ID_PARAM]) {...}`

it is safer as it checks for additional cases:
- if empty string ''
- if NaN 
- if 0 or false (which aren't possible here)

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
```diff
@@ -0,0 +1,123 @@
...
+	_isOptionPresent(optionId, aliasId, args) {
+		return args[optionId] || args[aliasId];
+	}
+
+	_getOptions() {
+		return this._commandMetadata.options;
+	}
+
+	_isDefaultMode(params) {
+		return (params[SELECT_PARAM] === null || params[SELECT_PARAM] === undefined);
```
TheJorgeBorras: Same deal as last comment, I think here it's better to use if (!params[SELECT_PARAM]) {...}
as it is safer.

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
```diff
@@ -0,0 +1,123 @@
...
+		//Other parameters which are not select and auth_id are not allowed
+		for (const optionId in this._getOptions()) {
+			const aliasId = this._getOptions()[optionId].alias;
+			if (this._getOptions().hasOwnProperty(optionId) &&
+				optionId !== SELECT_PARAM && optionId !== AUTH_ID_PARAM &&
+				this._isOptionPresent(optionId, aliasId, params)) {
+				validationErrors.push(NodeTranslationService.getMessage(IS_NOT_ALLOWED_SELECT_MODE, this._getOptions()[optionId].name))
+			}
+		}
+		return validationErrors;
```
TheJorgeBorras: Something I want to raise here is that we don't usually return multiple error messages, Side by side.
We usually return 1 and when that gets solved we return the next one

For instance, validation of AuthId, "aa %%% $$$" will ONLY that it shouldn't contain spaces and when that gets solved do we solve that AuthId shouldn't contain special characters.

I'm not against what you're doing here, but It's something new. 
I want to make sure we are all aligned on this (returning multiple errors all at once) @davidecorreu

packages/node-cli/src/services/TranslationKeys.js
```diff
@@ -309,6 +311,7 @@ module.exports = {
 	COMMAND_SETUPACCOUNTCI: {
 		OUTPUT: {
 			NEW_OAUTH: 'COMMAND_SETUPACCOUNTCI_OUTPUT_NEW_OAUTH',
+			SELECT_DEFAULT_ACCOUNT: 'COMMAND_SETUPEACCOUNTCI_OUTPUT_SELECT_DEFAULT_ACCOUNT'
```
TheJorgeBorras: Nitpicking here, but I would prefer
...`EXISTING_ACCOUNT`, `EXISTING_AUTHID`, `EXISTING_OAUTH`, `SELECTED_EXISTING_ACCOUNT` or `SELECTED_EXISTING_OAUTH`...

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
carolnavarropereira: The first thing we need to check when running a CI authentication is we are in CI mode.

About the validation into the preExecute it is a functional thing, I mean, first we need to make sure the params are okay into the prExec, and later execute, but it is just my personal opinion, I don't main change it, i can even include an intermediate validate step, I would like this better.  something like 	async validate(params) before or after the preExec in BaseAction and override it in the child classes if needed, it can be executed in Command.run after o before the preExec

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
carolnavarropereira: yes, and I also would to move it to the constants class.

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
carolnavarropereira: I do not think it is a good idea, is(Action) assumes you are returning a boolean, this is not the case, we are performing an action only the validate, so we use the verb.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
```diff
@@ -43,6 +49,15 @@ module.exports = class AccountSetupCiAction extends BaseAction {
...
-		return await authenticateCi(params, this._sdkPath, this._executionPath, this._executionEnvironmentContext);
+		if (this._isDefaultMode(params)) {
+			return await authenticateCi(params, this._sdkPath, this._executionPath, this._executionEnvironmentContext);
+		} else {
+			return await selectAuthenticationCI(params[COMMAND.OPTIONS.AUTHID], this._sdkPath, this._projectFolder);
+		}
 	}
-};
+
+	_isDefaultMode(params) {
```
davidecorreu: I would rename it to make it more clear. Because which is the default mode?

What about `_isSetupMode`

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
davidecorreu: I'm fine showing multiple errors here.

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
```diff
@@ -0,0 +1,123 @@
...
+			validationErrors.push(NodeTranslationService.getMessage(IS_MANDATORY_SELECT_MODE, this._getOptions()[AUTH_ID_PARAM].name));
+		}
+
+		//Other parameters which are not select and auth_id are not allowed
+		for (const optionId in this._getOptions()) {
+			const aliasId = this._getOptions()[optionId].alias;
+			if (this._getOptions().hasOwnProperty(optionId) &&
+				optionId !== SELECT_PARAM && optionId !== AUTH_ID_PARAM &&
+				this._isOptionPresent(optionId, aliasId, params)) {
+				validationErrors.push(NodeTranslationService.getMessage(IS_NOT_ALLOWED_SELECT_MODE, this._getOptions()[optionId].name))
```
davidecorreu: I think this code isn't very clear.
I would tweak it to make it more readable.
Another thing to notice is that `account:setup:ci` options doesn't have any alias.

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
carolnavarropereira: These lines check that parameters which are not select and authId are not allowed, for instance, if we include domain and account it will be two error messages.
account:setup:ci  --select --authid  m1m --account ACCOUNT --privatekeypath pathFolder

packages/node-cli/src/services/TranslationKeys.js
carolnavarropereira: I don't understand, anyway, I tried to follow the existing conventionss, there is an already existing NEW_OAUTH: 'COMMAND_SETUPACCOUNTCI_OUTPUT_NEW_OAUTH' so I tried with something similar.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
TheJorgeBorras: SuiteCloud CLI, prExecute() calls:

`validator.validateAuthID(params[COMMAND.OPTIONS.AUTHID]);`
`validator.validateActionParametersByMode(params);`

then execute():
`validateMachineToMachineAuthIsAllowed();`
into `selectAuthenticationCI(params[COMMAND.OPTIONS.AUTHID], this._sdkPath, this._projectFolder);`
into getting AuthIds and ensuring it exists.


Aside for the "check that AuthId exists", all AuthId validations happen before the Execution Context One! 
It is not the first thing we check.

carolnavarropereira: The point is we will have to move the validation to execute.  It is not difficult but I would rather to create a new method in BaseAction which is executed in the middle of these two which will allow us to keep the code separated. For the moment we may move the full validation code to the execute

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
TheJorgeBorras: I like `validateAuthIDFormat(...)` because it's more descriptive.

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
TheJorgeBorras: Resolving this thread as we are all aligned on this

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
davidecorreu: My main point was to add `AuthIDFormat` to the validation name to make clear what is the validation about.
I gave a couple of suggestions one with validate verb, can you change the method name to `validateAuthIDFormat` or something that states what is the validation about?

packages/node-cli/src/metadata/SdkCommandsMetadata.json
TheJorgeBorras: Have we reviewed this and the new Messages with Montsant?

If these are still not (looking at `authorization id`), we should set up a meeting

packages/node-cli/src/ApplicationConstants.js
```diff
@@ -68,5 +68,23 @@ module.exports = {
...
+				PRIVATEKEYPATH: 'privatekeypath',
+				DOMAIN: 'domain',
+				SELECT: 'select',
+			},
+			SDK_COMMAND: 'authenticateci',
+			OPTIONAL_IN_SETUP_MODE: ['domain'],
+		},
+		ACTION_RESULT: {
+			MODES: { REUSE: 'REUSE' },
+		},
```
davidecorreu: I would prefer to not add this constants here.
The values for ApplicationConstants are mostly for constants that are at "global" level. If we want to introduce some constants that are command specific, I would prefer at least to nest them into a `COMMANDS`  section.

My preference (in order) would be to go for one of the following approaches:

- expose static constants from `packages/node-cli/src/commands/account/setupci/AccountSetupCiCommand.js` or  `packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js` 
- create `packages/node-cli/src/commands/CommandConstants.js` with all commands constants
- create `COMMANDS` nested property into `packages/node-cli/src/ApplicationConstants.js`
- create `packages/node-cli/src/commands/account/setupci/AccountSetupCiConstants.js`

packages/node-cli/src/services/actionresult/AuthenticateActionResult.js
```diff
@@ -20,7 +20,9 @@ class AuthenticateActionResult extends ActionResult {
 		if (parameters.status === STATUS.SUCCESS) {
 			assert(parameters.mode, 'mode is required when ActionResult is a success.');
 			assert(parameters.authId, 'authId is required when ActionResult is a success.');
-			assert(parameters.accountInfo, 'accountInfo is required when ActionResult is a success.');
+			if (parameters.mode !== REUSE) {
+				assert(parameters.accountInfo, 'accountInfo is required when ActionResult is a success.');
+			}
```
davidecorreu: We do have `accountInfo` when selecting an existing authid, let's add this info to `AuthenticateActionResult` from the `AuthenticationUtils.selectAuthenticationCI`method. Then we could remove this if statement.

packages/node-cli/src/utils/AuthenticationUtils.js
```diff
@@ -6,8 +6,12 @@
...
 const FileUtils = require('./FileUtils');
 const NodeTranslationService = require('../services/NodeTranslationService');
-const { ERRORS, UTILS } = require('../services/TranslationKeys');
-const { FILES } = require('../ApplicationConstants');
+const {
+	ERRORS,
+	UTILS,
+	COMMAND_SETUPACCOUNTCI: { ERRORS: { NOT_EXISTING_AUTH_ID } },
+} = require('../services/TranslationKeys');
+const { FILES, COMMAND_AUTHENTICATE_CI_REUSE } = require('../ApplicationConstants');
```
davidecorreu: `COMMAND_AUTHENTICATE_CI_REUSE` it's unused. Please remove import and check we are not leaving behind some unused constant.

packages/node-cli/src/utils/AuthenticationUtils.js
```diff
@@ -175,6 +179,25 @@ async function authenticateCi(params, sdkPath, projectFolder, executionEnvironme
...
+		throw authIDActionResult.errorMessages;
+	}
+	const authIDs = Object.keys(authIDActionResult.data);
+
+	if (authIDs.includes(authId)) {
+		setDefaultAuthentication(projectFolder, authId);
+		return AuthenticateActionResult.Builder.success()
+			.withMode(COMMANDS.AUTHENTICATE.MODES.REUSE)
+			.withAuthId(authId)
+			.build();
```
davidecorreu: Connected to `packages/node-cli/src/services/actionresult/AuthenticateActionResult.js` comment, I think we should add the accountInfo to the `AuthenticateActionResult` as I think it's available on `authIDs`.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
```diff
@@ -5,44 +5,56 @@
...
 	}
 
 	preExecute(params) {
 		this._projectInfoService.checkWorkingDirectoryContainsValidProject(this._commandMetadata.name);
-		
-		if (params[COMMAND.OPTIONS.ACCOUNT]) {
-			params[COMMAND.OPTIONS.ACCOUNT] = params[COMMAND.OPTIONS.ACCOUNT].toUpperCase();
+
+		if (params[OPTIONS.ACCOUNT]) {
+			params[OPTIONS.ACCOUNT] = params[OPTIONS.ACCOUNT].toUpperCase();
```
davidecorreu, line 27: I didn't realize about this `.toUpperCase()` change in my first CR pass.
Is this needed? Couldn't it be causing some conflicts in some cases?

We might need to check this  with different account names right? Maybe you have already tested it. 
I'm just leaving the comment so we can check it together.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
carolnavarropereira: Now I have moved to execute
validateMachineToMachineAuthIsAllowed();
this._validator.validateActionParametersByMode(params);
this._validator.validateAuthIDFormat(this._getAuthId(params));

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
TheJorgeBorras: fixed!

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
TheJorgeBorras: fixed!

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
TheJorgeBorras: Going to leave this here aswell as in testing ticket, 
but this is behavior that is also present on the SDFCLI (something I didn't know about).

If SDFCLI `authenticate` doesn't care about UPPER/LOWER_CASE then suitecloud's `account:setup:ci` shouldn't either.

image

packages/node-cli/src/services/TranslationKeys.js
TheJorgeBorras: This is no longer a relevant thread.
Resolving it

packages/node-cli/src/utils/AuthenticationUtils.js
```diff
@@ -175,6 +179,26 @@ async function authenticateCi(params, sdkPath, projectFolder, executionEnvironme
...
+	const authIDActionResult = await getAuthIds(sdkPath);
+	if (!authIDActionResult.isSuccess()) {
+		throw authIDActionResult.errorMessages;
+	}
+	const authIDs = Object.keys(authIDActionResult.data);
+
+	if (authIDs.includes(authId)) {
+		setDefaultAuthentication(projectFolder, authId);
+		return AuthenticateActionResult.Builder.success()
+			.withMode(COMMANDS.AUTHENTICATE.MODES.REUSE)
```
davidecorreu, line 193: I think that ideally we should use something like `.withMode(COMMANDS.AUTHENTICATE_CI.MODES.SELECT)` or the constants defined on `AccountSetupCiConstants.js`

packages/node-cli/src/commands/account/setupci/AccountSetupCiConstants.js
```diff
@@ -0,0 +1,24 @@
...
+				CERTIFCATEID: 'certificateid',
+				PRIVATEKEYPATH: 'privatekeypath',
+				DOMAIN: 'domain',
+				SELECT: 'select'
+			},
+			SDK_COMMAND: 'authenticateci',
+			MANDATORY_PARAMS_FOR_SETUP_MODE: ['account', 'authid', 'certificateid', 'privatekeypath']
+		},
+		ACTION_RESULT: {
+			MODES: { SELECT: 'REUSE' },
```
davidecorreu: The value for SELECT doesn't seem to be right, I think it should be `'SELECT'`. Please also check comment on `AuthenticationUtils.js` line 193 its connected to this constant too.

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
```diff
@@ -0,0 +1,94 @@
...
+		if (validationErrors.length > 0) {
+			throwValidationException(validationErrors, this._runInInteractiveMode, this._commandMetadata);
+		}
+	}
+
+	/**
+	 * All the parameters are set as non-mandatory in the configuration because the setup mode [account:setup:ci] mode
+	 * is shared with the [account:setup:ci] select mode where these parameters are not used.
+	 * The validation of the mandatory setup mode parameters is done here, instead
+	 * of being handled automatically by the library
```
davidecorreu: Can we review jsdocs in this file. Validation methods have been evolving somehow we left behind the comments.
I would prefer to have:

- general jsdocs comments explanation on `validateActionParametersByMode`
- leave short explanation here (`_validateActionParametersSetupMode`) and in (`_validateActionParametersSelectMode`)

In any case I'm just saying because sentences about "modes" sounds strange, maybe we can just work on it.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
carolnavarropereira: Closed by Carol

packages/node-cli/src/commands/account/setupci/AccountSetupCiConstants.js
carolnavarropereira: It comes from AuthenticationUtils.js COMMANDS.MODES.REUSE
Now it is using directly the value

packages/node-cli/src/utils/AuthenticationUtils.js
carolnavarropereira: Outdated comment

packages/node-cli/src/utils/AuthenticationUtils.js
carolnavarropereira: Now is using the constants in AuthenticationUtils

packages/node-cli/src/commands/account/setupci/AccountSetupCiValidation.js
carolnavarropereira: comments fixed

TheJorgeBorras (approved):

packages/node-cli/src/metadata/SdkCommandsMetadataPatch.json
```diff
@@ -33,6 +33,28 @@
...
+			},
+			"authid": {
+				"mandatory": false
+			},
+			"certificateid": {
+				"mandatory": false
+			},
+			"privatekeypath": {
+				"mandatory": false
+			},
```
alisyed-19, line 48: I have doubt about this, but will check it during testing

alisyed-19 (approved):

PR #945, opened 2025-09-09
pdpdevtool 6199 devassist proxy service implementation

carolnavarropereira:


packages/node-cli/src/services/DevAssistProxyService.js
antoniofruci, line 1: I believe the code would be more readable if we put the public functions on top.
Also we should remove some console.log lines

packages/node-cli/src/services/DevAssistProxyService.js
```diff
@@ -0,0 +1,278 @@
...
+	 * @param authId
+	 * @param proxyPort
+	 * @returns {Promise<void>}
+	 */
+	async start(authId, proxyPort) {
+
+		//Parameters validation
+		if (!authId) {
+			throw NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.MISSING_AUTH_ID);
+		} else {
```
antoniofruci: is this else needed?

packages/node-cli/src/services/DevAssistProxyService.js
```diff
@@ -0,0 +1,278 @@
...
+				});
+
+				proxyReq.write(body);
+				proxyReq.end();
+			});
+		});
+
+		this._localProxy.listen(proxyPort, LOCAL_HOSTNAME, () => {
+			const localURL = `http://${LOCAL_HOSTNAME}:${proxyPort}`;
+			console.log(`SuiteCloud Proxy server listening on ${localURL}`);
```
antoniofruci: We should remove these console.log lines. Some of them are repeated

packages/node-cli/src/services/DevAssistProxyService.js
```diff
@@ -0,0 +1,278 @@
...
+
+	async _createProxyReq(options, body, res, attempts) {
+		return https.request(options, async (proxyRes) => {
+			console.log(`Proxy response ${attempts}: ${proxyRes.statusCode}`);
+			if (proxyRes.statusCode === UNAUTHORIZED_RESPONSE && attempts <= MAX_RETRY_ATTEMPTS) {
+				proxyRes.resume();
+				let authSuccess = await this._forceRefreshAuth();
+				if (authSuccess) {
+					this._updateOptionsAccessToken(options);
+					let clientRequest = await this._createProxyReq(options, body, res, attempts + 1);
```
antoniofruci: Can we call it proxyReq as above? We should also listen to the error event as above.

packages/node-cli/src/services/DevAssistProxyService.js
```diff
@@ -0,0 +1,278 @@
...
+			const bodyChunks = [];
+			req.on('data', function(chunk) {
+				bodyChunks.push(chunk);
+			});
+
+			req.on('end', async function() {
+				const body = Buffer.concat(bodyChunks);
+				const proxyReq = await self._createProxyReq(options, body, res, 0);
+
+				proxyReq.on('error', function(err) {
```
antoniofruci: The callback function can be extracted and re-used in the recursive call below

antoniofruci (changes requested):

packages/node-cli/src/services/DevAssistProxyService.js
```diff
@@ -0,0 +1,278 @@
...
+const MAX_RETRY_ATTEMPTS = 1;
+const LOCAL_HOSTNAME = '127.0.0.1';
+
+/** Target server port */
+const TARGET_SERVER_PORT = 443;
+
+/** Http codes */
+const UNAUTHORIZED_RESPONSE = 401;
+const FORBIDDEN_RESPONSE = 403;
+const INTERNAL_SERVER_ERROR_RESPONSE = 500;
```
davidecorreu: I would group them and use HTTP_CODE instead of response.
Let's go for something like
```
const HTTP_CODE = {
   UNAUTHORIZED: 401,
   FORBIDDED: 403,
   INTERNAL_SERVER_ERROR: 500
}
```

packages/node-cli/src/services/DevAssistProxyService.js
vic-nik: I assume the first console.log line is needed. Should we extract the string template from it to messages.json?

packages/node-cli/src/services/DevAssistProxyService.js
carolnavarropereira: Agree, I will do it.

packages/node-cli/src/services/DevAssistProxyService.js
carolnavarropereira: No, just following old C readability guidelines (Use only one exit point) I can remove it.

packages/node-cli/src/services/DevAssistProxyService.js
carolnavarropereira: I used HTTP_RESPONSE_CODE  (Because it is the response codes)

packages/node-cli/src/services/DevAssistProxyService.js
```diff
@@ -0,0 +1,278 @@
...
+		if (isNaN(proxyPort)) {
+			throw NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.PORT_MUST_BE_NUMBER);
+		}
+
+		//Retrieve from authId accessToken and target host
+		let { accessToken, hostName } = await this._retrieveCredentials();
+		this._targetHost = hostName;
+		this._accessToken = accessToken;
+
+		this._localProxy = http.createServer();
```
davidecorreu: We should make sure to stop any existing instance before creating a new server.

packages/node-cli/src/services/DevAssistProxyService.js
carolnavarropereira: now start, stop and reloadAccessToken are on the bottom

carolnavarropereira: I have also removed most of the console.log, just left a few.

packages/node-cli/src/services/DevAssistProxyService.js
carolnavarropereira: I left only these two just to show that the server is running

carolnavarropereira: These logs are for debugging purposes only currently, we do not know which are going to be left, but this case we may talk about it.

packages/node-cli/src/services/DevAssistProxyService.js
carolnavarropereira: Good point, I will do it.

packages/node-cli/src/services/DevAssistProxyService.js
```diff
@@ -71,36 +77,56 @@ class DevAssistProxyService extends EventEmitter {
...
+			const emitObject = { message: emitObjectMessage, authId: authId };
+			return Object.freeze({ opSuccessful, emitEvent, emitObject, errorMsg, httpStatusCode });
+		}
+	}
+
 	/**
 	 * This method refreshes authorization.
 	 * If successful returns true and updates this._accessToken
-	 * If not successful returns false an emits an event
+	 * If not successful returns false an emits an event.
```
alisyed-19: typo: an emits -> and emits

packages/node-cli/src/services/DevAssistProxyService.js
carolnavarropereira: Done

antoniofruci (approved):

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -0,0 +1,294 @@
...
+			hostName: authIDActionResult.data[this._authId].hostInfo.hostName,
+		};
+	}
+
+	/**
+	 * Record which returns the results of _updateAccessToken
+	 * @type {*}
+	 */
+	_buildResponseUpdateAccessToken(opSuccessful, emitEvent, emitObjectMessage, errorMsg, httpStatusCode, authId) {
+		//boolean (true|false) whether the token has been refresh or has been any problem
```
antoniofruci: wheter the token has been refreshed or there has been any problem

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -0,0 +1,294 @@
...
+		proxy.on('error', (err) => {
+			console.error('Proxy request error:', err);
+			res.writeHead(HTTP_RESPONSE_CODE.INTERNAL_SERVER_ERROR);
+			res.end('SuiteCloud Proxy error: ' + err.message);
+		});
+		return proxy;
+	}
+}
+
+module.exports = { DevAssistProxyService: SuiteCloudAuthProxyService, EVENTS };
```
antoniofruci: I believe the export should be named as the class. We might be importing this class outside the DevAssist context

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: whether

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -0,0 +1,294 @@
...
+		if (!authId) {
+			throw NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.MISSING_AUTH_ID);
+		}
+		this._authId = authId;
+
+		if (!proxyPort) {
+			throw NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.MISSING_PORT);
+		}
+
+		if (isNaN(proxyPort)) {
```
TheJorgeBorras: I'm not sure if we want to enforce stronger port check,
but **isNan(...) will return true for all Real Numbers, not just Naturals!**

This means negative numbers and decimals wont get properly discarded


[https://www.w3schools.com/jsref/jsref_isnan.asp](https://www.w3schools.com/jsref/jsref_isnan.asp)

TheJorgeBorras: Feel free to discard/resolve this thread without fixing it.

We will probably check this on the command trigger class

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -0,0 +1,294 @@
...
+	 * @returns {Promise<void>}
+	 */
+	async stop() {
+		if (this._localProxy) {
+			this._localProxy.close(() => console.log('SuiteCloud Proxy server stopped.'));
+			this._localProxy = null;
+		} else {
+			console.log('No server instance to stop.');
+		}
+	}
```
davidecorreu: I have 2 options here, and I'm not sure which is better:

- remove async from method signature as with current implementation seems to be unnecessary
- keep async and resolve promise when `.close` callback is called or resolve right away when there is no `_localProxy`.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: Node:http listen already shows the error message. We have decided to let the exception to be thrown and be shown later into the output view.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: I have removed the async

PR #946, opened 2025-09-26
feature/pdpdevtool-6201: Dev Asisst service in vscode

davidecorreu:


packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -0,0 +1,253 @@
...
+    // adding listener to trigger manual reauthentication from vscode
+    devAssistProxyService.on(PROXY_SERVICE_EVENTS.REAUTHORIZE, async (emitParams: { authId: string, message: string }) => {
+        // trigger refresh on emited authID 
+        const refreshIsSuccessful = await refreshAuthorizationWithNotifications(emitParams.authId);
+        if (refreshIsSuccessful) {
+            // edgy case: user might have changed configured authid while waiting for the manual refresh to complete
+            updateDevAssistConfigStatus();
+            try {
+                // TODO: we could be using something like devAsssitProxy.reloadAccessToken() to avoid extra forceRefresh in next cline request
+                stopDevAssistService(devAssistStatusBar);
```
TheJorgeBorras, line 115: Shouldn't this be called with the `await` keyword?

davidecorreu: Not really as SuiteCloudAuthProxy.stop() is not an asynchronous method. I'll do some corrections to the code to properly reflect this fact.

Corrections list:
- remove Promise from  `JavascriptNodeCli.d.ts.SuiteCloudAuthProxyServiceInterface.stop()` return type
- remove async keyword from `stopDevAssistService` method

davidecorreu: Fixed in next commit.

TheJorgeBorras: I saw the above call included an `await` so I simply assumed this was the same.

Either way, this works for me.

packages/vscode-extension/src/service/MessageService.ts
```diff
@@ -110,9 +114,61 @@ export default class MessageService {
...
+			.showErrorMessage(
+				includeProjectName ? this.addProjectNameToMessage(message) : message,
+				this.translationService.getMessage(DEVASSIST_SERVICE.IS_STOPPED.NOTIFICATION_BUTTON)
+			)
+			.then(this.showOutputIfClicked);
+	}
+
+	showDevAssistStartUpMessage(infoMessage?: string) {
+		if (!this.vscodeCommandName) {
+			throw COMMAND_NOT_DEFINED;
```
TheJorgeBorras: I keep seeing this `if (!this.vscodeCommandName) {...}` but when how could this actually happen?

Is it just for general safety or is there a specific edge case for this?

packages/vscode-extension/src/service/MessageService.ts
```diff
@@ -110,9 +114,61 @@ export default class MessageService {
...
+		if (message === this.translationService.getMessage(BUTTONS.SEE_DETAILS)) {
+			output.show();
+		}
+
+		if (message === this.translationService.getMessage(DEVASSIST_SERVICE.IS_STOPPED.NOTIFICATION_BUTTON)) {
+			// show suitecloud output
 			output.show();
+			// open devassist settings
+			this.openDevAssistSettings();
 		}
```
TheJorgeBorras: I think this method could be shortened a bit. I'm always a huge fan of simplifying and reducing the length of long files.

++ I dont get why it went from a simple null check into a BUTTONS.SEE_DETAILS

```
private showOutputIfClicked = (message?: string) => {

		if (message) {
			// show suitecloud output on console
			output.show();

			if (message === this.translationService.getMessage(DEVASSIST_SERVICE.IS_STOPPED.NOTIFICATION_BUTTON)) {
				// open devassist settings
				this.openDevAssistSettings();
			}
		}
}
```

TheJorgeBorras: While we are on the topic, I see this as a method that with high likelihood of being reused by future features...

So instead of a message match i'd suggest to write:
-  nested method
- OR pass a Boolean Parameter to open the settings page aside of showing the Output Console

packages/vscode-extension/src/service/MessageService.ts
```diff
@@ -3,11 +3,15 @@
...
+import { DEVASSIST } from '../ApplicationConstants';
 import { output } from '../suitecloud';
-import { BUTTONS, COMMAND } from './TranslationKeys';
+import { BUTTONS, COMMAND, DEVASSIST_SERVICE } from './TranslationKeys';
 import { VSTranslationService } from './VSTranslationService';
 
+
+
+
 const DEFAULT_TIMEOUT = 5000;
```
TheJorgeBorras, line 11: Why so many empty lines?

packages/vscode-extension/src/service/MessageService.ts
davidecorreu: This is not coming form this story, but from previous implementation. I was jus trying to follow same approach as in other methods from `MessageService.ts`. The reason behind it's probably related to  have a way to produce a generic error message for all the commands in case a specific one is not defined.

I was focusing on finishing `DevAssistConfiguration.ts` and left this part of code change in a second place.
I completely agree that the MessageService class can be improved and I'm sure I can do it with some extra time.

TheJorgeBorras: I had a feeling this was the case.

To me it looks okay (for the time being). I'll close the discussion here.

packages/vscode-extension/package.json
```diff
@@ -184,6 +190,35 @@
...
+					"order":1,
+					"type": "number",
+					"default": 8181,
+					"description": "Pick the local port for SuiteCloud Dev Assist service."
+				},
+				"suitecloud.devAssist.enable": {
+					"order": 2,
+					"type": "boolean",
+					"default": true,
+					"markdownDescription": "By enabling the SuiteCloud Dev Assist Service, you're accepting the terms and conditions listed [here](devassist.license.link)\n\nTurn this on to make SuiteCloud Dev Assist service available in your Cline extension. You need to enter a valid auth ID and a local port. When the service is running and Cline is installed, update the following Cline settings:\n* `API Provider`: Set to `OpenAI Compatible`\n* `Base URL`: Set to `http://127.0.0.1:[local port]/api/internal/devassist`\n\nWant more info? See [Setting up SuiteCloud Dev Assist for Cline in the SuiteCloud Extension for Visual Studio Code](https://nlcorp.app.netsuite.com/app/help/helpcenter.nl?fid=article_0916095619.html)."
```
carolnavarropereira: Has this text been validated?

packages/vscode-extension/package.json
```diff
@@ -184,6 +190,35 @@
...
+				},
+				"suitecloud.devAssist.showNotification": {
+					"order":3,
+					"type": "boolean",
+					"default": true,
+					"description": "Want a heads-up each time you start up and the SuiteCloud Dev Assist service isn't running? Turn this on to get a notification."
```
carolnavarropereira: Has this text been validated?

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -0,0 +1,253 @@
...
+    const authID = devAssistConfigSection.get<string>(DEVASSIST.CONFIG_KEYS.auhtID, DEVASSIST.DEFAULT_VALUES.authID);
+    const localPort = devAssistConfigSection.get<number>(DEVASSIST.CONFIG_KEYS.localPort, DEVASSIST.DEFAULT_VALUES.localPort);
+    const startupNotificationEnabled = devAssistConfigSection.get<boolean>(DEVASSIST.CONFIG_KEYS.startupNotificationEnabled, DEVASSIST.DEFAULT_VALUES.startupNotificationEnabled);
+
+    return { proxyEnabled, authID, localPort, startupNotificationEnabled }
+}
+
+const devAssistConfigStatusHasEffectivelyChanged = (): boolean => {
+    // we don't know exaclty were configuration change comes from when devAssistConfigurationChangeHandler is called
+    // it could be that configuration has changed from globalValue (user) and there is workspaceValue that has left intact
```
carolnavarropereira: Maybe the comment needs a bit of grammar review (where, exactly...)

packages/vscode-extension/src/startup/StatusBarItemsFunctions.ts
```diff
@@ -34,6 +34,19 @@ export function createAuthIDStatusBar(): vscode.StatusBarItem {
...
 
+/**
+ * Create the status bar item used to display DevAssist service status
+ *
+ * @returns {vscode.StatusBarItem} The authID status bar item
+ */
+export function createDevAssistStatusBar(): vscode.StatusBarItem {
+	const devAssistStatusBar = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, STATUS_BAR_PRIORITY - 2);
+	devAssistStatusBar.tooltip = 'Shows DevAssist service status';
+	devAssistStatusBar.hide();
```
TheJorgeBorras, line 47: This code is perfectly fine.

But just for maintainability's sake, personally, I would add a comment or a row in the JSDoc to explain that `.hide()` is to hide the statusBar by default and then `.show()` might get called later from `devAssistConfigurationChangeHandler()` on VSCode startup.

To me it was something I had to figure out. But if you believe it's obvious then we can leave it as is

packages/vscode-extension/package.json
davidecorreu: Text has been provided by our tech-writer but it could be modified if required.

packages/vscode-extension/src/service/MessageService.ts
davidecorreu: I'll remove them in next commit.

packages/vscode-extension/src/startup/StatusBarItemsFunctions.ts
davidecorreu: I guess you are right, it's not obvious. I'll add a comment about it.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: Comment reviewed in next commit.

packages/vscode-extension/package.json
davidecorreu: Text has been provided by our tech-writer but it could be modified if required.

carolnavarropereira (approved):

packages/vscode-extension/package.json
```diff
@@ -192,7 +227,7 @@
 		"watch": "tsc -watch -p ./"
 	},
 	"dependencies": {
-		"@oracle/suitecloud-cli": "^3.1.0"
+		"@oracle/suitecloud-cli": "^3.1.0-snapshot-241"
```
TheJorgeBorras: Small comment I want to make here...
Is this correct?

We shouldn't be pointing at a specific version

davidecorreu: This a mistake, I'll set it back to the expected value in the next commit.

packages/vscode-extension/src/service/MessageService.ts
TheJorgeBorras: I agree with the new changes.
Closing thread.

PR #950, opened 2025-10-03
PDPDEVTOOL-6254-fix-vslogger-and-types

davidecorreu:
use correct vslogger type from ImportObjectService
define SuiteCloudAuthProxy as typescript class
replace term Interface for Instance on JavascriptNodeCli.d.ts

### DIFF
--- a/packages/vscode-extension/src/types/JavascriptNodeCli.d.ts
+++ b/packages/vscode-extension/src/types/JavascriptNodeCli.d.ts
@@ -26,7 +26,7 @@ export interface ConsoleLoggerInstance {
 	error(message: string): void;
 }
 export interface ConsoleLoggerConstructor {
-	new(): ConsoleLoggerInterface
+	new(): ConsoleLoggerInstance
 }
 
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/types/JavascriptNodeCli.d.ts
```diff
@@ -29,19 +29,19 @@ export interface ConsoleLoggerConstructor {
 	new(): ConsoleLoggerInterface
```
TheJorgeBorras: Hey, did you miss this one on the constructor?
`new(): ConsoleLoggerInstance`

davidecorreu: This how you define the constructor in a typescript interface/type.
Instead of `constructor` the `new` keyword/operator is used to define it. Not very intuitive but it works.

davidecorreu: Sorry I didn't get your point. Fixing it in next commit to use `new(): ConsoleLoggerInstance` .

PR #952, opened 2025-10-06
Feature/pdpdevtool 6203 merge dev with release 25.2

carolnavarropereira:
-  Merge dev to release 25.2
   -  Removal of schemas
-  Review versions to release
-  Update base package-lock.json

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,19 +13,6 @@
 				"packages/unit-testing"
 			]
 		},
-		"node_modules/@ampproject/remapping": {
-			"version": "2.3.0",
-			"resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
-			"integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
-			"license": "Apache-2.0",
-			"dependencies": {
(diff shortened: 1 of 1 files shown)


packages/node-cli/src/commands/account/setupci/AccountSetupCiOutputHandler.js
```diff
@@ -5,24 +5,36 @@
...
-			COMMAND_SETUPACCOUNTCI.OUTPUT.NEW_OAUTH,
-			actionResult.accountInfo.companyName,
-			actionResult.accountInfo.roleName,
-			actionResult.authId
-		);
 
-		this._log.result(resultMessage);
-		this._log.result(NodeTranslationService.getMessage(UTILS.AUTHENTICATION.SUCCESS_SETUP));
+		if (actionResult.mode === REUSE) {
+			//SELECT mode
```
davidecorreu, line 23: This doesn't seem to be right. We should keep it in mind when we do testing.

davidecorreu (approved):

TheJorgeBorras:
On `packages/node-cli/README.md`, We should probably update this.
Add another row for 3.0.2 (since this table usually contains older versions)

image

packages/node-cli/src/commands/account/setupci/AccountSetupCiOutputHandler.js
TheJorgeBorras: what doesn't seem right?
Please elaborate

TheJorgeBorras (approved):

PR #959, opened 2025-10-14
bugfix/PDPDEVTOOL 6260 DevAssist not handling user ports proxy service

carolnavarropereira:


packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -125,9 +126,15 @@ const initializeDevAssistService = (devAssistStatusBar: vscode.StatusBarItem) =>
...
         vsLogger.error(translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message));
-        // add line sepparator
+        // add line separator
         vsLogger.error('');
     });
+
+    devAssistProxyService.on(PROXY_SERVICE_EVENTS.ALREADY_USED_PORT, (emitParams: { authId: string, message: string }) => {
+        vsLogger.error(translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message));
+        vsLogger.error('');
+        showStartDevAssistProblemNotification('alreadyUsedPort', emitParams.message, devAssistStatusBar)
```
davidecorreu, line 137: Adding the listener for `ALREADY_USED_PORT` event it's a good idea. 
We might want to review how to log the messages as `showStartDevAssistProblemNotification` already logs and error message in the vscode-output (`vsLogger`) and also shows a notification.

I would like to think deeper if we could somehow start the devassistProxyService automatically if the port is released and available again.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -100,6 +101,15 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
 		});
+
+		this._localProxy.on('error', (error) => {
+			if (error.code === 'EADDRINUSE') {
+				const errorMsg = `Port ${proxyPort} is already being used. Choose a different port and try again.`;
+				console.error(errorMsg);
+				const emitObject = { message: errorMsg, authId: this._authId };
+				this.emit(EVENTS.ALREADY_USED_PORT, emitObject);
+			}
+		});
```
davidecorreu, line 114: I guess we missed to listen on proxy errors in the in initial implementation.
We might also want to emit other possible occurring errors, and not only the `'EADDRINUSE'`.
The error message we'll need to be moved to messages file.

carolnavarropereira: yes, we can emit a message when and error is generated but I will emit two different messages, port in use and generic error since we may deal different with each type on error.

TheJorgeBorras, line 114: It would be nice to log any errors we get from Suitecloud Proxy.

Maybe leave this as it is but log any errors onto the VSCode internal console so that we can check them out later (very useful if we ever need to debug a user instance).

TheJorgeBorras: or better even, just pass any and all errors to VSCode and handle them there?

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
carolnavarropereira: I don't think it will work since the code just checks whether the port is available and returns error if it is not, the only way to do it is just write a process that checks each X fractions of time. I do not see the point to do this, besides it would be an automatic start (not manual) which is even a worse idea. I think it is better to leave it like that.

PR #963, opened 2025-10-20
pdpdevtool 6260 dev assist not handling user ports proxy service dev

carolnavarropereira:


packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -100,6 +102,16 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
 			console.log(`SuiteCloud Proxy server listening on ${localURL}`);
 		});
+
+		this._localProxy.on('error', (error) => {
+			if (error.code === 'EADDRINUSE') {
+				const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.ALREADY_USED_PORT, proxyPort);
+				this._handleListeningPortError(errorMsg, EVENTS.ALREADY_USED_PORT);
+			} else {
+				const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.LISTENING_PORT_ERROR, proxyPort);
+				this._handleListeningPortError(errorMsg, EVENTS.LISTENING_PORT_ERROR);
```
TheJorgeBorras: This part is wrapping any error message we get while the proxy is starting/running.

In the future it might benefit us to be able to get the original message.
Would it be possible to also pass the original message instead of hiding it?
Maybe even writing it into the internal vscode console?

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -125,9 +127,22 @@ const initializeDevAssistService = (devAssistStatusBar: vscode.StatusBarItem) =>
...
+    devAssistProxyService.on(PROXY_SERVICE_EVENTS.ALREADY_USED_PORT, (emitParams: { authId: string, message: string }) => {
+        vsLogger.error(translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message));
+        vsLogger.error('');
+        showStartDevAssistProblemNotification('alreadyUsedPort', emitParams.message, devAssistStatusBar)
+    });
+
+    devAssistProxyService.on(PROXY_SERVICE_EVENTS.LISTENING_PORT_ERROR, (emitParams: { authId: string, message: string }) => {
+        vsLogger.error(translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message));
+        vsLogger.error('');
+        showStartDevAssistProblemNotification('listeningPortError', emitParams.message, devAssistStatusBar)
```
TheJorgeBorras: Probably use the constant PROXY_SERVICE_EVENTS.LISTENING_PORT_ERROR (instead of  "listeningPortError").
Same for the 'alreadyUsedPort'

Looking into it, this errorStage is never really used anywere and should be deleted, right?
React

TheJorgeBorras (approved):

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -100,6 +102,16 @@ class SuiteCloudAuthProxyService extends EventEmitter {
 			const localURL = `http://${LOCAL_HOSTNAME}:${proxyPort}`;
 			console.log(`SuiteCloud Proxy server listening on ${localURL}`);
 		});
+
+		this._localProxy.on('error', (error) => {
+			if (error.code === 'EADDRINUSE') {
+				const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.ALREADY_USED_PORT, proxyPort);
+				this._handleListeningPortError(errorMsg, EVENTS.ALREADY_USED_PORT);
+			} else {
+				const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.LISTENING_PORT_ERROR, proxyPort, error.message ?? '');
```
TheJorgeBorras, line 111: instead of `error.message ?? ''`, maybe we could go for `"\n" + error.message ?? ''`.

This way we introduce a lineskip between the generic message and the actual cause of the error (which should result in better UX).
Feel fee to ignore this comment, I don't feel like it's important enough to block the PR.

PR #968, opened 2025-10-28
PDPDEVTOOL 6260 dev assist not handling user ports proxy service dev

carolnavarropereira:


packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -100,6 +102,16 @@ class SuiteCloudAuthProxyService extends EventEmitter {
 			const localURL = `http://${LOCAL_HOSTNAME}:${proxyPort}`;
 			console.log(`SuiteCloud Proxy server listening on ${localURL}`);
 		});
+
+		this._localProxy.on('error', (error) => {
+			if (error.code === 'EADDRINUSE') {
+				const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.ALREADY_USED_PORT, proxyPort);
+				this._handleListeningPortError(errorMsg, EVENTS.ALREADY_USED_PORT);
+			} else {
+				const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.LISTENING_PORT_ERROR, proxyPort, error.message ?? '');
```
TheJorgeBorras: instead of `error.message ?? ''`, maybe we could go for `"\n" + error.message ?? ''`.

This way we introduce a lineskip between the generic message and the actual cause of the error (which should result in better UX).
Feel fee to ignore this comment, I don't feel like it's important enough to block the PR.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -125,9 +127,22 @@ const initializeDevAssistService = (devAssistStatusBar: vscode.StatusBarItem) =>
...
         // just forwarding info into suitecloud output for now
         vsLogger.error(translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message));
-        // add line sepparator
+        // add line separator
         vsLogger.error('');
     });
+
+    devAssistProxyService.on(PROXY_SERVICE_EVENTS.ALREADY_USED_PORT, (emitParams: { authId: string, message: string }) => {
+        showStartDevAssistProblemNotification(PROXY_SERVICE_EVENTS.ALREADY_USED_PORT, emitParams.message, devAssistStatusBar);
+        vsLogger.error(translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message));
```
davidecorreu: The `showStartDevAssistProblemNotification` function is already going to show in the output the `emitParams.message`.
I think we should not duplicate error info in the next line.
Let's try to concatenate the `DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT` message and forward it to `showStartDevAssistProblemNotification`.

The result would be something like:
```
const errorMessage = translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message);
showStartDevAssistProblemNotification(PROXY_SERVICE_EVENTS.ALREADY_USED_PORT, errorMessage, devAssistStatusBar);
vsLogger.error('');
```

We could apply same logic for lines 141-143.

davidecorreu: I also have some doubts when this kind of errors are going to occur.
Is it only going to happen at initialization time?
Could be happening sometime after proxy has started?

Just leaving here my second thoughts about this.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -100,6 +102,16 @@ class SuiteCloudAuthProxyService extends EventEmitter {
 			const localURL = `http://${LOCAL_HOSTNAME}:${proxyPort}`;
 			console.log(`SuiteCloud Proxy server listening on ${localURL}`);
 		});
+
+		this._localProxy.on('error', (error) => {
+			if (error.code === 'EADDRINUSE') {
+				const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.ALREADY_USED_PORT, proxyPort, error.message ?? '');
+				this._handleListeningPortError(errorMsg, EVENTS.ALREADY_USED_PORT);
+			} else {
+				const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.INTERNAL_PROXY_SERVER_ERROR, proxyPort, error.message ?? '');
```
TheJorgeBorras:
```
DEV_ASSIST_PROXY_SERVICE_INTERNAL_PROXY_SERVER_ERROR = "There was an Internal Client SuiteCloud Server problem while listening on the port {0}.\n{1}"
```

On second thoughts, maybe instead of writing the `\n` onto the message, we should add it to line 111 instead.


If we leave it as it is, we might print unnecessary `\n` if there is no error.message.
if we go for something like (notice the `"\n" + error.message ?? ''` ): 

```
const errorMsg = NodeTranslationService.getMessage(DEV_ASSIST_PROXY_SERVICE.INTERNAL_PROXY_SERVER_ERROR, proxyPort, "\n" + error.message ?? '');
```
Then we only print the lineskip wheneaver there is an error message.
Do the same for `INTERNAL_PROXY_SERVER_ERROR` and `ALREADY_USED_PORT`.

Mb on this one, it was an oversight I made when I suggested this change

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -121,13 +123,25 @@ const initializeDevAssistService = (devAssistStatusBar: vscode.StatusBarItem) =>
...
+		const errorMsgWrap = translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message);
+        showStartDevAssistProblemNotification(PROXY_SERVICE_EVENTS.ALREADY_USED_PORT, errorMsgWrap, devAssistStatusBar);
+		vsLogger.error('');
     });
+
+    devAssistProxyService.on(PROXY_SERVICE_EVENTS.PROXY_ERROR, (emitParams: { authId: string, message: string }) => {
+		const errorMsgWrap = translationService.getMessage(DEVASSIST_SERVICE.SERVER_ERROR.OUTPUT, emitParams.message);
+        showStartDevAssistProblemNotification(PROXY_SERVICE_EVENTS.PROXY_ERROR, errorMsgWrap, devAssistStatusBar);
+        vsLogger.error('');
+    });
```
TheJorgeBorras, line 138: Just leaving this comment here as well. I think we went overboard with the error Wrapping.

Image

Either the first or second line should be enough

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -32,7 +32,9 @@ const devAssistConfigStatus: { current: devAssistConfig, previous: devAssistConf
 
 const PROXY_SERVICE_EVENTS = {
     REAUTHORIZE: 'authRefreshManual',
-    SERVER_ERROR: 'serverError'
+    SERVER_ERROR: 'serverError',
+    ALREADY_USED_PORT: 'alreadyUsedPort',
```
davidecorreu: We can delete `ALREADY_USED_PORT` as is no longer used.

packages/node-cli/src/services/TranslationKeys.js
```diff
@@ -450,6 +450,8 @@ module.exports = {
 		PORT_MUST_BE_NUMBER: 'DEV_ASSIST_PROXY_SERVICE_ERROR_PORT_MUST_BE_NUMBER',
 		NOT_EXISTING_AUTH_ID: 'DEV_ASSIST_PROXY_SERVICE_NOT_EXISTING_AUTH_ID',
 		NEED_TO_REAUTHENTICATE: 'DEV_ASSIST_PROXY_SERVICE_NEED_TO_REAUTHENTICATE',
+		ALREADY_USED_PORT: 'DEV_ASSIST_PROXY_SERVICE_ALREADY_USED_PORT',
+		INTERNAL_PROXY_SERVER_ERROR: 'DEV_ASSIST_PROXY_SERVICE_INTERNAL_PROXY_SERVER_ERROR',
```
davidecorreu: In line 447 we have defined parent `DEV_ASSIST_PROXY_SERVICE` which I really think we should change for `SUITECLOUD_AUTH_PROXY_SERVICE` or similar.
Main point would be to go for a name that doesn't include `DEV_ASSIST`.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
carolnavarropereira: Fixed

packages/node-cli/src/services/TranslationKeys.js
```diff
@@ -444,12 +444,14 @@ module.exports = {
...
 	},
-	DEV_ASSIST_PROXY_SERVICE: {
+	SUITECLOUD_AUTH_PROXY_SERVICE: {
 		MISSING_AUTH_ID: 'DEV_ASSIST_PROXY_SERVICE_ERROR_MISSING_AUTH_ID',
 		MISSING_PORT: 'DEV_ASSIST_PROXY_SERVICE_ERROR_MISSING_PORT',
 		PORT_MUST_BE_NUMBER: 'DEV_ASSIST_PROXY_SERVICE_ERROR_PORT_MUST_BE_NUMBER',
 		NOT_EXISTING_AUTH_ID: 'DEV_ASSIST_PROXY_SERVICE_NOT_EXISTING_AUTH_ID',
 		NEED_TO_REAUTHENTICATE: 'DEV_ASSIST_PROXY_SERVICE_NEED_TO_REAUTHENTICATE',
+		ALREADY_USED_PORT: 'DEV_ASSIST_PROXY_SERVICE_ALREADY_USED_PORT',
+		INTERNAL_PROXY_SERVER_ERROR: 'DEV_ASSIST_PROXY_SERVICE_INTERNAL_PROXY_SERVER_ERROR',
```
davidecorreu: Please remove also `DEV_ASSIST` mention from key values (here and in the messages file).

packages/vscode-extension/messages.json
```diff
@@ -54,6 +54,9 @@
 	"DEPLOY_QUESTIONS_CHOICES_ACCOUNT_SPECIFIC_VALUES_CANCEL_PROCESS": "Cancel the deployment process",
 	"DEPLOY_QUESTIONS_CHOICES_ACCOUNT_SPECIFIC_VALUES_DISPLAY_WARNING": "Display a warning and continue the deployment process",
 
+	"DEVASSIST_SERVICE_OUTPUT_PROXY_ERROR": "There was a problem while running SuiteCloud Developer Assistant service.\n{0}",
+	"DEVASSIST_SERVICE_OUTPUT_SERVER_ERROR": "There was a server error while running SuiteCloud Developer Assistant service.\n{0}",
+	"DEVASSIST_SERVICE_OUTPUT_SERVER_ERROR_ON_REFRESH": "There was a problem refreshing credentials.\n{0}",
```
davidecorreu: Even if we change this in a future story, could we follow similar approach as 2 messages above?

TheJorgeBorras (approved):

davidecorreu (approved):

PR #969, opened 2025-10-28
PDPDEVTOOL 6274: Add Timestamps to the DevAssist Log Messages

TheJorgeBorras:
also did some code quality changes


packages/vscode-extension/src/loggers/VSConsoleLogger.ts
```diff
@@ -34,13 +34,13 @@ export default class VSConsoleLogger extends ConsoleLogger {
 		}
 	}
 
-	hiddeInitialProjectFolerNameDetails() {
+	public hiddeInitialProjectFolerNameDetails() {
 		this.hideProjectFolderName = true;
```
carolnavarropereira, line 38: bad spelling hiddeInitialProjectFolerNameDetails : missing a d
Optionally: To keep it consistent you can make it to return a void in the definition

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -114,44 +115,48 @@ const initializeDevAssistService = (devAssistStatusBar: vscode.StatusBarItem) =>
...
 
     setErrorDevAssistStausBarMessage(devAssistStatusBar);
     // only notify that devassist service has been disabled in the transition from enabled to disabled
     if (!devAssistConfigStatus.current.proxyEnabled && devAssistConfigStatus.previous.proxyEnabled) {
-        vsLogger.info(translationService.getMessage(DEVASSIST_SERVICE.IS_DISABLED.NOTIFICATION))
         vsNotificationService.showCommandInfo(translationService.getMessage(DEVASSIST_SERVICE.IS_DISABLED.OUTPUT));
-        // add extra line to differenciate logs
+
+        vsLogger.printTimestamp();
+        vsLogger.info(translationService.getMessage(DEVASSIST_SERVICE.IS_DISABLED.NOTIFICATION));
```
carolnavarropereira, line 158: I don't like too much the idea of using two lines when they are supposed to be together. My solution is if you do not want to create a new function for all of them info, error.... you can create one only method something like printTimestamp(message, Level) being a level an enum like INFO, ERROR, but it is only an optional suggestion, feel free to resolve this command.

packages/vscode-extension/src/loggers/VSConsoleLogger.ts
TheJorgeBorras: 🤯  didn't even realize when I modified the access level of the method.

Pushed fix

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -176,15 +181,16 @@ const showDevAssistStartUpNotification = () => {
 
 const showStartDevAssistProblemNotification = (errorStage: string, error: string, devAssistStatusBar: vscode.StatusBarItem) => {
     // console.log(`There was a problem when starting DevAssist service. (${errorStage})\n${error}`)
-    setErrorDevAssistStausBarMessage(devAssistStatusBar)
+    setErrorDevAssistStausBarMessage(devAssistStatusBar);
+    vsLogger.printTimestamp();
```
davidecorreu, line 185: I'm not sure that it's a good idea to add the timestamp here, as this method is usually called as an already ongoing action that most probably has already shown the timestamp.

I'll need to check it a little bit further to be sure about it.

TheJorgeBorras: I had the same thought.

My conclusion is that indeed a single action can generate more than one log message... But that's okay.
Simply each message will be prefaced by a Timestamp (which doesn't look that bad imo) and separated by a lineskip.

Nowhere should there be 2 timestamps back to back without a log in between.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: So far we don't have "coloring" for vscode output console and I would not worry about it at this moment.
I do like that the `printTimestamp` logic is part of `VSConsoleLogger` and that we'll allow us to chance to modify the logic there when needed.

TheJorgeBorras: We can have a back-and-forth as how to archieve the desired result. There are multiple ways.

Personally I too like the idea of creating new vsLogger methods to print a Timestamp then (inside the same method) call `vsLogger.info` or `vsLogger.error`. I'd create 2 different methods, but yeah.

We can also extend the current .info and .error to include a new boolean parameter.

We can also add a new global bool to the constructor to always print the Timestamp (set to false by default except inside the DevAssist code). I like this one very much as well

David prefered this approach, however.

packages/vscode-extension/src/loggers/VSConsoleLogger.ts
```diff
@@ -49,22 +49,29 @@ export default class VSConsoleLogger extends ConsoleLogger {
...
 		this.println(message);
 	}
 
-	error(message: string): void {
+	public error(message: string): void {
 		this.println(message);
 	}
 
+	// Log the current Timestamp
+	// used to preface specific Logs outside the execution context of SuiteCloud commands (where no execution details are available)
```
davidecorreu: Let's add the comment as JSDocs so they will be shown as description in editors.

TheJorgeBorras: solved. 
Please check the most recent comment.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: It looks good after reviewing it deeper.

PR #972, opened 2025-11-12
pdpdevtool 6317 dev assist avoid to stop button when proxy server error

carolnavarropereira:


packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -211,6 +211,11 @@ const showStartDevAssistProblemNotification = (errorStage: string, error: string
     vsNotificationService.showCommandErrorWithSpecificButtonsAndActions(errorMessage, buttonsAndActions);
 }
 
+const showDevAssistEmitProblemLog = (emitError: string) => {
+	vsLogger.printTimestamp();
+	vsLogger.error(emitError);
+}
+
```
TheJorgeBorras: So before we called `showDevAssistEmitProblemNotification(...)` to:
1. setStatusBar DevAssist to stopped.
2. print error Log.
3. show VSCode Notification message.

On the ticket it was agreed to **skip step 1** but **keep 2 and 3**. But with this implementation we are just printing the Error Log and NOT showing a Notification.

As long as we remove **step 1** I'm ok. But we should think if this implementation is better for the User! @davidecorreu what's your take?

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -211,6 +211,19 @@ const showStartDevAssistProblemNotification = (errorStage: string, error: string
...
+    vsLogger.error(emitError);
+    const errorMessage = translationService.getMessage(DEVASSIST_SERVICE.IS_STOPPED.NOTIFICATION);
+    const buttonsAndActions: { buttonMessage: string, buttonAction: () => void }[] = [
+        {
+            buttonMessage: translationService.getMessage(DEVASSIST_SERVICE.SEE_DETAILS_BUTTON),
+            buttonAction: () => output.show(),
+        },
+    ];
+    vsNotificationService.showCommandErrorWithSpecificButtonsAndActions(errorMessage, buttonsAndActions);
+};
```
TheJorgeBorras, line 225: Optional: maybe we could go for the `MessageService.showCommandError()` ?

Idk if it makes more sense since the code would be simpler but this is technically not a command output..

davidecorreu (approved):

PR #974, opened 2025-11-17
PDPDEVTOOL-6313 Dev Assists - Allow list of Dev Assist endpoints while proxy is running

carolnavarropereira:


packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -14,7 +14,8 @@ const EVENTS = {
 	SERVER_ERROR: 'serverError',
 	SERVER_ERROR_ON_REFRESH: 'serverErrorOnRefresh',
 	AUTH_REFRESH_MANUAL_EVENT: 'authRefreshManual',
-	PROXY_ERROR: 'proxyError'
+	PROXY_ERROR: 'proxyError',
+	PROXY_SETTINGS_ERROR: 'proxySettingsError'
```
TheJorgeBorras: To me this sounds like a `PROXY_REQUEST_ERROR` error rather than a `PROXY_SETTINGS_ERROR` (the settings problem is with Cline, not the proxy itself).

If the Url-Filter part of the code is reached, then it's a given that the Proxy service is configured properly. 
It's just that destination URL of the Cline request is wrong.

packages/vscode-extension/messages.json
```diff
@@ -55,6 +55,7 @@
 	"DEPLOY_QUESTIONS_CHOICES_ACCOUNT_SPECIFIC_VALUES_DISPLAY_WARNING": "Display a warning and continue the deployment process",
 
 	"DEVASSIST_SERVICE_OUTPUT_PROXY_ERROR": "There was a problem while running SuiteCloud Developer Assistant service.\n{0}",
+	"DEVASSIST_SERVICE_OUTPUT_PROXY_SETTINGS_ERROR": "There was a problem while running SuiteCloud Developer Assistant service.\n{0}",
```
TheJorgeBorras: Duplicated message.

packages/vscode-extension/src/service/TranslationKeys.ts
```diff
@@ -101,6 +101,7 @@ export const DEVASSIST_SERVICE = {
 	EMIT_ERROR: {
 		OUTPUT: {
 			PROXY_ERROR: 'DEVASSIST_SERVICE_OUTPUT_PROXY_ERROR',
+			PROXY_SETTINGS_ERROR: 'DEVASSIST_SERVICE_OUTPUT_PROXY_SETTINGS_ERROR',
```
TheJorgeBorras: Same deal as before, I strongly think that it should be a "PROXY_REQUEST_ERROR" or something similar

packages/vscode-extension/messages.json
carolnavarropereira: It is a different error technically, it is I have to decide which is the proper message error

packages/node-cli/messages.json
```diff
@@ -293,6 +293,7 @@
 	"SUITECLOUD_AUTH_PROXY_SERVICE_NEED_TO_REAUTHENTICATE" : "Manual authentication is needed. Sign in again to continue.",
 	"SUITECLOUD_AUTH_PROXY_SERVICE_ALREADY_USED_PORT" : "Port {0} is already being used. Choose a different port and try again.\n{1}",
 	"SUITECLOUD_AUTH_PROXY_SERVICE_INTERNAL_PROXY_SERVER_ERROR" : "There was an internal problem in the SuiteCloud client server while listening on the port {0}.\n{1}",
+	"SUITECLOUD_AUTH_PROXY_SERVICE_INVALID_CLINE_BASE_URL_ERROR" : "Cline Base URL is invalid. Valid format may be http://127.0.0.1:[port]/api/internal/devassist",
```
davidecorreu: At node-cli layer we should not mention anything about CLINE.
This message should go in the line of `The received request doesn't honor allowed path requirements`.

packages/node-cli/src/ApplicationConstants.js
```diff
@@ -68,5 +68,9 @@ module.exports = {
 		SUITECLOUD_CI: 'SUITECLOUD_CI',
 		SUITECLOUD_CI_PASSKEY: 'SUITECLOUD_CI_PASSKEY',
 		SUITECLOUD_FALLBACK_PASSKEY: 'SUITECLOUD_FALLBACK_PASSKEY'
+	},
+
+	SUITECLOUD_AUTH_PROXY: {
+		ALLOWED_PATH: '/api/internal/devassist/*'
```
davidecorreu: This constant is defined here and never used.
I'm not sure how to define the alllowed paths, but we might want to delete this constant if its not used.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: I agree with Jordi about this.
The problem is not with the proxy configuration but with the received request.

carolnavarropereira: Changed by proxyRequestError

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -14,7 +14,8 @@ const EVENTS = {
 	SERVER_ERROR: 'serverError',
 	SERVER_ERROR_ON_REFRESH: 'serverErrorOnRefresh',
 	AUTH_REFRESH_MANUAL_EVENT: 'authRefreshManual',
-	PROXY_ERROR: 'proxyError'
+	PROXY_ERROR: 'proxyError',
+	NOT_ALLOWED_PATH_ERROR: 'notAllowedPathError'
```
TheJorgeBorras: I would really appreciate it if we went for "PATH_NOT_ALLOWED_ERROR" instead of "NOT_ALLOWED_PATH_ERROR". It rolls off the tongue easier.

++ It follows better the HTTP_ERROR conventions (ex "Method_Not_Allowed", "Range Not Satisfiable", "HTTP Version Not Supported")

packages/vscode-extension/src/service/TranslationKeys.ts
```diff
@@ -101,6 +101,7 @@ export const DEVASSIST_SERVICE = {
 	EMIT_ERROR: {
 		OUTPUT: {
 			PROXY_ERROR: 'DEVASSIST_SERVICE_OUTPUT_PROXY_ERROR',
+			NOT_ALLOWED_PATH_ERROR: 'DEVASSIST_SERVICE_OUTPUT_NOT_ALLOWED_PATH_ERROR',
```
TheJorgeBorras: Same here: "PATH_NOT_ALLOWED_ERROR" : "DEVASSIST_SERVICE_OUTPUT_PATH_NOT_ALLOWED_ERROR"

carolnavarropereira: I disagree with PATH_NOT_ALLOWED_ERROR, see previous comment.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -48,10 +49,11 @@ const HTTP_RESPONSE_CODE = {
 };
 
 class SuiteCloudAuthProxyService extends EventEmitter {
-	constructor(sdkPath, executionEnvironmentContext) {
+	constructor(sdkPath, executionEnvironmentContext, proxyUrl) {
 		super();
 		this._sdkPath = sdkPath;
 		this._executionEnvironmentContext = executionEnvironmentContext;
+		this._proxyUrl = proxyUrl;
```
davidecorreu: What's your reasoning to name this one `proxyUrl`?

Isn't it the allowed_path prefix?

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -158,6 +170,21 @@ class SuiteCloudAuthProxyService extends EventEmitter {
 		}
 	}
 
+	/**
+	 * Matches /api/internal/devassist/ followed by one or more valid URL path characters
+	 * @param path
+	 * @returns {boolean}
+	 * @private
+	 */
+	_matchesDevAssistPath(path) {
```
davidecorreu: Please do not make mentions about devassist in node-cli layer.
This method seems to check if the requestPath follows the a certain pattern, please name it accordingly.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -158,6 +170,21 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+		// This regex matches any string that starts with "/api/internal/devassist/" followed by anything
+		let validRequest = this._proxyUrl.PATH;
+		if (!validRequest.endsWith('/')) {
+			validRequest = validRequest.concat('/');
+		}
```
davidecorreu: I don't thing we should be adding the  `/` character at the end, not here neither when constructing the error message.
Let the code that instantiates this class be responsible to add it.

packages/node-cli/messages.json
```diff
@@ -293,6 +293,7 @@
 	"SUITECLOUD_AUTH_PROXY_SERVICE_NEED_TO_REAUTHENTICATE" : "Manual authentication is needed. Sign in again to continue.",
 	"SUITECLOUD_AUTH_PROXY_SERVICE_ALREADY_USED_PORT" : "Port {0} is already being used. Choose a different port and try again.\n{1}",
 	"SUITECLOUD_AUTH_PROXY_SERVICE_INTERNAL_PROXY_SERVER_ERROR" : "There was an internal problem in the SuiteCloud client server while listening on the port {0}.\n{1}",
+	"SUITECLOUD_AUTH_PROXY_SERVICE_INVALID_BASE_URL_ERROR" : "Invalid request. The request path must start with {0}.",
```
davidecorreu: The message that @Montsant19 has share has double quotes around `{0}`.

packages/node-cli/src/services/TranslationKeys.js
```diff
@@ -452,6 +452,7 @@ module.exports = {
 		NEED_TO_REAUTHENTICATE: 'SUITECLOUD_AUTH_PROXY_SERVICE_NEED_TO_REAUTHENTICATE',
 		ALREADY_USED_PORT: 'SUITECLOUD_AUTH_PROXY_SERVICE_ALREADY_USED_PORT',
 		INTERNAL_PROXY_SERVER_ERROR: 'SUITECLOUD_AUTH_PROXY_SERVICE_INTERNAL_PROXY_SERVER_ERROR',
+		INVALID_BASE_URL_ERROR: 'SUITECLOUD_AUTH_PROXY_SERVICE_INVALID_BASE_URL_ERROR'
```
davidecorreu: What about `REQUEST_PATH_NOT_ALLOWED_ERROR` ?

packages/node-cli/messages.json
Montsant19: In similar messages we use the quotes. I would use them here as well.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -137,6 +138,16 @@ const initializeDevAssistService = (devAssistStatusBar: vscode.StatusBarItem) =>
         vsLogger.error('');
     });
 
+    devAssistProxyService.on(PROXY_SERVICE_EVENTS.NOT_ALLOWED_PATH_ERROR, (emitParams: { authId: string, message: string }) => {
+        //path for the error message: http://127.0.0.1:[port]/api/internal/devassist/
+        let path = `${DEVASSIST.PROXY_URL.SCHEME}${DEVASSIST.PROXY_URL.LOCALHOST_IP}:[port]${DEVASSIST.PROXY_URL.PATH}`;
+        path = path.endsWith('/') ? path : path.concat('/');
```
davidecorreu: After checking current `packages/vscode-extension/src/ApplicationConstants.ts`.`PROXY_URL.PATH` I understand why you have been conditionally adding `/` at the `path` here and  in other places in this pull request.
I really thing it will be more efficient to just create a new variable that includes it. It could be named something like `PROXY_PATH_FILTER` or  `ALLOWED_PROXY_PATH_PREFIX` .

packages/node-cli/src/services/TranslationKeys.js
TheJorgeBorras: +1

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: You can choose any name, all of them are valid, and it is the url of the proxy. But if you like better _allowedPathPrefix I can rename it.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: You got a point

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: I disagree. /api/internal/devassist2 matches this path but it is invalid.

packages/node-cli/messages.json
carolnavarropereira: I had to remove it because the output in cline modified the content scaping the characters . It showed th is: \/api\/interna\l/devassist\/

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -82,6 +84,16 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+			if (!this._matchesDevAssistPath(request.url)) {
+				//path for the error message: /api/internal/devassist/*
+				const path = this._proxyUrl.PATH.endsWith('/') ? this._proxyUrl.PATH : this._proxyUrl.PATH.concat('/');
+				const errorMessage = NodeTranslationService.getMessage(SUITECLOUD_AUTH_PROXY_SERVICE.INVALID_BASE_URL_ERROR, path);
+				this._writeResponseMessage(response, HTTP_RESPONSE_CODE.FORBIDDEN, errorMessage);
+				//We do not send the error message because in this case vscode will not use it.
+				this._handleListeningErrors('', EVENTS.NOT_ALLOWED_PATH_ERROR);
+				return;
+			}
+
```
TheJorgeBorras: I'm Okay with the changes and what you are doing here, but this class is already too hard to follow.

Is there a way we could change the name of the function `_matchesDevAssistPath`? Maybe wrap everything on a separate function? add a comment on top of the if statement?

I would like to make it crystal clear that this part of the code is what takes care of filtering the URL Path.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
TheJorgeBorras: While we are cleaning this part of the code, we should remove the comment referencing a REGEX match that no longer exists.

The JSDoc comment isn't accurate anymore either (and doesn't provide relevant info). I'd prefer to remove it, but If you want to update it I'll be okay with it.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
TheJorgeBorras: `this._proxyUrl.PATH` should always end with a `/`, I like this (so that the validation makes sense).

But we shouldn't be checking for it's existence here.
Check and add it to the constant on the constructor or on the code that calls the constructor (either is fine with me but David has already made his mind 🙃)

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -48,10 +49,11 @@ const HTTP_RESPONSE_CODE = {
 };
 
 class SuiteCloudAuthProxyService extends EventEmitter {
-	constructor(sdkPath, executionEnvironmentContext) {
+	constructor(sdkPath, executionEnvironmentContext, allowedUrl) {
 		super();
 		this._sdkPath = sdkPath;
 		this._executionEnvironmentContext = executionEnvironmentContext;
+		this._allowedPathPrefix = this._extractPathPrefix(allowedUrl);
```
davidecorreu: It unnecessary "complex" to send an `allowedUrl` and then extract the `allowedPathPrefix` by using `this._extractPathPrefix(allowedUrl)`.

It would be as simple as just pass the `allowedPathPrefix` in the constructor. Please refactor this constructor to make the code simpler.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -158,6 +184,16 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+	/**
+	 * Matches /api/internal/devassist/ followed by one or more valid URL path characters
+	 * @param path
+	 * @returns {boolean}
+	 * @private
+	 */
+	_requestPathPatternValid(path) {
+		return path.startsWith(this._allowedPathPrefix);
+	}
+
```
davidecorreu, line 176: I don't think we should mention devassist in the comments. 
The parameter could be named `requestPath` to make it clear we are talking about request paths and not fileSystem paths.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -82,6 +84,17 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
 
+			if (!this._requestPathPatternValid(request.url)) {
+				const errorMessage = NodeTranslationService.
+				getMessage(SUITECLOUD_AUTH_PROXY_SERVICE.REQUEST_PATH_NOT_ALLOWED_ERROR, this._allowedPathPrefix);
+
+				this._writeResponseMessage(response, HTTP_RESPONSE_CODE.FORBIDDEN, errorMessage);
+
+				//We do not send the error message because in this case vscode will not use it.
+				this._handleListeningErrors('', EVENTS.PATH_NOT_ALLOWED_ERROR);
+				return;
```
davidecorreu: I would send the errorMessage into the emitted event even when vscode is not using the information.
It could be useful for for some other consumer.
Also I think is a good idea to follow same pattern for most of the emitted events by SuiteCloudAuthProxyService, and the extra information won't hurt in any case.

I would prefer also to remove the comment as this service implementation should be agnostic to vscode/devassist implementation.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
TheJorgeBorras: +1

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
TheJorgeBorras, line 176: I think we should remove the comment as it doesn't provide much info. Both it and the name of function are misleading to me (we are not running a REGEX anymore). I would keep things simple another function name (and rework/remove the JSDoc comment):

* `isPathAllowed`
* `verifyRequestPath`
* `validatePathIsAllowed`
* `validateRequestPathIsAllowed`
...

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: I think this comment has not been addressed. I really think a constant like `ALLOWED_PROXY_PATH_PREFIX="/api/internal/devassist/"` will solve many of my code comments.

packages/vscode-extension/src/ApplicationConstants.ts
```diff
@@ -48,5 +48,9 @@ export const DEVASSIST = {
...
 		LOCALHOST_IP: '127.0.0.1',
 		PATH: '/api/internal/devassist'
-	}
-}
+	},
+};
+
+// Allowed prefix for proxy server validation and error messaging.
+// The format is http(s)://host:[PORT]/requestPath  [PORT] should be replaced by the current port.
+export const PROXY_SERVER_ALLOWED_URL: string ='http://127.0.0.1:[PORT]/api/internal/devassist/';
```
davidecorreu: Just to be clear, this is the constant I would like to change for something similar one of the following:
 - `export const ALLOWED_REQUEST_PATH_PREFIX='/api/internal/devassist/'`
 - `export const ALLOWED_PROXY_PATH_PREFIX='/api/internal/devassist/'`
 - `export const ALLOWED_PROXY_PATH_REQUEST_PREFIX='/api/internal/devassist/'`
 - `export const ALLOWED_PATH_REQUEST_PREFIX='/api/internal/devassist/'`

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
TheJorgeBorras: I would also be happy if you removed this block of code as well. It's an alternative solution you could go for... Personally, I would be fine if you called `path.startsWith(this._allowedPathPrefix);` directly under the `this._localProxy.addListener`.

It

davidecorreu: Another think about this validation method, let's write in a way it will work if `_allowedPathPrefix` is undefined.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: Agreed to change it.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: Already addressed.

packages/vscode-extension/src/ApplicationConstants.ts
davidecorreu: `DEVASSIST.ALLOWED_PROXY_PATH_PREFIX: '/api/internal/devassist/'` has been created.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: Resolved.

PR #977, opened 2025-12-01
feature/pdpdevtool-6268: apiKey authorization use for devassist service

davidecorreu:


packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -11,11 +11,12 @@ const EventEmitter = require('events');
...
 const EVENTS = {
-	SERVER_ERROR: 'serverError',
-	SERVER_ERROR_ON_REFRESH: 'serverErrorOnRefresh',
 	AUTH_REFRESH_MANUAL_EVENT: 'authRefreshManual',
 	PROXY_ERROR: 'proxyError',
-	PATH_NOT_ALLOWED_ERROR: 'pathNotAllowedError'
+	REQUEST_PATH_NOT_ALLOWED: 'requestPathNotAllowed',
+	SERVER_ERROR: 'serverError',
+	SERVER_ERROR_ON_REFRESH: 'serverErrorOnRefresh',
+	UNAUTHORIZED_PROXY_REQUEST: 'unauthorizedProxyRequest',
```
TheJorgeBorras: This list is becoming very big and it doesn't help that the different error types are mixed together.
In my opinion we could group them to make the code clearer:

```
const EVENTS = {
    SERVER_ERROR: {
        DEFAULT: 'serverError',
        ON_AUTH_REFRESH: 'authRefreshManual',
    },
    PROXY_ERROR: {
        DEFAULT: 'proxyError',
        MANUAL_AUTH_REFRESH_MANUAL_NEEDED: 'authRefreshManual'
    },
    REQUEST_ERROR: {
        PATH_NOT_ALLOWED: 'pathNotAllowedError',
        UNAUTHORIZED: 'unauthorizedProxyRequest'
    },
}
```

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -72,24 +75,19 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
-			if (!this._isPathAllowed(request.url)) {
-				const errorMessage = NodeTranslationService.
-				getMessage(SUITECLOUD_AUTH_PROXY_SERVICE.REQUEST_PATH_NOT_ALLOWED_ERROR, this._allowedPathPrefix);
-
-				this._writeResponseMessage(response, HTTP_RESPONSE_CODE.FORBIDDEN, errorMessage);
-
-				this._handleListeningErrors(errorMessage, EVENTS.PATH_NOT_ALLOWED_ERROR);
-				return;
+			// Validate incoming request (apiKey & allowed path) in a dedicated method
+			if (!this._isValidIncomingRequest(request, response)) {
```
TheJorgeBorras: `in a dedicated method` does not provide additional info, therefore is redudant here. 
It also looks a tad ugly to me `apiKey` being camelCase but `allowed path` being separate words.

```
// Validate incoming request (Api Key & URL Path)
Comment view
			if (!this._isValidIncomingRequest(request, response)) { ... }
``

File: packages/node-cli/src/services/SuiteCloudAuthProxyService.js [hunk 3]
```diff
@@ -195,6 +205,46 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+	 * @param {http.ServerResponse} response
+	 * @returns {boolean} true if valid, false if rejected
+	 */
+	_isValidIncomingRequest(request, response) {
+		// Authentication filter: check authorization header if an API key is configured
+		if (this._apiKey) {
+			const authHeader = request.headers['authorization'];
+			// TODO remove console logs
+			console.log('--->>>   SuiteCloudAuthProxyService.validateIncomingRequest.  <<<---');
+			console.log({ authHeader, proxyApiKey: this._apiKey })
```
TheJorgeBorras: Todo: remove logs (reminder)

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -195,6 +205,46 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+				this._writeResponseMessage(response, HTTP_RESPONSE_CODE.UNAUTHORIZED, unauthorizedMessage);
+				const emitData = { message: unauthorizedMessage, authId: this._authId, requestUrl: request.url };
+				this.emit(EVENTS.UNAUTHORIZED_PROXY_REQUEST, emitData);
+				return false;
+			}
+		}
+
+		// Allowed path filter: check allowed prefix if configured
+		if (this._allowedPathPrefix) {
+			if (!request.url.startsWith(this._allowedPathPrefix)) {
```
TheJorgeBorras: Unnecessary nested if.

Let's go for:
```if (this._allowedPathPrefix && !request.url.startsWith(this._allowedPathPrefix)) {```

File: packages/vscode-extension/package.json [hunk 5]
```diff
@@ -117,6 +117,17 @@
...
+            {
+              "command": "suitecloud.createdevassistapikey",
+              "title": "Create Developer Assistant Service API Key",
+              "category": "SuiteCloud",
+              "enablement": "false"
+            },
+            {
+              "command": "suitecloud.deletedevassistapikey",
+              "title": "Delete stored Developer Assistant Service API Key - TODO: delete this command",
+              "category": "SuiteCloud"
```
TheJorgeBorras, line 130: TODO: Remove this command

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -117,17 +115,40 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+						reject(err);
+					} else {
+						console.log('SuiteCloud Proxy server stopped.');
+						resolve();
+					}
+				});
+			});
+
+			await closePromise;
 			this._localProxy = null;
```
carolnavarropereira: Maybe we can do it in another way.  Move this._localProxy to a temp variable, and then set this._localProxy to null the first task, then keep the rest of the code using the temp variable instead.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -117,17 +115,40 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+	 * Updates the stored API key, which is used to authenticate and filter incoming requests to the local server.
+	 *
+	 * @param {string} newApiKey - The new API key to set for request filtering.
+	 * @returns {void}
+	 */
+	updateApiKey(newApiKey) {
+		this._apiKey = newApiKey;
+	}
+
+
```
carolnavarropereira, line 165: use _ as prefix  as standart for inner methods. Rename as _updateApiKey

packages/vscode-extension/messages.json
```diff
@@ -57,6 +57,14 @@
 	"DEPLOY_QUESTIONS_CHOICES_ACCOUNT_SPECIFIC_VALUES_CANCEL_PROCESS": "Cancel the deployment process",
 	"DEPLOY_QUESTIONS_CHOICES_ACCOUNT_SPECIFIC_VALUES_DISPLAY_WARNING": "Display a warning and continue the deployment process",
 
+	"DEVASSIST_SERVICE_CREATE_API_KEY_CANCELED_MESSAGE": "API key generation canceled. ",
+	"DEVASSIST_SERVICE_CREATE_API_KEY_CONFIRMATION_MESSGE": "API key copied to clipboard. ",
+	"DEVASSIST_SERVICE_CREATE_API_KEY_INITIAL_CREATION_ERROR": "The API key was not generated. Enable SuiteCloud Developer Assistant service again to generate the required API key.",
+	"DEVASSIST_SERVICE_CREATE_API_KEY_MODAL_MAIN_MESSAGE": "A new API key has been generated for SuiteCloud Developer Assistant:\n\n{0}\n\nPaste the API key into the OpenAI Compatible API Key field in Cline settings.\n\nWhen your SuiteCloud Developer Assistant service starts, copy the base URL from the SuiteCloud output panel into the Base URL field in Cline settings.\n\nKeep this API key confidential and do not share it.",
+	"DEVASSIST_SERVICE_CREATE_API_KEY_MODAL_COPY_BUTTON": "Copy API Key",
+
+	"DEVASSIST_SERVICE_EMIT_ERROR_EMIT_ERROR_BUTTON_CREATE_NEW_API_KEY": "Generate New API Key",
```
TheJorgeBorras: In some comments I see `API key` and in others `API Key`.

My guess is the first is the right one, but since its the Specific Name of a Specific CLINE Field maybe we should go for the second one? 

please confirm and/or review the comments.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -195,6 +205,46 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
 
+	/**
+	 * Validates an incoming proxy request for apiKey & allowed path.
+	 * Responds and emits the correct events on failure.
+	 * @param {http.IncomingMessage} request
+	 * @param {http.ServerResponse} response
+	 * @returns {boolean} true if valid, false if rejected
+	 */
+	_isValidIncomingRequest(request, response) {
+		// Authentication filter: check authorization header if an API key is configured
```
carolnavarropereira: _isValidIncomingRequest does two different validations. For readibility I will keep the _isValidIncomingRequest which makes a call to two different methods. The reason for this is to keep it clean this method in case we modify any validation preventing the interaction between each other, something like that: (you can choose the name for your methods and add the proper brackets)
_isValidIncomingRequest () {
  if (!isValidIncomingHeader()) return false;
  if (!isValidAllowedPathPrefix()) return false;
  return true;
}

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: We have a method to _updateApiKey, for keeping consistant all the access to this method should be wrapped then, so instead of reading it directly, you can create _getApiKey()

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -195,6 +205,46 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+			// TODO remove console logs
+			console.log('--->>>   SuiteCloudAuthProxyService.validateIncomingRequest.  <<<---');
+			console.log({ authHeader, proxyApiKey: this._apiKey })
+			if (authHeader !== `Bearer ${this._apiKey}`) {
+				const unauthorizedMessage = NodeTranslationService.getMessage(SUITECLOUD_AUTH_PROXY_SERVICE.UNAUTHORIZED_PROXY_REQUEST);
+				// using 401-Unauthorized http response code as CLINE won't activate the retry mechanism with it
+				// not using 407-Proxy Authentication Required as CLINE activates the retry mechanism with it
+				this._writeResponseMessage(response, HTTP_RESPONSE_CODE.UNAUTHORIZED, unauthorizedMessage);
+				const emitData = { message: unauthorizedMessage, authId: this._authId, requestUrl: request.url };
+				this.emit(EVENTS.UNAUTHORIZED_PROXY_REQUEST, emitData);
```
carolnavarropereira: Maybe you can use _handleListeningErrors modifying to accept a optional third parameter which will be the url instead of building the emit and responseMessage manually and keep it consistant with the rest of the code.

packages/vscode-extension/src/util/APIKeyGenerator.ts
```diff
@@ -0,0 +1,10 @@
+import { randomBytes } from 'crypto';
+
+/**
+ * Generates a secure random API key.
+ * @param length Number of random bytes (default 32 = 64 hex chars).
+ * @returns Hex-encoded API key.
+ */
+export  const generateApiKey = (length = 32): string => {
+  return randomBytes(length).toString('hex');
+}
```
TheJorgeBorras, line 10: Not sure if you knew about it, but we have this function on Node CLI:

```
# packages/node-cli/src/utils/CryptoUtils.js
function generateRandomKey() {
	return crypto
		.randomBytes(ENCRYPTION_KEY_BUFFER_LENGTH)
		.toString(HEX)
		.substr(0, 32);
}
```


As far as I can tell, they do the same. Do we really need this code duplicated?
If yes, then make it more generic and call this class something like "CryptoUtils", "TokenGenerationUtils". or "RandomTokenGenerationUtils"

packages/vscode-extension/src/commands/CreateDevAssistApiKey.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Registers the suitecloud.createdevassistapikey command.
+ * This command generates and stores a Developer Assistant API Key,
+ * then displays it in a modal dialog.
+ */
+/**
+ * Command handler for suitecloud.createdevassistapikey
+ */
+export async function createDevAssistApiKeyCommand(extensionContext: vscode.ExtensionContext): Promise<string | undefined> {
```
TheJorgeBorras: This file practically has the same amount of **Comment lines** as **Code lines**.

Can we trim some of the fat here? Surely all comments are not needed here (most of it is deducible)

Low priority comment, feel free to ignore (although I'd like you to address it)

packages/vscode-extension/src/commands/CreateDevAssistApiKey.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    // DEVASSIST_SERVICE.CREATE_API_KEY.MODAL.COPY_BUTTON
+    if (selection === copyButtonText) {
+        // save apiKey
+        await extensionContext.secrets.store(DEVASSIST.SECRET_KEY, apiKey);
+        // copy apiKey to clipboard
+        vscode.env.clipboard.writeText(apiKey);
+        // confirmation message
+        vscode.window.showInformationMessage(translationService.getMessage(DEVASSIST_SERVICE.CREATE_API_KEY.CONFIRMATION_MESSGE));
+        // show suitecloud output
```
carolnavarropereira: Small typo CONFIRMATION_MESSGE, should be MESSAGE

packages/vscode-extension/src/ApplicationConstants.ts
```diff
@@ -44,11 +45,12 @@ export const DEVASSIST = {
...
 	},
+	MODELS_PATH: '/api/internal/devassist/models',
 	PROXY_URL: {
 		SCHEME: 'http://',
 		LOCALHOST_IP: '127.0.0.1',
 		BASE_PATH: '/api/internal/devassist',
 		FEEDBACK_PATH: '/api/internal/devassist/feedback',
 	},
-	ALLOWED_PROXY_PATH_PREFIX: '/api/internal/devassist/'
+	SECRET_KEY: 'DEVASSIT_API_KEY',
```
TheJorgeBorras: I would prefer something like `SECRET_STORAGE_KEY_ID`.

It feels clearer to me. Let me know what you think

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -117,17 +115,45 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
 			console.log('No server instance to stop.');
 		}
+		
+		this._localProxy = null;
 	}
 
+	/**
+	 * Updates the stored API key, which is used to authenticate and filter incoming requests to the local server.
+	 *
+	 * @param {string} newApiKey - The new API key to set for request filtering.
```
TheJorgeBorras, line 158: Something to consider here, maybe we could emphasize that this function whole purpose is to expose a way to update the API Key from the outside (VSCode Layer).

packages/vscode-extension/src/suitecloud.ts
```diff
@@ -83,23 +84,41 @@ export async function activate(context: vscode.ExtensionContext) {
...
+			() => openDevAssistFeedbackForm(context)),
+		// Command to create and store Developer Assistant service API Key
+		vscode.commands.registerCommand('suitecloud.createdevassistapikey',
+			() => createDevAssistApiKeyCommand(context)
+		)
 	);
 
-	// DevAssist Feedback Form WebView
+	// TODO: reomve this command
+	// Command to check Develoepr Assistant service API Key
```
carolnavarropereira: several spelling words, remove and developer.

carolnavarropereira (changes requested):

packages/vscode-extension/src/suitecloud.ts
```diff
@@ -83,23 +84,41 @@ export async function activate(context: vscode.ExtensionContext) {
...
 	context.subscriptions.push(
-		vscode.commands.registerCommand('suitecloud.opendevassistfeedbackform',
-			() => openDevAssistFeedbackForm(context)
+		vscode.commands.registerCommand('suitecloud.deletedevassistapikey',
+			async () => {
+				const apiKey = await context.secrets.get(DEVASSIST.SECRET_KEY);
+				// delete Devleoper Assistant API Key
+				await context.secrets.delete(DEVASSIST.SECRET_KEY);
```
carolnavarropereira: Spelling error Developer

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: comment replaced for `// Validate incoming request (Api Key & Request Path)`

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: Categorization done.
I have also renamed `authRefreshManual` value for `manualAuthRefreshRequired`.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: This is not a "private" method but "public".

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: I'll need to think about it. We might need to talk about it as I see pros and cons.

packages/vscode-extension/src/commands/CreateDevAssistApiKey.ts
davidecorreu: I'll remove some of the comments in the next commit.

packages/vscode-extension/src/commands/CreateDevAssistApiKey.ts
davidecorreu: Fixing in next commit.

packages/vscode-extension/src/ApplicationConstants.ts
davidecorreu: Changing it in next commit.

packages/vscode-extension/src/suitecloud.ts
davidecorreu: Changing it in the next commit although this code will be removed.

packages/vscode-extension/src/suitecloud.ts
davidecorreu: Changing it in the next commit although this code will be removed.

packages/vscode-extension/messages.json
davidecorreu: I asked Montsant for clarification. Her input was: 
 - Use “API Key” with a capital K in titles and labels (especially after a colon).
 - Use “API key” with a lowercase k within sentences in normal text.

Revision of messages has been done by Montsant and are in JIRA story description. If you find any discrepancy please let me know.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
TheJorgeBorras: Here is my take: please don't. I dislike 1 or 2 liner functions and this file has already too many functions and nested logic.

If you really want 2 functions you could break it down into 2 and call them from `start(...)`:
Then this `_isValidIncomingRequest()`

```
if (isValidIncomingHeader() && isValidAllowedPathPrefix()) { 
     ...
}
```

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: Done.

packages/vscode-extension/src/commands/CreateDevAssistApiKey.ts
davidecorreu: Check new version and let me know.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -97,51 +104,109 @@ export const devAssistConfigurationChangeHandler = async (configurationChangeEve
...
+ * Handles authentication, error, and proxy events.
+ */
+const addListenersToDevAssistProxyService = (devAssistProxyService: SuiteCloudAuthProxyService, devAssistStatusBar: vscode.StatusBarItem) => {
+    // Listener to trigger manual reauthentication from vscode
     devAssistProxyService.on(PROXY_SERVICE_EVENTS.REAUTHORIZE, async (emitParams: { authId: string, message: string }) => {
-        // trigger refresh on emited authID 
         const refreshIsSuccessful = await refreshAuthorizationWithNotifications(emitParams.authId);
         if (refreshIsSuccessful) {
-            // edgy case: user might have changed configured authid while waiting for the manual refresh to complete
             updateDevAssistConfigStatus();
```
TheJorgeBorras, line 168: This chunk of code has as the same number of lines of code as lines of comments.
Let's try to remove some of them

packages/vscode-extension/src/util/APIKeyGenerator.ts
davidecorreu: The functions are similar but not doing exactly the same.

In any case I would like to decouple this logic from node-cli layer. I think we should keep the file.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: I think the current comment is clear enough, but I’m open to other proposals if the team prefers. If needed, I’d suggest using a more generic reference instead of mentioning VSCode specifically to keep things flexible.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: Agreed on a solution to be check in next commit.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: We might change it or not.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -42,11 +49,13 @@ const LOCAL_HOSTNAME = '127.0.0.1';
...
 
 class SuiteCloudAuthProxyService extends EventEmitter {
-	constructor(sdkPath, executionEnvironmentContext, allowedPathPrefix) {
+
+	constructor(sdkPath, executionEnvironmentContext, allowedPathPrefix, apiKey) {
 		super();
 		this._sdkPath = sdkPath;
 		this._executionEnvironmentContext = executionEnvironmentContext;
 		this._allowedPathPrefix = allowedPathPrefix;
+		this._apiKey = apiKey;
```
TheJorgeBorras, line 58: (Putting this comment here as there isn't a good place to do so)

Reminder: Please update the code of the Feedback Forf so it doesn't get blocked by the API Key Filter

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -47,9 +48,10 @@ let devAssistProxyService: SuiteCloudAuthProxyService;
 const vsLogger: VSConsoleLogger = new VSConsoleLogger();
 const vsNotificationService = new MessageService('DevAssistService');
 const translationService = new VSTranslationService();
+let awaitingApiKeyCreation = false;
```
TheJorgeBorras, line 51: I've been thinking of a better way of implementing this "Lock" or "Traffic Light", but found no better answer.

We should leave some comments in the right places (I think the ones we have now are OK) and personally, I would rename de variable to:
* apiKeyCreationLock
* isCreatingApiKeyLock
* apiKeyCreationInProgressLock

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -97,51 +104,109 @@ export const devAssistConfigurationChangeHandler = async (configurationChangeEve
...
+        }
+        else (
+            await initializeDevAssistService(extensionContext, devAssistStatusBar)
+        )
+    }
+};
 
-    // adding listener to trigger manual reauthentication from vscode
+const initializeDevAssistService = async (extensionContext: vscode.ExtensionContext, devAssistStatusBar: vscode.StatusBarItem) => {
+    // check if API Key is available
```
TheJorgeBorras: unnecessary comment

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -97,51 +104,109 @@ export const devAssistConfigurationChangeHandler = async (configurationChangeEve
...
+        awaitingApiKeyCreation = false;
+        // no apiKey was created
+        if (!createApiKeyCommandResult) {
+            // disable devassist service via setting change to force user to do it again if required
+            const devAssistConfigSection = vscode.workspace.getConfiguration(DEVASSIST.CONFIG_KEYS.devAssistSection);
+            devAssistConfigSection.update(DEVASSIST.CONFIG_KEYS.proxyEnabled, false);
+            
+            // throw to stop initialization flow
+            const initialApiKeyCreationError = translationService.getMessage(DEVASSIST_SERVICE.CREATE_API_KEY.INITIAL_CREATION_ERROR)
+            throw initialApiKeyCreationError;
```
TheJorgeBorras, line 142: Comments on line 143, 145 and 149 are safe remove I think

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -97,51 +104,109 @@ export const devAssistConfigurationChangeHandler = async (configurationChangeEve
...
+        }
+
+        // apiKey was created by triggerCreateNewApiKeyCommand
+        // there should no no way a devAssistProxyService could be created without devassistApiKey
+        devAssistProxyService = new SuiteCloudAuthProxyService(getSdkPath(), executionEnvironmentContext, DEVASSIST.ALLOWED_PROXY_PATH_PREFIX, createApiKeyCommandResult as string);
+        // Set up all event listeners in one place
+        addListenersToDevAssistProxyService(devAssistProxyService, devAssistStatusBar);
```
TheJorgeBorras, line 147: no need for comments on line 155 and 157.

154 we can keep

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -97,51 +104,109 @@ export const devAssistConfigurationChangeHandler = async (configurationChangeEve
...
+        return
+    }
+
+    if (devAssistProxyService) {
+        // if an existing instance was already created just update the
+        devAssistProxyService.updateApiKey(devassistApiKey)
+    } else {
+        devAssistProxyService = new SuiteCloudAuthProxyService(getSdkPath(), executionEnvironmentContext, DEVASSIST.ALLOWED_PROXY_PATH_PREFIX, devassistApiKey);
+        // Set up all event listeners in one place
+        addListenersToDevAssistProxyService(devAssistProxyService, devAssistStatusBar);
```
TheJorgeBorras, line 155: comments 163 we can remove.

167 you can remove or substitute it for a comment after the else saying
"// instantiating a new SuiteCloudAuthService and set up listeners"

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -97,51 +104,109 @@ export const devAssistConfigurationChangeHandler = async (configurationChangeEve
...
+
+    // this should only happen when no secret is available (initial devassist setup)
+    if (!devassistApiKey) {
+        // starting the here the creation of the devassist apiKey if no secret is available
+
+        awaitingApiKeyCreation = true;
+        // if the user saves the new apiKey the devAssistSecretApiKeyChangeHandler will be called 
+        // using awaitingApiKeyCreation status to block undesired call to initializeDevAssistService
+        const createApiKeyCommandResult = await triggerCreateNewApiKeyCommand();
+        awaitingApiKeyCreation = false;
```
TheJorgeBorras: 134, 136, 139 and 140 could be merged into a single comment.

on line 136 we could go for (a proposal):
`// when no devassistApiKey secret is available, attempt to setup a new one and lock away ApiKeyCreation while the command is running`

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
TheJorgeBorras: One of my ideas was to somehow keep track of the variable on a separate class (a service or maybe a datastructure on the class responsible for the DevAssist Api Key Secret Creation), and have the check over there but it doesn't feel much better than this.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -201,23 +277,29 @@ const showDevAssistStartUpNotification = () => {
...
+    ];
+    vsNotificationService.showCommandInfoWithSpecificButtonsAndActions(infoMessage, buttonsAndActions);
 }
 
+// stoped status bar
+// logs into output
+//      (DEVASSIST_SERVICE.IS_STOPPED.OUTPUT, error)
+// show notifcation
+//      (DEVASSIST_SERVICE.IS_STOPPED.NOTIFICATION)
+//      with See Detials button (opens output and devassistSettings)
```
TheJorgeBorras, line 301: TODO: clean code after draft is finished

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -237,6 +319,11 @@ const showStartDevAssistProblemNotification = (errorStage: string, error: string
     vsNotificationService.showCommandErrorWithSpecificButtonsAndActions(errorMessage, buttonsAndActions);
 }
 
+// logs into output
+//      emitError
+// show notifcation
+//      (DEVASSIST_SERVICE.IS_STOPPED.NOTIFICATION)
+//      with See Detials button (opens output)
```
TheJorgeBorras, line 325: TODO: clean code after draft is finished

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -250,6 +337,31 @@ const showDevAssistEmitProblemLog = (errorStage: string, outputError: string, de
     vsNotificationService.showCommandErrorWithSpecificButtonsAndActions(errorMessage, buttonsAndActions);
 };
 
+
+// logs into output
+//      outputErrorMessge
+// show notifcation
+//      something about wrong API key
+//      with See Detials button (opens output)
```
TheJorgeBorras, line 344: TODO: clean code after draft is finished

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -250,6 +337,31 @@ const showDevAssistEmitProblemLog = (errorStage: string, outputError: string, de
...
+    ];
+    vsNotificationService.showCommandErrorWithSpecificButtonsAndActions(outputErrorMessge, buttonsAndActions);
+};
+
+// stoped status bar
+// logs into output
+//      emitError
+// show notifcation
+//      (DEVASSIST_SERVICE.IS_STOPPED.NOTIFICATION)
+//      with See Detials button (opens output)
```
TheJorgeBorras, line 363: TODO: clean code after draft is finished

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
```diff
@@ -59,21 +61,26 @@ export const startDevAssistProxyIfEnabled = async (devAssistStatusBar: vscode.St
...
             showStartDevAssistProblemNotification('startup', error as string, devAssistStatusBar);
         }
     } else {
         // TODO: We might want to propose to configure and enable service
         vsLogger.printTimestamp();
-        vsLogger.info(translationService.getMessage(DEVASSIST_SERVICE.IS_DISABLED.OUTPUT));
+        // TODO find the appropriate message
+        // we should be using a different message here, current: "SuiteCloud Developer Assistant service has been disabled."
+        // vsLogger.info(translationService.getMessage(DEVASSIST_SERVICE.IS_DISABLED.OUTPUT));
+        // it should be something like: "SuiteCloud Developer Assistant service is not enabled.\n{how to activate devassist instructions}"
```
TheJorgeBorras: TODO: Reminder to you to not forget about this part

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: About this series of comments.
While adding this comments to know what this "showFunctions" were doing in an easier observable way I thought we could create a customizable showFunction that could deal with the different options we are using.
Meanwhile if we are not implementing it I would like to keep the comments.

Of course we can talk about it and delete the comments if you think it's a no go.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: Message is already in place. Removing comment in next commit.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: Removing it.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: Removing 155 and 157.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: Removing 163 and 167.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: Resolving this one as it has been transformed into several specific ones that are being addressed.

packages/vscode-extension/src/startup/DevAssistConfiguration.ts
davidecorreu: Not performing change as we didn't find a good candidate for rename.

packages/vscode-extension/src/webviews/FeedbackFormWebviewController.ts
```diff
@@ -127,9 +127,14 @@ const handleSubmitFeedbackFormEvent = async (formData : FeedbackFormData, cssWeb
...
 		const currentProxySettings = getDevAssistCurrentSettings();
 		const requestBody = JSON.stringify(formData);
+		const apiKey = await extensionContext.secrets.get(DEVASSIST.SECRET_STORAGE_KEY_ID);
+		const authzHeaderValue = `Bearer ${apiKey}`;
 		const response = await fetch(`${PROXY_URL.SCHEME}${PROXY_URL.LOCALHOST_IP}:${currentProxySettings.localPort}${PROXY_URL.FEEDBACK_PATH}`, {
 			method: 'POST',
-			headers: { 'Content-Type': 'application/json' },
+			headers: { 
+				'Content-Type': 'application/json',
+				'authorization': authzHeaderValue
```
TheJorgeBorras: typo: authHeaderValue

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: added apiKey to feedback requests

packages/vscode-extension/src/webviews/FeedbackFormWebviewController.ts
```diff
@@ -127,9 +127,14 @@ const handleSubmitFeedbackFormEvent = async (formData : FeedbackFormData, cssWeb
 	try {
 		const currentProxySettings = getDevAssistCurrentSettings();
 		const requestBody = JSON.stringify(formData);
+		const apiKey = await extensionContext.secrets.get(DEVASSIST.SECRET_STORAGE_KEY_ID);
+		const authzHeaderValue = `Bearer ${apiKey}`;
```
TheJorgeBorras: Do we really need this `apiKey` variable here?

You can just call down bellow
``` 
headers: { 
     'Content-Type': 'application/json',
     'Authorization': `Bearer ${apiKey}` 
} 
```

packages/vscode-extension/src/suitecloud.ts
```diff
@@ -83,23 +84,41 @@ export async function activate(context: vscode.ExtensionContext) {
...
-			() => openDevAssistFeedbackForm(context)
+		vscode.commands.registerCommand('suitecloud.deletedevassistapikey',
+			async () => {
+				const apiKey = await context.secrets.get(DEVASSIST.SECRET_STORAGE_KEY_ID);
+				// delete Developer Assistant API Key
+				await context.secrets.delete(DEVASSIST.SECRET_STORAGE_KEY_ID);
+				// Show previously stored key
+				const message = `This is the deleted API key for Developer Assistant: ${apiKey}`;
+				vscode.window.showInformationMessage(message, { modal: true },);
+			}
```
TheJorgeBorras, line 112: TODO: remove this command after testing

packages/vscode-extension/src/types/JavascriptNodeCli.d.ts
```diff
@@ -39,9 +39,10 @@ export interface ExecutionEnvironmentContextConstructor {
 
 export interface SuiteCloudAuthProxyServiceInstance extends EventEmitter {
 	start(authId: string, localProxyPort: number): Promise<void>;
-	stop(): void;
-	reloadAccessToken(): Promise<void>
+	stop(): Promise<void>;
+	reloadAccessToken(): Promise<void>;
+	updateApiKey(newApiKey?: string): void;
```
TheJorgeBorras, line 44: why is `newApiKey` defined as optional?

TheJorgeBorras (approved):

packages/vscode-extension/src/types/JavascriptNodeCli.d.ts
davidecorreu: Making this parameter optional reduces verbosity in other parts of the code. When retrieving an extension secret, the value can be undefined, so allowing the parameter to be optional simplifies the usage.

packages/vscode-extension/src/webviews/FeedbackFormWebviewController.ts
davidecorreu: Changed in next commit.

PR #987, opened 2025-12-17
add text/event-stream accept header

davidecorreu:


packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -304,6 +304,15 @@ class SuiteCloudAuthProxyService extends EventEmitter {
 			headers: { ...request.headers, host, authorization },
 		};
 
+		// added to get "stream responses" from netsuite devassist backend
+		const DEVASSIST_CHAT_COMPLETIONS_PATH = '/api/internal/devassist/chat/completions'
+		if (DEVASSIST_CHAT_COMPLETIONS_PATH.match(request.url)) {
```
TheJorgeBorras: I would prefer a `.startsWith(...)` instead of a `.match(...)`.
Both are good enough, but `.startsWith(...)`  It's more consistent with the way we read URL endpoints across the code (for instance, the allowed path code).

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -304,6 +304,15 @@ class SuiteCloudAuthProxyService extends EventEmitter {
 			headers: { ...request.headers, host, authorization },
 		};
 
+		// added to get "stream responses" from netsuite devassist backend
+		const DEVASSIST_CHAT_COMPLETIONS_PATH = '/api/internal/devassist/chat/completions'
```
TheJorgeBorras: I guess this is a good place to declare `DEVASSIST_CHAT_COMPLETIONS_PATH`, but may be it's better to do so on **line ~45** alongside the rest of `SuiteCloudAuthProxyService` constants?

Just a thought, feel free to leave it as it is.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -304,6 +304,15 @@ class SuiteCloudAuthProxyService extends EventEmitter {
 			headers: { ...request.headers, host, authorization },
 		};
 
+		// added to get "stream responses" from netsuite devassist backend
```
TheJorgeBorras, line 308: I feel like something is missing on this comment... 

```
// Accept multiple server responses as a "stream of text", instead of a single response containing the full answer generated by the LLM (once it finishes writing it).
```

Let me know if it's too much. I'm open to suggestions or even keeping it as it is

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: It is not the same, startsWith will include more cases, I would leave it as match

TheJorgeBorras (approved):

carolnavarropereira (approved):

PR #993, opened 2026-01-27
Add dataItem property to component type definition

Extend-Apps:
Added missing property to GridRow (tests confirmed existence)

### DIFF
--- a/packages/uif-types/uif/component.d.ts
+++ b/packages/uif-types/uif/component.d.ts
@@ -9557,6 +9557,8 @@ declare module '@uif-js/component' {
 
 		userData: object;
 
+		dataItem: any;
+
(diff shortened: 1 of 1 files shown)


alisyed-19:
@odanek Can you please take a look?

odanek:
Hi @Extend-Apps, adding the `dataItem` property to `GridRow` is technically not correct. `GridRow` is a base class for multiple types of rows, e.g. `GridDataRow`, `GridSyntheticRow` and `GridHeaderRow` where only the first one has an associated data item. And `GridDataRow` already defines the `dataItem` property.

So maybe the actual issue is that we are returning `GridRow` in a place where we should be in fact returning `GridDataRow`?

PR #1010, opened 2026-04-13
PDP-DEVTOOL 6434: Dev Assist proxy:generatekey command

TheJorgeBorras:


packages/node-cli/src/commands/proxy/generatekey/ProxyGenerateKeyAction.js
```diff
@@ -0,0 +1,48 @@
+/*
+ ** Copyright (c) 2024 Oracle and/or its affiliates.  All rights reserved.
+ ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+ */
+'use strict';
+
+const { ActionResult } = require('../../../services/actionresult/ActionResult');
+const BaseAction = require('../../base/BaseAction');
+const APIKeyGenerator = require('../../../utils/APIKeyGenerator');
+const ClientApiKeyDTO = require('../../../utils/ClientApiKeyDTO');
```
carolnavarropereira: Don't see the need to import the full class instead of only generateApiKey but I think it is not important. You can just ignore it.

packages/node-cli/src/utils/ClientApiKeyUtils.js
```diff
@@ -0,0 +1,68 @@
...
+
+	const operationResult = await executeWithSpinner({
+		action: sdkExecutor.execute(executionContext),
+		message: NodeTranslationService.getMessage(CLIENT_API_KEY_UTILS.READING_FILE_CONTENTS),
+	}).catch((error) => {
+		throw [NodeTranslationService.getMessage(CLIENT_API_KEY_UTILS.ERRORS.READING_FILE_CONTENTS, error),];
+	});
+
+	if (operationResult.status === SdkOperationResultUtils.STATUS.ERROR) {
+		throw [NodeTranslationService.getMessage(CLIENT_API_KEY_UTILS.ERRORS.READING_FILE_CONTENTS, operationResult.errorMessages)];
```
carolnavarropereira: I believe there is the possibility that we do not have errorMessages even it hasn't been a successful operation, I believe the wisest we can do it is just to check. (again my opinion, I will not enforce it)
1.- are there errors? throw them
2.- else, if not sucess then throw an error (Error reading api contents or some unespecific error)
3.- if success then return the data

packages/node-cli/messages.json
```diff
@@ -307,7 +310,11 @@
 	"UTILS_AUTHENTICATION_SUCCESS_SETUP": "The account has been successfully set up.",
 	"UTILS_AUTHENTICATION_SAVING_STARTING_OAUTH_FLOW": "Continue the authentication process in the prompted browser.",
 	"UTILS_ERRORS_COMMAND": "There was an error with the {0} command. {1}",
+	"UTILS_CLIENT_API_KEY_UTILS_ERRORS_INVALID_FILE_CONTENTS": "Client API contents must be a valid format. Please, delete client_api_key.p12 file and try again.",
+	"UTILS_CLIENT_API_KEY_UTILS_ERRORS_READING_FILE_CONTENTS": "Error reading Client API Key file contents.\n{0}",
+	"UTILS_CLIENT_API_KEY_UTILS_ERRORS_WRITING_FILE_CONTENTS": "Error writing Client API Key file contents.",
+	"UTILS_CLIENT_API_KEY_UTILS_READING_FILE_CONTENTS": "Reading Client API contents...",
+	"UTILS_CLIENT_API_KEY_UTILS_WRITING_FILE_CONTENTS": "Writing Client API contents...",
```
TheJorgeBorras: Verify with Montsant

packages/node-cli/src/utils/ClientApiKeyDTO.js
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	static get Builder() {
+		return new ClientApiKeyDTOBuilder();
+	}
+};
+
+
+class ClientApiKeyDTOBuilder {
+	constructor() {
```
carolnavarropereira: Not needed default constructor

packages/node-cli/src/utils/ClientApiKeyDTO.js
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	fromRawString(stringifiedData) {
+		try {
+			this.data = !stringifiedData
```
carolnavarropereira: SonarQube: Negated conditions should be avoided when an else clause is present

packages/node-cli/src/utils/ClientApiKeyDTO.js
```diff
@@ -0,0 +1,79 @@
...
+const NodeTranslationService = require('../services/NodeTranslationService');
+const { UTILS: { CLIENT_API_KEY_UTILS }, } = require('../services/TranslationKeys');
+
+const PROXY_KEY_ID = 'PROXY_KEY';
+const PROXY_KEY_API_KEY_ID = 'proxyAPIKey';
+
+/**
+ *
+ * DTO class that represents the contents of 'client_api_key.p12'
+ * 	The expected 'client_api_key.p12' file content/format goes as follows:
```
carolnavarropereira: Just align first and second line.

packages/node-cli/src/utils/ClientApiKeyDTO.js
```diff
@@ -0,0 +1,79 @@
...
+ *		{
+ *		 	"PROXY_KEY": {
+ *		 	    "creationDate": "2026-04-13T12:46:52.577Z",
+ *		 	    "proxyApiKey": "3d3dfe3ec58eb9b9826f4546de20f96f99f6b7ed1fcdcaaa5f3c252eb69dce56"
+ *		 	}
+ *		}
+ *
```
carolnavarropereira: Remove extra lines in the comments.

packages/node-cli/src/utils/ClientApiKeyDTO.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ *
+ * DTO class that represents the contents of 'client_api_key.p12'
+ * 	The expected 'client_api_key.p12' file content/format goes as follows:
+ *		{
+ *		 	"PROXY_KEY": {
+ *		 	    "creationDate": "2026-04-13T12:46:52.577Z",
+ *		 	    "proxyApiKey": "3d3dfe3ec58eb9b9826f4546de20f96f99f6b7ed1fcdcaaa5f3c252eb69dce56"
+ *		 	}
```
carolnavarropereira: Align brackets

packages/node-cli/messages.json
```diff
@@ -177,6 +177,9 @@
 	"COMMAND_PACKAGE_LOCAL_VALIDATION_NOT_EXECUTED": "The project is being packaged without validation. Make sure the project is validated.",
 	"COMMAND_PACKAGE_PACKAGING": "Packaging...",
 
+	"COMMAND_PROXY_GENERATEKEY_ERROR_UNABLE_TO_GENERATE_KEY": "Unable to generate new Proxy key.",
```
carolnavarropereira: Proxy better lowercase.

packages/node-cli/messages.json
```diff
@@ -177,6 +177,9 @@
...
+	"COMMAND_PROXY_GENERATEKEY_SUCCESS_KEY_GENERATED": "New proxy key \"{0}\" generated, store securely and remember to update it in your third party tool of choice.\nThis Passkey will never be shown again, so if you forget it you will need to generate a new one by triggering this command again.",
```
carolnavarropereira: There is no need for Passkey being uppercase the first letter. Same for Proxy key

packages/node-cli/messages.json
```diff
@@ -307,7 +310,11 @@
 	"UTILS_AUTHENTICATION_SUCCESS_SETUP": "The account has been successfully set up.",
 	"UTILS_AUTHENTICATION_SAVING_STARTING_OAUTH_FLOW": "Continue the authentication process in the prompted browser.",
 	"UTILS_ERRORS_COMMAND": "There was an error with the {0} command. {1}",
+	"UTILS_CLIENT_API_KEY_UTILS_ERRORS_INVALID_FILE_CONTENTS": "Client API contents must be a valid format. Please, delete client_api_key.p12 file and try again.",
```
carolnavarropereira: Client API must be in a valid format. Please, delete the client_api_key.p12 file and try again.

packages/node-cli/messages.json
```diff
@@ -307,7 +310,11 @@
...
+	"UTILS_CLIENT_API_KEY_UTILS_ERRORS_READING_FILE_CONTENTS": "Error reading Client API Key file contents.\n{0}",
```
carolnavarropereira: maybe use lowercase for Key to mach what is in java cli.

packages/node-cli/postinstall.js
```diff
@@ -4,5 +4,5 @@
  */
 'use strict';
 
-require('./src/core/sdksetup/SdkLicense').show();
-require('./src/core/sdksetup/SdkDownloadService').download();
+//require('./src/core/sdksetup/SdkLicense').show();
```
carolnavarropereira: Please, revert this. It was done just for testing issues.

packages/vscode-extension/src/util/ExtensionUtil.ts
```diff
@@ -96,3 +96,7 @@ export const InteractiveAnswersValidator: {
 	validateFolderDoesNotExist(path: string): boolean;
 	validateFileName(fileName: string): boolean;
 } = require('@oracle/suitecloud-cli/src/validation/InteractiveAnswersValidator');
+
+export const APIKeyGenerator : {
+	generateApiKey(length?: number) : any;
```
carolnavarropereira: It should return a string instead of any

packages/node-cli/src/utils/ClientApiKeyUtils.js
TheJorgeBorras: This is already happening within this class, no?

Both if sdkExecutor encounters an error or Java returns an OperationResult with status === Error,
Both should provide an errorMessage here. Even if they are not, we are adding a message on top for better contextualization. 

Unfortunately, I don't see any case in which we return an empty error. I'm failing to see what your point is here.

packages/node-cli/src/commands/proxy/generatekey/ProxyGenerateKeyAction.js
TheJorgeBorras: I don't know what the coding standards say about this one, but I like it better when we reference a static method by referencing the class than when we reference the function directly.

In this case, I have wrapped the method within a class so we can also call it from VSCode and avoid duplication. Let me know if this is a dealbreaker.

TheJorgeBorras: Changed my mind on the matter. 

I noticed that for the rest of utils within NodeCLI, we usually declare static functions. What I was previously doing didn't feel right... We only declare and use static functions alongside the class within VSCode...

packages/node-cli/src/utils/ClientApiKeyDTO.js
```diff
@@ -0,0 +1,78 @@
...
+
+	fromRawString(stringifiedData) {
+		try {
+			this.data = !stringifiedData
+				? {}
+				: JSON.parse(stringifiedData);
+		} catch (error) {
+			throw NodeTranslationService.getMessage(CLIENT_API_KEY_UTILS.ERRORS.INVALID_FILE_CONTENTS);
+			// TODO: we do not want to show the content of the file
+			// Therefore, we dont want to show the contents of the caughtError (due to the risk of it sharing unwanted info)
```
TheJorgeBorras: Remove comment before merging PR

packages/node-cli/src/utils/ClientApiKeyUtils.js
carolnavarropereira: This is defensive coding. We should not assume everything will be all right on the other side. Anyway, I will close the comment.

packages/node-cli/src/commands/proxy/generatekey/ProxyGenerateKeyAction.js
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	async execute() {
+		try {
+			const newApiKey = generateAPIKey();
+			const existingFileContent = await readClientAPIKeyFileContents(this._sdkExecutor);
+			const clientApiKeyFileContents = ClientApiKeyDTO.Builder
+				.fromRawString(existingFileContent.data)
+				.withNewProxyKey(newApiKey)
+				.build();
```
davidecorreu: I like the idea of reading the file a test to know everything works as expected, like a validation to surface potential problems.
In this initial implementation we are trying to just overwrite the value and it doesn't seem right to try to build from previous value. In any case I would like to rethink a little bit about the current ClientApiKeyDTO.

packages/node-cli/src/metadata/CommandGenerators.json
```diff
@@ -78,5 +78,10 @@
 		"commandName": "object:create",
 		"supportsInteractiveMode": true,
 		"generator": "commands/object/create/CreateObjectCommand"
+	},
+	{
+		"commandName": "proxy:generatekey",
+		"supportsInteractiveMode": false,
+		"generator": "commands/proxy/generatekey/ProxyGenerateKeyCommand"
```
davidecorreu, line 80: I would prefer to keep alphabetical order on commandName.

packages/node-cli/src/commands/proxy/generatekey/ProxyGenerateKeyAction.js
TheJorgeBorras: Sure, we can look into ways of enchancing it and makint it future-proof. What do you suggest?

One thing to keep in mind, specifically about Proxy_Key we agreed that when generating a new key "previous APIs key won't be available" (see ticket). I see little point in storing APIs that cannot be used anymore.

packages/node-cli/src/metadata/CommandGenerators.json
TheJorgeBorras: Changes done. Please verify

packages/node-cli/src/utils/ClientApiKeyDTO.js
```diff
@@ -0,0 +1,78 @@
...
+ *		 	"PROXY_KEY": {
+ *		 	    "creationDate": "2026-04-13T12:46:52.577Z",
+ *		 	    "proxyApiKey": "3d3dfe3ec58eb9b9826f4546de20f96f99f6b7ed1fcdcaaa5f3c252eb69dce56"
+ *		 	}
+ *		}
+ *
+ *
+ */
+
+class ClientApiKeyDTO {
```
davidecorreu: I would like to review this class.

packages/node-cli/src/utils/ClientAPIKeyFileContent.js
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Parses the stringified contents of the client_api_key.p12 file into a ClientAPIKeyFileContent object.
+ *
+ * @param {string} stringifiedData a stringified JSON that follows the defined ClientAPIKeyFile schema.
+ * @returns {ClientAPIKeyFileContent}
+ */
+function parseClientApiKeyContent(stringifiedData) {
+	try {
+		const data = !stringifiedData
```
carolnavarropereira: Do not use negate condition. Just revert the ternary.

packages/node-cli/src/utils/ClientAPIKeyFileContent.js
```diff
@@ -0,0 +1,103 @@
...
+	 */
+	toCommandlineCompatibleString() {
+		return '"' + JSON.stringify(this.data).replaceAll('"', String.raw`\"`) + '"';
+	}
+}
+
+/**
+ * Parses the stringified contents of the client_api_key.p12 file into a ClientAPIKeyFileContent object.
+ *
+ * @param {string} stringifiedData a stringified JSON that follows the defined ClientAPIKeyFile schema.
```
carolnavarropereira: remove the a in "a stringfied"

packages/node-cli/src/utils/ClientAPIKeyFileContent.js
```diff
@@ -0,0 +1,103 @@
...
+
+		const defaultKeyName = this.data[FILE_FIELDS.DEFAULT_KEY];
+		this.data[FILE_FIELDS.KEYS][defaultKeyName] = {
+			[KEY_FIELDS.CREATION_DATE] : creationDate,
+			[KEY_FIELDS.VALUE] : proxyAPIKey,
+		};
+	}
+
+	/**
+	 * Returns the internal structure of a ClientApiKey Object that can be passed as a Commandline argument.
```
carolnavarropereira: Commandline does not need to be uppercase

packages/vscode-extension/src/commands/CreateDevAssistApiKey.ts
davidecorreu, line 1: I would like to avoid touching or modifying any vscode-extension files in this PR.
This goes for CreateDevAssistApiKey.ts, APIKeyGenerator.ts and ExtensionUtil.ts. 

Can we revert them to its original state and deal with changes if required in a specific PR?

packages/node-cli/src/commands/proxy/generatekey/ProxyGenerateKeyAction.js
TheJorgeBorras: class is updated. Feel free to open any new threads if we need to discuss it further

packages/node-cli/src/utils/ClientApiKeyDTO.js
TheJorgeBorras: class has been updated. 
Feel free to open any new threads if we need to discuss it further :+1:

packages/vscode-extension/src/commands/CreateDevAssistApiKey.ts
TheJorgeBorras: Done.
THough I've added a comment as a reminder to ourselves to remove the duplicated code later on.

davidecorreu (approved):

packages/node-cli/src/metadata/SdkCommandsMetadata.json
```diff
@@ -784,5 +784,13 @@
 				"disableInIntegrationMode": false
 			}
 		}
+	},
+	"proxy:generatekey": {
+		"name": "proxy:generatekey",
+		"sdkCommand": "generatekey",
+		"usage": "",
+		"description": "Generates a new key for proxy authentication.",
```
davidecorreu: Can you check this against:
`"COMMAND_PROXY_GENERATEKEY_ERROR_UNABLE_TO_GENERATE_KEY": "Unable to generate a new Proxy API key."`

davidecorreu (changes requested):

packages/node-cli/messages.json
```diff
@@ -307,7 +310,11 @@
 	"UTILS_AUTHENTICATION_SUCCESS_SETUP": "The account has been successfully set up.",
 	"UTILS_AUTHENTICATION_SAVING_STARTING_OAUTH_FLOW": "Continue the authentication process in the prompted browser.",
 	"UTILS_ERRORS_COMMAND": "There was an error with the {0} command. {1}",
+	"UTILS_CLIENT_API_KEY_UTILS_ERRORS_INVALID_FILE_CONTENTS": "The Client API Key file contents are not in a valid format. Delete the \"client_api_key.p12\" file and try again.",
```
davidecorreu: Ideally this should have the absolute path to the file.

davidecorreu (approved):

carolnavarropereira (approved):

PR #1016, opened 2026-04-28
feature/pdpdevtool6392: CLI proxy for SuiteCloud Developer Assistant

davidecorreu:


packages/node-cli/messages.json
```diff
@@ -179,7 +179,28 @@
...
+	"COMMAND_PROXY_START_MESSAGES_STOPPED": "Proxy for SuiteCloud Developer Assistant service stopped.",
+	"COMMAND_PROXY_START_MESSAGES_STOP_INSTRUCTIONS": "Press Ctrl+C to stop the proxy.",
+	"COMMAND_PROXY_START_MESSAGES_STOPPING": "Stopping proxy. Waiting for active requests to finish...",
+	"COMMAND_PROXY_START_MESSAGES_STARTED": "Starting proxy for SuiteCloud Developer Assistant service...",
+	"COMMAND_PROXY_START_ERRORS_PORT_MUST_BE_NUMBER": "The port must be a valid number between {0} and {1}.",
+	"COMMAND_PROXY_START_ERRORS_PORT_OUT_OF_RANGE": "The port must be a valid number between {0} and {1}.",
+	"COMMAND_PROXY_START_ERRORS_PORT_ALREADY_IN_USE": "Port {0} is already in use.",
+	"COMMAND_PROXY_START_ERRORS_READING_CLIENT_API_CONTENTS": "Error reading Client API contents.",
+	"COMMAND_PROXY_START_ERRORS_INVALID_PROXY_KEY_PAYLOAD": "Could not parse proxy key payload from SDK output.",
+	"COMMAND_PROXY_START_ERRORS_MISSING_API_KEY": "No client API key was found. Run \"suitecloud proxy:generatekey\" and try again.",
```
TheJorgeBorras: The `'c'` in `'client API key'` should be uppercase to be consistent with the rest of messages we have in Java and Javascript.

Final message: `"No Client API key was found. Run \"suitecloud proxy:generatekey\" and try again."`

packages/node-cli/src/utils/ClientAPIKeyUtils.js
```diff
@@ -72,11 +73,33 @@ async function writeClientAPIKeyFileContents(sdkExecutor, newFileContent) {
...
+ * @returns {Promise<{apiKey: string}>} An object containing the resolved API key.
+ * @throws {string|Array<string>} Throws translated/SDK errors when the key file cannot be read or parsed.
+ */
+async function resolveDefaultClientApiKey(sdkExecutor) {
+	const readOperationResult = await readClientAPIKeyFileContents(sdkExecutor);
+	const clientApiKeyObjectWrapper = new ClientAPIKeyObjectWrapper(readOperationResult.data);
+	const apiKey = clientApiKeyObjectWrapper.getDefaultKeyValue();
+
+	if (typeof apiKey !== 'string' || !apiKey.trim()) {
+		throw NodeTranslationService.getMessage(COMMAND_PROXY_START.ERRORS.MISSING_API_KEY);
```
TheJorgeBorras: I feel like the key for this message doesn't fit the class. This is a service consumed by different commands. As such I propose:

1.  `COMMAND_PROXY_START.ERRORS.MISSING_API_KEY` =>  `CLIENT_API_KEY_UTILS.ERRORS.MISSING_API_KEY`
2. move this validation outside this class (to the code calling this method)


 (I prefer option 1).

davidecorreu: I agree with your input here. I'll go for option 1 in the next commit.

Montsant19: LGTM

packages/node-cli/messages.json
Montsant19: We can remove Client.

packages/node-cli/src/services/TranslationKeys.js
```diff
@@ -265,13 +265,43 @@ module.exports = {
 		PACKAGING: 'COMMAND_PACKAGE_PACKAGING',
 	},
 
-	COMMAND_PROXY_GENERATEKEY: {
+		COMMAND_PROXY_GENERATEKEY: {
```
TheJorgeBorras: typo? (unnecessary indentation here)

packages/node-cli/src/commands/proxy/start/ProxyStartInputHandler.js
```diff
@@ -0,0 +1,97 @@
...
+			const accountInfo = `${credentials.accountInfo.companyName} [${credentials.accountInfo.roleName}]`;
+			choices.push({
+				name: NodeTranslationService.getMessage(COMMAND_SETUPACCOUNT.QUESTIONS_CHOICES.SELECT_AUTHID.EXISTING_AUTH_ID, authId, accountInfo, ''),
+				value: authId,
+			});
+		});
+
+		choices.push(new Separator());
+		return choices;
+	}
```
TheJorgeBorras, line 76: This is very much the same as `SetupInputHandler.js`, lines `103 to 114`.

Duplicated code is not the end of the world, so we can leave it like this. 
But still, I wanted to let you know just in case 😉

packages/node-cli/src/commands/proxy/start/ProxyStartInputHandler.js
```diff
@@ -0,0 +1,97 @@
...
+		if (port < PORT_RANGE.MIN || port > PORT_RANGE.MAX) {
+			return this._getInvalidPortMessage();
+		}
+
+		return true;
+	}
+
+	_getInvalidPortMessage() {
+		return NodeTranslationService.getMessage(COMMAND_PROXY_START.ERRORS.PORT_MUST_BE_NUMBER, PORT_RANGE.MIN, PORT_RANGE.MAX);
+	}
```
TheJorgeBorras: This could be a single function.
```suggestion
	_validatePortInput(value) {
		const port = Number(value);
		if (Number.isNaN(port) || port < PORT_RANGE.MIN || port > PORT_RANGE.MAX) {
			return NodeTranslationService.getMessage(COMMAND_PROXY_START.ERRORS.PORT_MUST_BE_NUMBER, PORT_RANGE.MIN, PORT_RANGE.MAX);
		}

		return true;
	}
```

packages/node-cli/messages.json
davidecorreu: Message reviewed by Montsant.
`No API key was found. Run "suitecloud proxy:generatekey" and try again.`
Fixed in next commit.

packages/node-cli/src/commands/proxy/start/ProxyStartInputHandler.js
davidecorreu: There is some subtle difference that I let you find making harder to share the code.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,171 @@
...
+	OPTIONS: {
+		AUTH_ID: 'authid',
+		PORT: 'port',
+	},
+};
+
+const PORT_RANGE = {
+	MIN: 1024,
+	MAX: 65535,
+};
```
TheJorgeBorras: This code is duplicated.
They are declared both here (`ProxyStartAction.js`) and in ProxyStartInputHandler.js`.

In my opinion, we should only declare it once.

I see two options here: 

- Declare it once and export it to the other class.
- Perhaps we should follow the same approach as `AccountSetupCiValidation.js` and create a class for ProxyStartConstants. I'm leaning towards this one.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		if (!this._validatePortAvailability(port)) {
+			throw NodeTranslationService.getMessage(COMMAND_PROXY_START.ERRORS.PORT_ALREADY_IN_USE, port);
+		}
+	}
+
+	_getInvalidPortMessage() {
+		return NodeTranslationService.getMessage(COMMAND_PROXY_START.ERRORS.PORT_MUST_BE_NUMBER, PORT_RANGE.MIN, PORT_RANGE.MAX);
+	}
```
TheJorgeBorras: This code too exists in two different classes: here (`ProxyStartAction.js`) and in ProxyStartInputHandler.js`.

I think we should go for a `ProxyStartValidation.js` class with the Command Constants and the methods used in multiple classes (same approach as `AccountSetupCiValidation.js`).

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,171 @@
...
+			if (this._isShuttingDown) {
+				return;
+			}
+
+			this._isShuttingDown = true;
+
+			try {
+				if (this._proxyService) {
+					// Ensure shutdown feedback starts on a new line instead of being appended to terminal '^C' echo.
+					process.stdout.write('\n');
```
TheJorgeBorras: Question: why `process.stdout.write('\n');` instead of the usual `this._log.info('');`?
Because '^C' doesn't work well with it? or is there another reason?

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -464,6 +474,17 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
+	_isDebugEnabled() {
+		return process.env[DEBUG_ENV_VAR] === 'true';
+	}
+
+	_debug(...args) {
+		if (this._isDebugEnabled()) {
+			console.debug('[SuiteCloudAuthProxyService]');
+			args.forEach((arg) => console.debug(arg));
+		}
+	}
```
TheJorgeBorras, line 486: This fits our purposes well, but it pains my soul a little bit also in the process.

I wouldn't create a single-line single-use function for just a simple boolean condition. Instead, I suggest we go for this (utilizing the name of the class constant):

```
	_debug(...args) {
		if (process.env[DEBUG_IS_ENABLED] === 'true') {
			console.debug('[SuiteCloudAuthProxyService]');
			args.forEach((arg) => console.debug(arg));
		}
	}
```

`DEBUG_IS_ENABLED`, `DEBUG_ENABLED`, `DEBUG_ENABLED_ENV_VAR`, etc.
Feel free to ignore this comment.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,171 @@
...
+	async _handleManualAuthRefreshRequired({ message, authId }) {
+		if (this._manualRefreshInFlight) {
+			await this._manualRefreshInFlight;
+			return;
+		}
+
+		this._manualRefreshInFlight = (async () => {
+			this._log.error(message);
+			await this._log.info(NodeTranslationService.getMessage(COMMAND_REFRESH_AUTHORIZATION.MESSAGES.CREDENTIALS_NEED_TO_BE_REFRESHED, authId));
+			const refreshAuthzOperationResult = await refreshAuthorization(authId, this._sdkPath, this._executionEnvironmentContext);
```
TheJorgeBorras, line 138: typo => `refreshAuthOperationResult`

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,171 @@
...
+			this._proxyService = new SuiteCloudAuthProxyService(this._sdkPath, this._executionEnvironmentContext, ALLOWED_PROXY_PATH_PREFIX, apiKey);
+			this._registerProxyEvents(authId, port);
+			this._registerShutdownHandlers();
+
+			await this._proxyService.start(authId, port);
+
+			return ActionResult.Builder.withData({ authId, port }).build();
+		} catch (error) {
+			return ActionResult.Builder.withErrors([error]).build();
+		}
```
TheJorgeBorras, line 77: If I'm not mistaken, `readClientAPIKeyFileContents(sdkExecutor)` within `resolveDefaultClientApiKey(this._sdkExecutor)` can throw an array of Error messages.

This is why on `ProxyGenerateKeyAction.js` we check if the caught object is an array of errors or not to avoid creating multiple nested arrays.
```
// ProxyGenerateKeyAction.js line 38
...
		} catch (error) {
			const errorMessages = Array.isArray(error) ? error : [error];
			return ActionResult.Builder.withErrors([
				NodeTranslationService.getMessage(UNABLE_TO_GENERATE_KEY),
				...errorMessages,
			]).build();
		}
```

We should do something similar here, right?

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,171 @@
...
+			const port = params[COMMAND.OPTIONS.PORT];
+			const apiKey = params.apiKey || (await resolveDefaultClientApiKey(this._sdkExecutor)).apiKey;
+
+			this._proxyService = new SuiteCloudAuthProxyService(this._sdkPath, this._executionEnvironmentContext, ALLOWED_PROXY_PATH_PREFIX, apiKey);
+			this._registerProxyEvents(authId, port);
+			this._registerShutdownHandlers();
+
+			await this._proxyService.start(authId, port);
+
+			return ActionResult.Builder.withData({ authId, port }).build();
```
TheJorgeBorras: Codex makes a fair point here, which I was also able to verify locally.

This code returns a `SUCCESS` ActionResult if the `this._proxyService.start(authId, port);` runs without encountering errors. However, it doesn't wait for the Listening callback before showing this message.

Therefore, we can encounter situations in which we show  a `Starting proxy for SuiteCloud Developer Assistant service...` followed immediately by an error message. I don't think it's a big deal, it doesn't look ugly but maybe we should discuss it among the team.

Image

packages/node-cli/src/services/TranslationKeys.js
davidecorreu: I'll remove extra indentation in next commit.

packages/node-cli/src/commands/proxy/start/ProxyStartInputHandler.js
davidecorreu: Validation has been changed and moved to `ProxyStartValidation.js`.
You can check it next commits.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
davidecorreu: Moved constants to `ProxyStartConstants.js` in next commits.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
davidecorreu: The validation has been move to `ProxyStartValidation.js` in next commits.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
TheJorgeBorras: never mind, the 'z' comes from 'AuthoriZation', right?

This change is not needed

packages/node-cli/src/commands/proxy/start/ProxyStartInputHandler.js
```diff
@@ -0,0 +1,85 @@
...
+		return choices;
+	}
+
+	_validatePortInput(value) {
+		if (!isValidProxyStartPort(value)) {
+			return getInvalidProxyStartPortMessage();
+		}
+
+		return true;
+	}
```
TheJorgeBorras, line 84: We moved most of the duplicated logic/validations to `ProxyStartValidation.js`.

Still, this function remains duplicated both here and in `ProxyStartAction.js` ... But this is intentional, right? For readability purposes?

If this is the case, feel free to close the thread.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
davidecorreu: Agreed to not do.

packages/node-cli/src/commands/proxy/start/ProxyStartInputHandler.js
TheJorgeBorras: Code is similar, but not the same.
No duplication here :+1: :+1: 
(closing thread)

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	_markProxyReadyIfPending() {
+		if (this._proxyReadiness.settled) {
+			return;
+		}
+
+		this._proxyReadiness.settled = true;
+		this._proxyReadiness.resolve();
+	}
```
TheJorgeBorras: Just for visibility's sake, we are still discussing these functions.
Functionality-wise we believe it's good, but it's a bit obscure and could be more readable.

Renaming the functions to `resolveProxyReadynessOnce` && `rejectProxyReadynessOnce`, invert the if statement to reduce clutter.

or go for remove them altogether (since they are one time use anyways):

```suggestion
	_handleServerListening() {
		// trigger SpinnerExecute Stop
		if (!this._proxyReadiness.settled) {
			this._proxyReadiness.settled = true;
			this._proxyReadiness.resolve();
		}
	}
	
	_handleProxyErrorDefault({ message }) {
		// trigger SpinnerExecute stop
		if (!this._proxyReadiness.settled) {
			this._proxyReadiness.settled = true;
			this._proxyReadiness.reject(message);
			return;
		}

		this._log.error(message);
	}
```

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,205 @@
...
+				process.exit(0);
+			} catch (error) {
+				this._log.error(error);
+				process.exit(1);
+			}
+		};
+
+		process.on('SIGINT', shutdown); 	// Ctrl+C on Linux, macOS, and Windows terminals.
+		process.on('SIGTERM', shutdown); 	// Termination requests from the OS, shell, or process managers.
+		process.on('SIGBREAK', shutdown); 	// Ctrl+Break on Windows terminals.
```
alisyed-19, line 194: Process in Node is an EventEmitter.
Every process.on('SIGINT', handler) call adds a new handler; it does not replace existing ones.

execute() calls _registerShutdownHandlers() on each run.
_registerShutdownHandlers() always does 3 process.on(...) calls:

So if the command is started multiple times in the same Node process (common in IDE/plugin hosts or tests), listeners stack.
One SIGINT then triggers all of them, causing duplicate shutdown behavior/logging, and eventually MaxListenersExceededWarning once enough accumulate.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
davidecorreu: Replaced for `this._log.info('');`

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
davidecorreu: Fixed this and some other related problems to error creation from `ClientAPIKeyUtils.js`.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
davidecorreu: Reworked to complete ActionResult.SUCCESS only after underlying proxy has emitted LISTEN event.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
davidecorreu: Removed dedicated resolve/reject functions for its inline code.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
davidecorreu: When this is "technically" true, is not going to happen when using our current implementation on suitecloud CLI as we are not creating an instance of `ProxyStartAction` and calling multiple times `ProxyStartAction.execute`.
Each command execution of any of the CLI command its own Node process that starts and ends.
We could start many "sutiecloud proxy:start" commands in different terminals and each one of them is going to be a separate Node process.
I would prefer to leave this as it is.

packages/node-cli/messages.json
```diff
@@ -179,7 +179,27 @@
...
+	"COMMAND_PROXY_START_MESSAGES_GUIDE_MODEL_ID": "  * Model ID: NetSuite",
+	"COMMAND_PROXY_START_MESSAGES_LISTENING_ON": "SuiteCloud Proxy server listening on {0}",
+	"COMMAND_PROXY_START_MESSAGES_READING_CLIENT_API_CONTENTS": "Reading the Client API Key file contents...",
+	"COMMAND_PROXY_START_MESSAGES_RUNNING_WITH_AUTH_ID": "SuiteCloud Developer Assistant service is now running on port {1} and using the auth ID {0}.",
+	"COMMAND_PROXY_START_MESSAGES_STOPPED": "Proxy for SuiteCloud Developer Assistant service stopped.",
+	"COMMAND_PROXY_START_MESSAGES_STOP_INSTRUCTIONS": "Press Ctrl+C to stop the proxy.",
+	"COMMAND_PROXY_START_MESSAGES_STOPPING": "Stopping proxy. Waiting for active requests to finish...",
+	"COMMAND_PROXY_START_MESSAGES_STARTED": "Starting proxy for SuiteCloud Developer Assistant service...",
+	"COMMAND_PROXY_START_ERRORS_PORT_MUST_BE_NUMBER": "The port must be a valid number between {0} and {1}.",
+	"COMMAND_PROXY_START_ERRORS_PORT_OUT_OF_RANGE": "The port must be a valid number between {0} and {1}.",
```
TheJorgeBorras: Same message declared twice.
We should only have one, right?

packages/node-cli/messages.json
```diff
@@ -179,7 +179,27 @@
...
+	"COMMAND_PROXY_START_ERRORS_PORT_ALREADY_IN_USE": "Port {0} is already in use.",
+	"COMMAND_PROXY_START_ERRORS_READING_CLIENT_API_CONTENTS": "Error reading Client API contents.",
```
TheJorgeBorras: This one doesn't look quite right to me. Do we need it? Maybe we can reause this one that already exists (validated with TW): 

```
"UTILS_CLIENT_API_KEY_UTILS_ERRORS_READING_FILE_CONTENTS": "Error reading the Client API Key file contents."
```

packages/node-cli/src/services/TranslationKeys.js
```diff
@@ -272,6 +272,35 @@ module.exports = {
 		SUCCESS_KEY_GENERATED: 'COMMAND_PROXY_GENERATEKEY_SUCCESS_KEY_GENERATED',
 	},
 
+	COMMAND_PROXY_START: {
+		QUESTIONS: {
+			SELECT_AUTH_ID: 'COMMAND_PROXY_START_QUESTIONS_SELECT_AUTH_ID',
+			PORT: 'COMMAND_PROXY_START_QUESTIONS_PORT',
+		},
+		MESSAGES: {
+			CHECKING_SERVER: 'COMMAND_PROXY_START_MESSAGES_CHECKING_SERVER',
```
TheJorgeBorras: This message doesn't have an equivalent on messages.json

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
```diff
@@ -0,0 +1,196 @@
+/*
+ ** Copyright (c) 2024 Oracle and/or its affiliates.  All rights reserved.
```
TheJorgeBorras: Hmmmmmm... copyright should be 2026 for new files, no?

🫤

packages/node-cli/src/services/TranslationKeys.js
davidecorreu: removing it in next commit.

packages/node-cli/src/commands/proxy/start/ProxyStartAction.js
davidecorreu: Fixed in next commit.

TheJorgeBorras (approved):

alisyed-19 (approved):

PR #1032, opened 2026-05-14
Bugfix/pdpdevtool 6490 cli proxy does not work behind oracle vpn

carolnavarropereira:


packages/node-cli/src/services/ProxyService.js
```diff
@@ -0,0 +1,77 @@
...
+			scheduling: 'lifo',
+			proxy: proxyURL,
+		};
+
+		return this.createHttpsProxyAgent(proxyAgentOptions);
+	}
+
+	createHttpsProxyAgent(options) {
+		return new HttpsProxyAgent(options);
+	}
```
TheJorgeBorras: Do we truly need 3 levels of single-use nested functions.
In my opinion, this class would be simplier and easier to follow if we merge them.


```suggestion
getProxyAgent() {
	const proxyURL = this.getConfiguredProxy();
	if (!proxyURL) return undefined;

	return HttpsProxyAgent({
		keepAlive: true,
		keepAliveMsecs: 1000,
		maxSockets: 256,
		maxFreeSockets: 256,
		scheduling: 'lifo',
		proxy: proxyURL,
	});
}
```

packages/node-cli/src/services/ProxyService.js
```diff
@@ -0,0 +1,77 @@
...
+
+
+	getConfiguredProxy() {
```
TheJorgeBorras: If possible, `getConfiguredProxyURL()` is a more self explicative name.
Can we rename it?

packages/node-cli/src/services/ProxyService.js
```diff
@@ -0,0 +1,77 @@
...
+
+		if (!SUPPORTED_PROXY_PROTOCOLS.has(proxyURL.protocol)) {
+			throw NodeTranslationService.getMessage(
+				SUITECLOUD_AUTH_PROXY_SERVICE.UNSUPPORTED_PROXY_PROTOCOL,
+				ENV_VARS.SUITECLOUD_PROXY,
+			);
+		}
+
+		return proxyURL;
+	}
```
TheJorgeBorras: What do we do in the Java Layer if the ProxyURL is invalid? Do we also throw error messages?

I was under the impression that we ignore bad values and proceed as if the user didn't specify any Proxy values... Is this assumption wrong? Maybe we should do the same here.


Note: I'm not saying this code is wrong, I just want to double check. I'm not sure if we should add this code.

packages/node-cli/src/services/ProxyService.js
```diff
@@ -0,0 +1,77 @@
...
+
+	getConfiguredProxyValue() {
+		return process.env[ENV_VARS.SUITECLOUD_PROXY];
+	}
```
TheJorgeBorras: Personally, I wouldn't go for a single use single line nested function. The code feels simple enough for this specific case. Just add the code in the parent function directly and add a comment if you feel like it.

But if we want to go for a nested function, that's also fine.
 In that case, I would recommend that we gave it a more appropriate name: `getSuiteCloudProxyValue()` or `getSuiteCloudProxyValueFromEnvVariables()` or `getSuiteCloudProxyValueFromExecutionContext()` or `getConfiguredSuiteCloudProxyValue()`.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -321,10 +328,13 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
-		}		
+		}
 
-		// Add agent for insecure connections when connecting to runboxes
-		if (this._targetHost && this._targetHost.includes('vm.eng')) {
+		if (this._isProductionUrl(this._targetHost)) {
+			//Add proxy agent for production in order to work properly with vpn
+			requestOptions.agent = this._proxyService.getProxyAgent();
+		} else {
+			//Add agent for insecure connections when connecting to runboxes
```
TheJorgeBorras, line 332: I would refrain from adding comments with internal data or confidential information.

```suggestion
			// Add a proxy agent to support restricted network environments (e.g: VPN)
			requestOptions.agent = this._proxyService.getProxyAgent();
		} else {
			// Add an agent to support relaxed TLS settings (development environments)
```

packages/node-cli/src/services/ProxyService.js
carolnavarropereira: I am going to remove this code because I do not think provides any value

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: This message was already there.

packages/node-cli/src/services/ProxyService.js
carolnavarropereira: I left just an error message saying the the suitecloud_proxy is wrong

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
TheJorgeBorras: Yes. But, still, my point remains and I would use this opportunity to fix it now

packages/node-cli/src/utils/UriUtils.js
```diff
@@ -0,0 +1,53 @@
+/*
+ ** Copyright (c) 2024 Oracle and/or its affiliates.  All rights reserved.
+ ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+ */
```
TheJorgeBorras, line 4: Copyright notice should point to `2026`,  `2024` is not right.

(We should probably find the culprit config_file and update that one, but this is very much outside the scope so obviously I won't enforce it).

packages/node-cli/src/services/ProxyAgentService.js
```diff
@@ -0,0 +1,171 @@
+/*
+ ** Copyright (c) 2022 Oracle and/or its affiliates.  All rights reserved.
+ ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+ */
```
TheJorgeBorras, line 4: Copyright notice should point to `2026`,  `2022` is not right.

(We should probably find the culprit config_file and update that one, but this is very much outside the scope so obviously I won't enforce it).

packages/node-cli/src/ApplicationConstants.js
```diff
@@ -68,12 +68,19 @@ module.exports = {
...
 	},
 	HTTP_RESPONSE_CODE: {
 		UNAUTHORIZED: 401,
 		FORBIDDEN: 403,
 		INTERNAL_SERVER_ERROR: 500,
 		SERVICE_UNAVAILABLE: 503,
 	},
+	SUITECLOUD_PROXY_SERVICE: {
+		SUPPORTED_PROTOCOLS: ['http:', 'https:'],
+		REGEX_SYSTEM_URL : /^system\.netsuite\.com$/,
```
TheJorgeBorras, line 82: I'm not sure if you've noticed, but this REGEX is more restrictive that the one we've been using.
See  `ApplicationConstants.js`, `line 44`:

```
	DOMAIN: {
		PRODUCTION: {
			GENERIC_NETSUITE_DOMAIN: 'system.netsuite.com',
			PRODUCTION_DOMAIN_REGEX: '^system\\.(\\w+\\.)*netsuite\\.com$',
		...
		}
}
```

So the old Regex would allow `system.na1.netsuite.com` but the new one wouldn't...
We should verify which is the one we want to use and then use the same value, right?

Probably David, Antonio or Ali can provide more context here. Let me know what do you think

carolnavarropereira:
we have replaced by a different pull request with an different version of the code.

PR #1036, opened 2026-05-20
cli proxy does not work behind oracle vpn sdk download

carolnavarropereira:


packages/node-cli/src/services/ProxyAgentService.js
```diff
@@ -0,0 +1,53 @@
...
+	PROXY_AGENT_SERVICE,
+} = require('./TranslationKeys');
+
+const {
+	ENV_VARS: {
+		SUITECLOUD_PROXY,
+	},
+} = require('../ApplicationConstants');
+
+class ProxyAgentService {
```
davidecorreu: Both ProxyAgentService and UriUtils are stateless static-method bags. In JS that usually reads cleaner as plain exported functions, because there’s no instance state, lifecycle, inheritance, or polymorphism being used.
I would refactor the code to just export the required functions.

packages/node-cli/src/services/ProxyAgentService.js
```diff
@@ -0,0 +1,53 @@
...
+		} catch (error) {
+			const proxyError = new Error(
+				NodeTranslationService.getMessage(
+					PROXY_AGENT_SERVICE.INVALID_PROXY_CONFIGURATION,
+					SUITECLOUD_PROXY,
+					configuredProxyUri,
+				),
+			);
+			proxyError.code = error.code;
+			throw proxyError;
```
davidecorreu: We don't precisely know from where the proxy value being evaluated comes from. This could be a misleading message if value was not resolved from  SUITECLOUD_PROXY.

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
```diff
@@ -100,7 +102,13 @@ class SuiteCloudAuthProxyService extends EventEmitter {
...
 			}
 
-			const requestOptions = this._buildRequestOptions(request);
+			let requestOptions;
+			try {
+				requestOptions = this._buildRequestOptions(request);
+			} catch (error) {
+				this._localProxy.emit('error', error);
+				return;
+			}
```
davidecorreu: I would prefer to a more defensive approach to check proxy value (currently done inside _buildRequestOptions) . As we know the _targetHost in advance I would validate the proxy value if it's a production url, even before starting the server to fail fast and not wait until a request is received.

We are also missing to send an unsuccessful response here (consumer will be waiting for it), but that won't be necessary if we go for the more defensive proposal.

packages/node-cli/src/utils/UriUtils.js
```diff
@@ -0,0 +1,33 @@
...
+
+const {
+	SUITECLOUD_PROXY_SERVICE: {
+		REGEX_SYSTEM_URL,
+		REGEX_ACCOUNT_SPECIFIC_URL,
+		REGEX_SUITETALK_API_PRODUCTION_URL,
+	}, ENV_VARS,
+} = require('../ApplicationConstants');
+
+class UriUtils {
```
davidecorreu: Some problems I see here:

As said in a previous comment this not need to be a class.

- There is a typo on sProductionDomain (missing an "i")
- We are mixing URL things with retrieving environment variables. We should separate that logic to a different util/service file.
- The names for environment variables functions could be improved. Some proposal could be `getSuiteCloudProxyEnvVariable` and `resolveProxyValueFromEnviroment`. The last one could be returning the value and the source (environment variable name) but that could be something to discus further.

packages/node-cli/src/services/ProxyAgentService.js
davidecorreu: We should also probably validate for appropriate protocol (`http`/`https`) as we `file` is valid for URL but we shouldn't allow it. Maybe we are also interested in check the port.

davidecorreu: This is our current message from java-cli: `The SUITECLOUD_PROXY environment variable must contain a protocol, a hostname, and a port. For example, http://my-proxy-domain:80.`

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
alisyed-19: If part 1 (fail-fast validation in start()) is in place, it covers the specific bad-host/proxy scenario David called out.
But part 2 is still valuable as safety hardening: you still have async request floww in `SuiteCloudAuthProxyService.js` (especially request.on('end', async ...)) where unexpected errors can occur and leave client behavior unclear unless you always write an error response.

What do you think @davidecorreu ?

packages/node-cli/src/utils/UriUtils.js
davidecorreu: Please refactor to functional style.

packages/node-cli/src/services/ProxyAgentService.js
carolnavarropereira: Done it

packages/node-cli/src/services/SuiteCloudAuthProxyService.js
carolnavarropereira: Done

packages/node-cli/src/utils/UriUtils.js
carolnavarropereira: Done

packages/node-cli/src/services/proxy/ProxyEnvironmentUtils.js
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * validates the Uri from the Proxy is correct.
+ *
+ * @param {{ proxyUri: string, envVarName: string }} configuredProxy Proxy configuration to validate. envVarName
+ * is the environment variable where the value is retrieved. proxyUri is the value itself.
+ * @throws {Error} When `proxyUri` is invalid, uses an unsupported protocol, or does not include a valid port.
+ * The thrown error message includes the source environment variable name and configured value.
+ */
+function validateUriProxy(configuredProxy) {
```
davidecorreu: Please rename this to `validateProxyUri`.

packages/node-cli/messages.json
```diff
@@ -314,6 +314,9 @@
 	"SUITECLOUD_AUTH_PROXY_SERVICE_REQUEST_PATH_NOT_ALLOWED_ERROR" : "Invalid request. The request path must start with {0}.",
 	"SUITECLOUD_AUTH_PROXY_SERVICE_UNAUTHORIZED_PROXY_REQUEST": "Unauthorized: The API key is missing or invalid.",
 
+	"PROXY_AGENT_SERVICE_INVALID_PROXY_CONFIGURATION": "The {0} environment variable must contain a protocol, a hostname, and a port. For example, http://my-proxy-domain:80.\nReceived: {1}",
+	"PROXY_AGENT_SERVICE_UNSUPPORTED_PROXY_PROTOCOL": "You specified an unsupported protocol for the {0} environment variable. Enter a proxy URL with either the http or https protocol.\nReceived: {1}",
+
```
davidecorreu: Let's add some context for @Montsant19 to review this one.

- This is our current message from java-cli: `The SUITECLOUD_PROXY environment variable must contain a protocol, a hostname, and a port. For example, http://my-proxy-domain:80.`
- Here in javascript code it could happening that error is thrown when reading the proxy value from one of the following environment variables: `SUITECLOUD_PROXY` or `npm_config_https_proxy` or `npm_config_https_proxy`. In java-cli it only can be `SUITECLOUD_PROXY`.
- The `{0}` placeholder is for the name of the environment variable
- The `{1}` placeholder is for the actual value from the environment variable.

Let's us know if the messages look good to you.

davidecorreu (approved):

alisyed-19 (approved):

PR #1037, opened 2026-05-26
Feature ts migration devbase

alisyed-19:


packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -141,9 +151,9 @@ module.exports = class DeployAction extends (
...
 
 			const isServerValidation = !!sdkParams[COMMAND.FLAGS.VALIDATE];
-			const isApplyInstallationPreferences = this._projectType === PROJECT_SUITEAPP && flags.includes(COMMAND.FLAGS.APPLY_INSTALLATION_PREFERENCES);
+			const isApplyInstallationPreferences = isApplyInstallationPreferencesForDeploy(this._projectType, flags, PROJECT_SUITEAPP);
 
-			return operationResult.status === SdkOperationResultUtils.STATUS.SUCCESS
+			return operationResult.status === SDK_OPERATION_STATUS.SUCCESS
 				? DeployActionResult.Builder.withData(operationResult.data)
 					.withResultMessage(operationResult.resultMessage)
 					.withServerValidation(isServerValidation)
```
carolnavarropereira: withServerValidation is not DeployActionResult. Implement it.

packages/sdk-core/src/commands/project/package/PackageProjectExecutor.ts
```diff
@@ -0,0 +1,131 @@
...
+			return errorResult('A project folder is required for project packaging.');
+		}
+		if (!input.destinationFolder) {
+			return errorResult('A destination folder is required for project packaging.');
+		}
+
+		await mkdir(input.destinationFolder, { recursive: true });
+		const targetZipFilePath = await getTargetZipFilePath(input.projectFolder, input.destinationFolder);
+		await execFileAsync(
+			ZIP_BINARY_NAME,
```
carolnavarropereira: Do not use external commands if it can be prevented.
Node has already support for zip files.
zip software must be installed in windows, it is not provided by default.

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -156,7 +166,71 @@ module.exports = class DeployAction extends (
...
+				params: sdkParams,
+				flags,
+				summaryContext: this._buildSummaryContext(authCredentials),
+			}),
+			message,
+		});
+	}
+
+	_buildSummaryContext(authCredentials) {
+		const accountInfo = authCredentials && authCredentials.accountInfo ? authCredentials.accountInfo : {};
```
carolnavarropereira, line 223: Maybe we can use an optional chain expression like;
const accountInfo = authCredentials?.accountInfo ? authCredentials.accountInfo : {};

packages/node-cli/src/commands/project/validate/ValidateAction.js
```diff
@@ -97,7 +104,71 @@ module.exports = class ValidateAction extends BaseAction {
...
+				flags,
+				summaryContext: this._buildSummaryContext(authCredentials),
+			}),
+			message,
+		});
+	}
+
+	_buildSummaryContext(authCredentials) {
+		const accountInfo = authCredentials && authCredentials.accountInfo ? authCredentials.accountInfo : {};
+		return {
```
carolnavarropereira, line 169: Maybe it can be used an optional chain expression
const accountInfo = authCredentials?.accountInfo ? authCredentials.accountInfo : {};

packages/sdk-core/src/commands/project/create/CreateProjectExecutor.ts
```diff
@@ -0,0 +1,348 @@
...
+    <!-- CUSTOM LISTS -->
+    <!-- The configuration below overwrites all customvalues of customlist_example except for the ones with the action tag PRESERVE. -->
+    <!-- <customlists>
+        <customlist scriptid="customlist_example">
+            <customvalues defaultAction="OVERWRITE">
+                <customvalue scriptid="customlist_value_example" action="PRESERVE" />
+            </customvalues>
+        </customlist>
+    </customlists> -->
+</preference>`;
```
TheJorgeBorras: I think we should create a template folder and store these as separate template files. Something similar to what we already have on the `node-cli` package.

Let me know your opinion.

```
packages/sdk-core/src/templates
├─ controlfiles
│   ├─ acp.manifest.xml
│   ├─ acp.deploy.xml
│   ├─ suiteapp.manifest.xml
│   └─ suiteapp.deploy.xml
├─ installationPreferences
│    ├─ hiding.xml
│    ├─ locking.xml
│    └─ overwritting.xml
└─ TemplateKeys.ts
```

packages/sdk-core/src/commands/project/create/CreateProjectExecutor.ts
```diff
@@ -0,0 +1,348 @@
...
+const FILE_CABINET_TEMPLATES_FOLDER = 'Templates';
+const FILE_CABINET_TEMPLATES_EMAIL_TEMPLATES_FOLDER = 'E-mail Templates';
+const FILE_CABINET_TEMPLATES_MARKETING_TEMPLATES_FOLDER = 'Marketing Templates';
+const FILE_CABINET_SUITE_APPS_FOLDER = 'SuiteApps';
+const ACCOUNT_CONFIGURATION_FOLDER = 'AccountConfiguration';
+const OBJECTS_FOLDER = 'Objects';
+const TRANSLATIONS_FOLDER = 'Translations';
+const HIDING_XML_FILENAME = 'hiding.xml';
+const LOCKING_XML_FILENAME = 'locking.xml';
+const OVERWRITING_XML_FILENAME = 'overwriting.xml';
```
TheJorgeBorras: Duplicated constant definitions.

We probably should think about creating an `ApplicationConstants.ts` and work on removing some of these values. It should improve readability and reduce duplicated definitions.

I would also go a step further and move some `ApplicationConstants` from `node-cli` down to the `sdk-core`. Same thing we already do on `vscode-extension` and `node-cli`.

packages/node-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@oracle/suitecloud-cli",
-	"version": "3.2.0",
+	"version": "4.2.0",
```
TheJorgeBorras: Should be `4.0.0`

packages/sdk-core/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "./commands/file/upload/UploadFilesHandler": "./build/commands/file/upload/UploadFilesHandler.js",
+    "./commands/file/import/ImportFilesHandler": "./build/commands/file/import/ImportFilesHandler.js",
+    "./package.json": "./package.json",
+    "./commands/object/ObjectCommandExecutor": "./build/commands/object/ObjectCommandExecutor.js",
+    "./commands/object/list/ListObjectsHandler": "./build/commands/object/list/ListObjectsHandler.js",
+    "./commands/object/import/ImportObjectsHandler": "./build/commands/object/import/ImportObjectsHandler.js",
+    "./commands/object/update/UpdateObjectsHandler": "./build/commands/object/update/UpdateObjectsHandler.js",
+    "./commands/proxy/generatekey/ProxyGenerateKeyHandler": "./build/commands/proxy/generatekey/ProxyGenerateKeyHandler.js",
+    "./commands/proxy/start/ProxyStartHandler": "./build/commands/proxy/start/ProxyStartHandler.js",
+    "./auth/AuthSessionManager": "./build/auth/AuthSessionManager.js"
```
TheJorgeBorras: Is this the best approach for wiring?

Maybe we can find a better and more scalable approach for exposing code to the `node-cli` package.
using `index.js` perhaps?

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
```diff
@@ -60,41 +59,51 @@ module.exports = class ManageAccountAction extends BaseAction {
...
-			action = MANAGE_ACTION.REMOVE;
-			authId = answers[COMMAND.OPTIONS.REMOVE];
-		} else if (answers.hasOwnProperty(COMMAND.OPTIONS.RENAME)) {
-			action = MANAGE_ACTION.RENAME;
-			authId = answers[COMMAND.OPTIONS.RENAME];
-		} else {
-			throwValidationException([NodeTranslationService.getMessage(ERRORS.OPTION_NOT_SPECIFIED)], this._runInInteractiveMode, this._commandMetadata);
+	_extractSelectedOptions(params) {
+		const selectedAction = selectManageAuthAction(params);
+		if (selectedAction.validationError) {
```
carolnavarropereira, line 78: It would be good selectedAction to have a type so the code can know which are the  properties of the object.
instead of exists validationError we can have a variable or method that returns either true or false.

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
```diff
@@ -60,41 +59,51 @@ module.exports = class ManageAccountAction extends BaseAction {
...
-		} else if (answers.hasOwnProperty(COMMAND.OPTIONS.LIST)) {
-			action = MANAGE_ACTION.LIST;
-		} else if (answers.hasOwnProperty(COMMAND.OPTIONS.REMOVE)) {
-			action = MANAGE_ACTION.REMOVE;
-			authId = answers[COMMAND.OPTIONS.REMOVE];
-		} else if (answers.hasOwnProperty(COMMAND.OPTIONS.RENAME)) {
-			action = MANAGE_ACTION.RENAME;
-			authId = answers[COMMAND.OPTIONS.RENAME];
-		} else {
-			throwValidationException([NodeTranslationService.getMessage(ERRORS.OPTION_NOT_SPECIFIED)], this._runInInteractiveMode, this._commandMetadata);
```
carolnavarropereira, line 93: It would be good a class that holds all the input parameters validation just to avoid noise into the main class.
Validation should be done outside the method that returns the selected options. They must be different methods.

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
carolnavarropereira, line 34: Unused in the current version.

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
```diff
@@ -60,41 +59,51 @@ module.exports = class ManageAccountAction extends BaseAction {
...
 		const message = this._getSpinnerMessage(selectedOptions);
 		const operationResult = await executeWithSpinner({
 			action: this._sdkExecutor.execute(executionContext),
 			message: message,
 		});
 		return operationResult.status === SdkOperationResultUtils.STATUS.SUCCESS
-			? ManageAccountActionResult.Builder.withData(this._prepareData(selectedOptions, operationResult.data))
+			? ManageAccountActionResult.Builder.withData(prepareManageAuthActionData(selectedOptions, operationResult.data))
 					.withResultMessage(operationResult.resultMessage)
 					.withExecutedAction(selectedOptions.action)
```
carolnavarropereira, line 70: withExecutedAction does not exist into the object

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
```diff
@@ -32,24 +38,17 @@ const COMMAND = {
...
 
 	async execute(params) {
+		const selectedOptions = this._extractSelectedOptions(params);
 		const sdkParams = CommandUtils.extractCommandOptions(params, this._commandMetadata);
 
 		const flags = [];
-		if (params[COMMAND.OPTIONS.LIST]) {
+		if (selectedOptions.action === MANAGE_ACTION.LIST) {
 			flags.push(COMMAND.OPTIONS.LIST);
 			delete sdkParams[COMMAND.OPTIONS.LIST];
```
carolnavarropereira, line 53: It looks like a patch. Can be solved in a different way?

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
```diff
@@ -22,37 +23,27 @@ module.exports = class AccountSetupCiAction extends BaseAction {
...
-
-		return params;
+		return normalizeSetupCiParams(params, GENERIC_NETSUITE_DOMAIN);
 	}
 
 	async execute(params) {
-		if (params[OPTIONS.DOMAIN] === GENERIC_NETSUITE_DOMAIN) {
-			delete params[OPTIONS.DOMAIN];
-		}
+		const normalizedParams = normalizeSetupCiParams(params, GENERIC_NETSUITE_DOMAIN);
```
carolnavarropereira: Parameters are both normalized in preExecute and execute.  It is possible to have a module of clearing params in preExecute and receive them in execute instead of the original ones.

packages/sdk-core/src/commands/account/setupci/SetupAccountCiHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+// - cli/.../handler/AuthenticateCIHandler.java (default production URL behavior)
+// - core/.../delegate/authentication/AuthenticateClientCredentialsDelegate.java (private key cleanup)
+const PRODUCTION_URL = 'system.netsuite.com';
+const BEGIN_PRIVATE_KEY = '-----BEGIN PRIVATE KEY-----';
+const END_PRIVATE_KEY = '-----END PRIVATE KEY-----';
+const CARRIAGE_RETURN = '\r';
+const NEWLINE_FEED = '\n';
+
+export type SetupCiParams = {
+	account?: string;
```
carolnavarropereira: Maybe it can be declared into an external file.

packages/sdk-core/src/commands/account/setupci/SetupAccountCiHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export function normalizeSetupCiParams(params: SetupCiParams, genericDomain: string = PRODUCTION_URL): SetupCiParams {
+	const normalizedParams: SetupCiParams = { ...params };
+
+	if (typeof normalizedParams.account === 'string') {
+		normalizedParams.account = normalizedParams.account.toUpperCase();
+	}
+
+	// Java command defaults to production URL when url is omitted.
+	// Keep CLI parameters aligned by omitting domain when the generic production domain is selected.
```
carolnavarropereira: account to uppercase and remove domain could be different methods.
I would make the inputParameters as a constant. I would create normalizedInputParameters as another different object. Any way, reomving parameters from the input object is not a good idea. It is better to use a method to identify these cases.

packages/sdk-core/src/commands/account/setupci/SetupAccountCiHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+
+	return normalizedParams;
+}
+
+export function getSetupCiAuthId(params: SetupCiParams, isSetupMode: boolean): string | undefined {
+	return isSetupMode ? (params.authid as string | undefined) : (params.select as string | undefined);
+}
+
+export function cleanPrivateKey(privateKey: string): string {
+	return privateKey
```
carolnavarropereira: This could be in a Util class

packages/sdk-core/src/commands/account/setupci/SetupAccountCiHandler.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+'use strict';
+
+// Java source of truth:
+// - cli/.../handler/AuthenticateCIHandler.java (default production URL behavior)
+// - core/.../delegate/authentication/AuthenticateClientCredentialsDelegate.java (private key cleanup)
+const PRODUCTION_URL = 'system.netsuite.com';
+const BEGIN_PRIVATE_KEY = '-----BEGIN PRIVATE KEY-----';
+const END_PRIVATE_KEY = '-----END PRIVATE KEY-----';
+const CARRIAGE_RETURN = '\r';
```
carolnavarropereira: CARRIAGE RETURN and NEWLINE should be in a StringUtils class

packages/sdk-core/src/commands/account/setupci/SetupAccountCiHandler.ts
```diff
@@ -0,0 +1,58 @@
+/*
+ ** Copyright (c) 2026 Oracle and/or its affiliates.  All rights reserved.
+ ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+ */
+'use strict';
+
+// Java source of truth:
+// - cli/.../handler/AuthenticateCIHandler.java (default production URL behavior)
+// - core/.../delegate/authentication/AuthenticateClientCredentialsDelegate.java (private key cleanup)
+const PRODUCTION_URL = 'system.netsuite.com';
```
carolnavarropereira: It should be in a external constants class

packages/sdk-core/src/commands/file/create/CreateFileHandler.ts
```diff
@@ -0,0 +1,83 @@
...
+	[key: string]: unknown;
+};
+
+function quoteString(value: string): string {
+	return `"${value}"`;
+}
+
+function unquoteString(value: string): string {
+	return value.replace(/^"(.*)"$/, '$1');
+}
```
TheJorgeBorras, line 34: Heavely duplicated functions.

These same functions appear on: here, `CommandUtils.js`, `ImportFilesHandler.ts`, `ListFilesHandler.ts`, `UploadFilesHandler.ts`, `AddDependenciesHandler.ts`, `CreateProjectHandler.ts` and `PackageHandler.ts`.

I propose we move `CommandUtilts.js` onto the `sdk-core` package and reference it from everywhere else.

packages/sdk-core/src/commands/file/create/CreateFileHandler.ts
```diff
@@ -0,0 +1,83 @@
+/*
+ ** Copyright (c) 2026 Oracle and/or its affiliates.  All rights reserved.
+ ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
+ */
+'use strict';
+
+import path from 'path';
```
TheJorgeBorras: Should be `import * as path from 'path';`

Otherwise WebStorm keeps complaining about `TS1259: Module 'path' can only be default-imported using the esModuleInterop flag`

packages/node-cli/src/commands/project/adddependencies/AddDependenciesAction.js
```diff
@@ -9,35 +9,36 @@ const BaseAction = require('../../base/BaseAction');
...
 	COMMAND_ADDDEPENDENCIES: { MESSAGES },
 } = require('../../../services/TranslationKeys');
 
-const COMMAND_OPTIONS = {
-	AUTH_ID: 'authid',
-	ALL: 'all',
-	PROJECT: 'project',
-};
-
 module.exports = class AddDependenciesAction extends BaseAction {
```
TheJorgeBorras, line 25: On the ticket it was decided that `project:adddependencies` would rely completly on Java implementation.

Therefore, we shouldn't mess with current implementation. We should revert the changes in this class.

packages/sdk-core/src/commands/project/adddependencies/AddDependenciesHandler.ts
```diff
@@ -0,0 +1,87 @@
...
+	if (Array.isArray(value)) {
+		return value.filter((entry): entry is string => typeof entry === 'string' && entry.trim().length > 0);
+	}
+	if (typeof value === 'string' && value.trim().length > 0) {
+		return [value];
+	}
+	return undefined;
+}
+
+function validateAddDependenciesInput(params: Record<string, unknown>): void {
```
TheJorgeBorras: On the ticket it was decided that `project:adddependencies` would rely completly on Java implementation.

We shouldn't mess with current flow.

packages/sdk-core/src/commands/proxy/generatekey/ProxyGenerateKeyHandler.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export const PROXY_GENERATE_KEY_ERROR = {
+	INVALID_KEY_FILE_CONTENTS: 'INVALID_KEY_FILE_CONTENTS',
+} as const;
+
+type ProxyGenerateKeyErrorCode = typeof PROXY_GENERATE_KEY_ERROR[keyof typeof PROXY_GENERATE_KEY_ERROR];
+
+export type ProxyGenerateKeyValidationError = {
+	errorCode: ProxyGenerateKeyErrorCode;
+};
```
TheJorgeBorras, line 38: This whole command, it's arquitecture and it's flow needs to be reviewed.

It doesn't make much sense to have the logic for this command be segmented equally into both `node-cli` and `sdk-core` (especially since we decided to rely on Java implementation).

Parsing and Validation of ClientAPIKeyFileContents is handled both here as well as `ClientAPIKeyObjectWrapper.js` (duplicated). But the SDKExecutor call to save/read only executed on Node-CLI `ClientAPIKeyUtils.js`

packages/node-cli/src/commands/project/validate/ValidateAction.js
```diff
@@ -52,36 +69,26 @@ module.exports = class ValidateAction extends BaseAction {
...
-				.addParams(sdkParams)
-				.addFlags(flags)
-				.build();
-
-			const operationResult = await executeWithSpinner({
-				action: this._sdkExecutor.execute(executionContext),
+			// Local validate is temporarily removed during TS-core migration.
+			const validateExecution = prepareValidateExecution({
+				...params,
+				[COMMAND_OPTIONS.SERVER]: true,
```
carolnavarropereira: This code: prepareValidateExecution is not used.
Currently both standart and server validation do the same.

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
TheJorgeBorras: These flags don't make much sense to me, even looking at Java. Aren't these only compatible with Deploy?
I'm okay with removing them 👍 .

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
TheJorgeBorras: It's not new code ++ this is already happening in a bunch of places in this codebase...

Let's leave this alone for now.

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
TheJorgeBorras: This comment seems super related to layer contracts, APIs and passing types between layers.

I share your concern, but if it works let's not worry about it now.
However, If you have a quick fix to simplify this flow, I'm okay with discussing it.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
TheJorgeBorras: Good catch, let's remove the 2nd `normalizeSetupCiParams()` call 
👍

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
TheJorgeBorras: I'm unsure about what you're speaking of.

The delete was already part of the codebase, it's not new code.

packages/node-cli/src/commands/project/adddependencies/AddDependenciesAction.js
TheJorgeBorras: I still think we should go for this. What do you think Carol?

packages/node-cli/src/commands/project/deploy/DeployAction.js
TheJorgeBorras: Old code. Let's leave it.

packages/sdk-core/src/commands/project/package/PackageProjectExecutor.ts
TheJorgeBorras: Let's fix this :+1:

packages/node-cli/src/commands/project/validate/ValidateAction.js
TheJorgeBorras: I'm okay with either

packages/sdk-core/src/commands/account/setupci/SetupAccountCiHandler.ts
TheJorgeBorras: Agreed on meeting to not go for this change yet.

packages/sdk-core/src/commands/account/setupci/SetupAccountCiHandler.ts
TheJorgeBorras: Agreed on meeting to not go for this change yet.

packages/sdk-core/src/commands/account/setupci/SetupAccountCiHandler.ts
TheJorgeBorras: Agreed on meeting to not go for this change yet.

packages/sdk-core/src/commands/file/create/CreateFileHandler.ts
TheJorgeBorras: Agreed on meeting to not go for this change yet.

packages/sdk-core/src/commands/project/create/CreateProjectExecutor.ts
TheJorgeBorras: Agreed on meeting to not go for this change yet.

packages/sdk-core/src/commands/project/create/CreateProjectExecutor.ts
TheJorgeBorras: Probably not worth the effort at the moment.

packages/sdk-core/src/commands/proxy/generatekey/ProxyGenerateKeyHandler.ts
TheJorgeBorras: If it works, we can leave it as it is.

Still, I prefer we either merge both implementations (might be too much effort atm) or remove the new implementation.

packages/sdk-core/package.json
TheJorgeBorras: Agreed on meeting to not go for this change yet.

packages/node-cli/src/utils/AuthenticationUtils.js
```diff
@@ -78,6 +82,37 @@ const FLAGS = {
...
+}
+
+function _extractHostNameFromManageAuthData(authData) {
+	if (!authData || typeof authData !== 'object') {
+		return undefined;
+	}
+
+	return authData.hostInfo?.hostName || authData.domain;
+}
+
```
TheJorgeBorras: We don't need this, I don't think we should keep this extra validations.

Unwanted noise.

packages/node-cli/src/commands/account/setup/SetupAction.js
```diff
@@ -21,7 +24,8 @@ module.exports = class SetupAction extends BaseAction {
 	async execute(params) {
 		try {
 			if (params.mode === AUTH_MODE.OAUTH) {
-				return await authenticateWithOauth(params, this._sdkPath, this._executionPath, this._executionEnvironmentContext);
+				const normalizedParams = normalizeSetupParams(params, GENERIC_NETSUITE_DOMAIN);
+				return await authenticateWithOauth(normalizedParams, this._sdkPath, this._executionPath, this._executionEnvironmentContext);
 			} else if (params.mode === AUTH_MODE.REUSE) {
```
TheJorgeBorras, line 25: This is another command that remains unchanged functionality wise (still relies on old Node-CLI+Java flow).

I'd feel safer if we removed the extra validations in this file as well... Let me know what do you think of this.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
```diff
@@ -11,6 +11,7 @@ const { ACCOUNT_SETUP_CI: { COMMAND: { OPTIONS } } } = require('./AccountSetupCi
 const ProjectInfoService = require('../../../services/ProjectInfoService');
 const { validateMachineToMachineAuthIsAllowed } = require('../../../services/ExecutionContextService');
 const AccountSetupCiValidation = require('./AccountSetupCiValidation');
+const { normalizeSetupCiParams, getSetupCiAuthId } = require('@oracle/suitecloud-sdk-core/commands/account/setupci/SetupAccountCiHandler');
 
 module.exports = class AccountSetupCiAction extends BaseAction {
```
TheJorgeBorras, line 15: This is another command that remains unchanged functionality wise (still relies on old Node-CLI+Java flow).

I'd feel safer if we removed the extra validations in this file as well... Let me know what do you think of this.

packages/sdk-core/src/commands/project/create/CreateProjectExecutor.ts
carolnavarropereira: I agree with you, however I wouldn't merge with the source code either. We need to create another folder for resources.

packages/node-cli/src/commands/account/manageauth/ManageAccountAction.js
carolnavarropereira: I do not like the idea of deleting members of a class, I think we need a better design, but we can leave it for now.

packages/node-cli/src/commands/project/adddependencies/AddDependenciesAction.js
carolnavarropereira: Yes, I agree.  Even if the results are the same, we do not need to change the small details about how it is working currently and they are both relying on the java code.

packages/sdk-core/src/commands/project/adddependencies/AddDependenciesHandler.ts
carolnavarropereira: New implementation is still relying on java code, so we can revert it to the old code.

packages/sdk-core/src/commands/proxy/generatekey/ProxyGenerateKeyHandler.ts
carolnavarropereira: We can leave it for now, but I think we may rethink a better design.

packages/node-cli/src/utils/AuthenticationUtils.js
carolnavarropereira: Agreed.

packages/node-cli/src/commands/account/setup/SetupAction.js
carolnavarropereira: If it is unchanged we do not need extra validations.

packages/node-cli/src/commands/account/setupci/AccountSetupCiAction.js
carolnavarropereira: If it is unchanged we do not need extra validations.

packages/node-cli/src/commands/project/package/PackageAction.js
```diff
@@ -4,56 +4,49 @@
...
-
 	preExecute(params) {
-		this._projectInfoService.checkWorkingDirectoryContainsValidProject(this._commandMetadata.name);
-
 		AccountSpecificValuesUtils.validate(params, this._projectFolder);
 
 		return {
-			[COMMAND_OPTIONS.DESTINATION]: CommandUtils.quoteString(path.join(this._executionPath, DEFAULT_DESTINATION_FOLDER)),
-			[COMMAND_OPTIONS.PROJECT]: CommandUtils.quoteString(this._projectFolder),
+			...preparePackageParams(params, this._executionPath, this._projectFolder),
```
carolnavarropereira, line 24: No need to create preparePackageParams. Keep these lines here like in the old version.

packages/node-cli/src/commands/project/package/PackageAction.js
```diff
@@ -4,56 +4,49 @@
...
 	}
 
 	async execute(params) {
 		const sdkParams = CommandUtils.extractCommandOptions(params, this._commandMetadata);
 
-		const executionContext = SdkExecutionContext.Builder.forCommand(this._commandMetadata.sdkCommand).integration().addParams(sdkParams).build();
-
 		await this._log.warning(NodeTranslationService.getMessage(COMMAND_PACKAGE.LOCAL_VALIDATION_NOT_EXECUTED));
 
 		const operationResult = await executeWithSpinner({
```
carolnavarropereira, line 34: Differences with previous version and new one:
Previous version: 
New version: 
  Does not filter files that are not from the project: added  <project>/src/example.txt, project.json, suitecloud.log
It does need to implement filtering.

packages/node-cli/src/metadata/SdkCommandsMetadataPatch.json
```diff
@@ -117,14 +136,36 @@
...
+					"name": "json",
+					"option": "json",
+					"alias": "j",
+					"description": "Prints the raw JSON response from the server and skips formatted CLI output.",
+					"mandatory": false,
+					"type": "FLAG",
+					"usage": "",
+					"defaultOption": false,
+					"disableInIntegrationMode": false
+				}
```
TheJorgeBorras, line 131: We need to evaluate this.
Most likely we don't want to show this option to users

ask Ali

packages/node-cli/src/metadata/SdkCommandsMetadataPatch.json
```diff
@@ -117,14 +136,36 @@
...
```
TheJorgeBorras, line 149: We need to evaluate this.
Most likely we don't want to show this option to users

ask Ali

packages/sdk-core/src/commands/project/deploy/DeployHandler.ts
```diff
@@ -0,0 +1,106 @@
...
+	PREVIEW: 'preview',
+} as const;
+
+export const DEPLOY_COMMAND = {
+	FLAGS: {
+		NO_PREVIEW: 'no_preview',
+		PREVIEW: 'dryrun',
+		SKIP_WARNING: 'skip_warning',
+		VALIDATE: 'validate',
+		APPLY_INSTALLATION_PREFERENCES: 'applyinstallprefs',
```
TheJorgeBorras, line 22: Review flags. Most of these are no longer used.
Proposal:
* Keep: `PREVIEW ('dryrun')` and `APPLY_INSTALLATION_PREFERENCES (applyinstallprefs)`
* Remove: `no_preview`, `skip_warning` and `validate`

Change flags on DeployAction.js as well.
Ask Ali.

packages/sdk-core/src/commands/project/ProjectApiClient.ts
```diff
@@ -0,0 +1,205 @@
...
+
+	try {
+		await execFileAsync(ZIP_BINARY_NAME, zipCommandArgs, { cwd: projectFolder });
+		return projectArchivePath;
+	} catch (error: any) {
+		throw new Error(
+			`Unable to archive project folder "${projectFolder}". Verify "${ZIP_BINARY_NAME}" is installed and available in PATH.`
+		);
+	}
+}
```
TheJorgeBorras: It has already been noticed by Carol,
but it makes virtually no sense that `project:package` and `project:deploy`, `project:deploy --dry-run` and `project:validate` don't share the same packaging into a zip flow.

packages/sdk-core/src/commands/file/FileCommandExecutor.ts
```diff
@@ -0,0 +1,779 @@
...
+
+async function unzipArchive(zipFilePath: string, destinationFolder: string): Promise<void> {
+	try {
+		await execFileAsync(UNZIP_BINARY_NAME, ['-o', zipFilePath, '-d', destinationFolder]);
+	} catch (error: unknown) {
+		throw new Error(
+			`Unable to extract imported files. Verify "${UNZIP_BINARY_NAME}" is installed and available in PATH.`
+		);
+	}
+}
```
TheJorgeBorras: If I'm not mistaken,
this too relies on the Linux/macOS system `unzip` command (not sure if Windows also has unzip or not).

We should review it, and substitute for a Node-only implementation that works for all OSs

packages/sdk-core/src/commands/file/create/CreateFileExecutor.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+function generateSuiteScriptContent(type?: string, module?: string): string {
+	const modules = normalizeModuleList(module);
+	const deps = modules.map((moduleName) => `'${moduleName}'`).join(', ');
+	const args = modules.map(toModuleAlias).join(', ');
+	const scriptTypeLine = type && !/^custommodule$/i.test(type) ? ` * @NScriptType ${type}\n` : '';
+
+	return `/**\n * @NApiVersion 2.x\n${scriptTypeLine} */\ndefine([${deps}], (${args}) => {\n\treturn {};\n});\n`;
+}
```
TheJorgeBorras: This core SuiteScript creation function is too basic.
Compared to old vs new files, it's clear it's missing fields, comments and functions for basic SuiteScript types.

See screenshot for reference (left file was generated using new core vs right using old core):

Image

TheJorgeBorras (approved):

carolnavarropereira (approved):

PR #1045, opened 2026-06-04
feature/pdpdevtool-6510: add checksum for jar

davidecorreu:


packages/node-cli/src/core/sdksetup/SdkArtifactVerifier.js
```diff
@@ -0,0 +1,44 @@
...
+const fs = require('fs');
+
+const NodeTranslationService = require('../../services/NodeTranslationService');
+const { SDK_DOWNLOAD_SERVICE } = require('../../services/TranslationKeys');
+
+const SHA256_PATTERN = /^[a-fA-F0-9]{64}$/;
+
+class SdkArtifactVerifier {
+	verify(sdkPath, sdkProperties) {
+		if (sdkProperties.isCustomSdkMetadataUsed()) {
```
carolnavarropereira: The method can return two values: True or Error.
I believe to be more consistant: true/false or void/error
Also SdkArtifactVerifier.verify(temporaryDestinationFilePath, SdkProperties); do not read the return.

So I think it is better void/error.
However (optional)we can also split it logically:
if (shouldChecksumBeVerified()) { //returns boolean
      verifyChecksum(); // void/Error.
}

packages/node-cli/src/core/sdksetup/SdkArtifactVerifier.js
```diff
@@ -0,0 +1,44 @@
...
+		if (actualSha256 !== normalizedExpectedSha256) {
+			throw new Error(NodeTranslationService.getMessage(SDK_DOWNLOAD_SERVICE.CHECKSUM_MISMATCH));
+		}
+
+		return true;
+	}
+
+	calculateSha256(sdkPath) {
+		return crypto.createHash('sha256').update(fs.readFileSync(sdkPath)).digest('hex');
+	}
```
carolnavarropereira, line 39: sha256 and hex can be class constants.

packages/node-cli/src/core/sdksetup/SdkArtifactVerifier.js
```diff
@@ -0,0 +1,44 @@
...
+		const actualSha256 = this.calculateSha256(sdkPath);
+
+		if (actualSha256 !== normalizedExpectedSha256) {
+			throw new Error(NodeTranslationService.getMessage(SDK_DOWNLOAD_SERVICE.CHECKSUM_MISMATCH));
+		}
+
+		return true;
+	}
+
+	calculateSha256(sdkPath) {
```
carolnavarropereira: sin normalize starts by _ we can do the same with this method.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -43,12 +43,17 @@ class SdkDownloadService {
 
 		const fullURL = `${SdkProperties.getDownloadURL()}/${SdkProperties.getSdkFileName()}`;
 		const destinationFilePath = path.join(sdkDirectory, SdkProperties.getSdkFileName());
+		const temporaryDestinationFilePath = `${destinationFilePath}.tmp`;
```
carolnavarropereira, line 46: This is optional, feel free to disregard it if you don't agree.
Not a good idea to have "dangerous" files in places that can be executed by an error. We could write it into the temp folder of the OS.

const os = require('os');
os.tmpdir();
However it is an opinion. It is okay this way too.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
```diff
@@ -106,6 +108,16 @@ class SdkDownloadService {
 			.filter((file) => /[.]jar$/.test(file))
 			.map((file) => fs.unlinkSync(path.join(folder, file)));
 	}
+
+	_removeFileIfExists(filePath) {
```
carolnavarropereira: This is a generic function and used in many places. It would be nice to be added into the FileUtils.js

packages/node-cli/src/core/sdksetup/SdkProperties.js
```diff
@@ -8,39 +8,32 @@ const path = require('path');
...
 	getSdkFileName() {
-		// read config.js file if exists or use package.json
-		const sdkFileName = this.configFileExists() ? CONFIG_FILE_CACHE.sdkFilename : sdkFilename
-		return sdkFileName;
+		return this._sdkMetadata.sdkFilename;
 	}
 
-	configFileExists() {
-		return CONFIG_FILE_CACHE !== null;
+	getSdkSha256() {
```
carolnavarropereira, line 31: Unused method.

packages/node-cli/src/core/sdksetup/SdkProperties.js
```diff
@@ -8,39 +8,32 @@ const path = require('path');
...
+		return this._sdkMetadata.sdkSha256;
 	}
 
-	_loadCache() {
-		if (fs.existsSync(path.resolve(__dirname, CONFIG_FILE))) {
-			CONFIG_FILE_CACHE = require(CONFIG_FILE);
-		}
+	isCustomSdkMetadataUsed() {
```
carolnavarropereira, line 35: Unused method

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
```diff
@@ -1,47 +1,45 @@
...
+export function getSdkSha256(): string {
+	return SDK_METADATA.sdkSha256;
+}
+
+export function isCustomSdkMetadataUsed(): boolean {
+	return IS_CUSTOM_SDK_METADATA_USED;
+}
+
+export function getSdkDownloadFullUrl(): string {
+	return getSdkDownloadUrl() + '/' + getSdkFilename();
```
carolnavarropereira, line 44: Optional: you can use new URL()

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
```diff
@@ -1,125 +1,149 @@
...
+import * as path from 'path';
+import * as https from 'https';
+import { homedir } from 'os';
+import { parse } from 'url';
+import { VSTranslationService } from '../../service/VSTranslationService';
+import { ERRORS, EXTENSION_INSTALLATION } from '../../service/TranslationKeys';
+import MessageService from '../../service/MessageService';
+import { validateSdk } from './SdkValidator';
+import { ApplicationConstants, EnvironmentInformationService, FileSystemService, SdkArtifactVerifier } from '../../util/ExtensionUtil';
+import * as SdkProperties from './SdkProperties';
```
carolnavarropereira, line 17: You can replace * by the full list of properties.

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
```diff
@@ -1,125 +1,149 @@
...
+	} catch (error) {
+		messageService.showErrorMessage(translationService.getMessage(EXTENSION_INSTALLATION.ERROR.GENERAL_ERROR, fullUrl, `${error}`));
+		throw error;
+	}
+}
+
+async function downloadFile(url: string, sdkDirectory: string) {
+	const sdkDestinationFile = path.join(sdkDirectory, SdkProperties.getSdkFilename());
+	const temporarySdkDestinationFile = `${sdkDestinationFile}.tmp`;
+	const parsedUrl = parse(url);
```
carolnavarropereira: This is deprecated.

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
```diff
@@ -1,125 +1,149 @@
...
+		const sdk = await save(options, file);
+		file.close();
+		SdkArtifactVerifier.verify(sdk, sdkProperties);
+		fs.renameSync(temporarySdkDestinationFile, sdkDestinationFile);
+		if (!(await validateSdk(sdkDestinationFile))) {
+			throw translationService.getMessage(EXTENSION_INSTALLATION.ERROR.SDK_INVALID);
+		}
+	} finally {
+		if (file) {
+			file.close();
```
carolnavarropereira: I think this line if (file) file.close() could be executed before the finally, because if it fails maybe removeFileIfExists is never executed

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
```diff
@@ -1,125 +1,149 @@
...
+	try {
+		return SdkArtifactVerifier.verify(sdkPath, sdkProperties);
+	} catch (error) {
+		removeFileIfExists(sdkPath);
+		return false;
+	}
+}
+
+function isSuccessStatusCode(statusCode: number | undefined) {
+	return statusCode !== undefined && statusCode >= 200 && statusCode < 300;
```
carolnavarropereira, line 143: Maybe we can have the statusCode in constants file (optional)

carolnavarropereira (changes requested):

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
davidecorreu: I'll move it to FileSystemService as it looks like a better candidate.

packages/node-cli/src/core/sdksetup/SdkArtifactVerifier.js
davidecorreu: Done.

packages/node-cli/src/core/sdksetup/SdkArtifactVerifier.js
davidecorreu: Done.

packages/node-cli/src/core/sdksetup/SdkArtifactVerifier.js
davidecorreu: Done.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
davidecorreu: Done.

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
davidecorreu: Changed to use URL object instead.

packages/vscode-extension/src/core/sdksetup/SdkProperties.ts
davidecorreu: Not changing  it here as new URL is used later based on this string composition.

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
davidecorreu: Not changing it as I do like to have all of them under SdkProperties so it's clear where does functions come from when being used.

packages/node-cli/src/core/sdksetup/SdkDownloadService.js
davidecorreu: I would not change it for the following reasons:

- The file is stored in a non-user-facing directory `<userHome>/.suitecloud-sdk`
- I don't want to deal with potential permission problems in a temp directory
- Keep code simple

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
davidecorreu: Removing file.close() from finally and keep cleanup unconditional.

Since the stream is already closed after save(...), I guess the minimal safe change is to remove the close from finally and keep removeFileIfExists(...) there.

packages/vscode-extension/src/core/sdksetup/SdkServices.ts
davidecorreu: Added a short comment to clarify that this checks the HTTP 2xx success range. I avoided constants because they seem too noisy for this simple predicate.

packages/node-cli/src/core/sdksetup/SdkProperties.js
davidecorreu: Used by SdkArtifactVerifier.verify

packages/node-cli/src/core/sdksetup/SdkProperties.js
davidecorreu: Used by SdkArtifactVerifier.verify

carolnavarropereira (approved):

packages/node-cli/package.json
```diff
@@ -23,8 +23,9 @@
 	"main": "./src/suitecloud.js",
 	"preferGlobal": true,
 	"nsCompatibleVersion": "2026.1",
```
alisyed-19, line 25: This should be 2026.1?

davidecorreu: I guess 2026.1 is currently the correct version for `dev` branch (at least for now).

alisyed-19 (approved):

PR #1068, opened 2026-06-12
bugfix/Proxy API Key Generation

alisyed-19:


packages/node-cli/src/utils/ClientAPIKeyUtils.js
```diff
@@ -95,7 +93,7 @@ async function resolveDefaultClientApiKey(sdkExecutor) {
 }
 
 function formatForSdkCommandlineArgument(fileContent) {
-	return '"' + fileContent.replaceAll('"', String.raw`\"`) + '"';
+	return fileContent;
 }
```
TheJorgeBorras: This function no longer does anything. It just returns the same fileContent as inputed.

Let's remove it.

TheJorgeBorras (approved):

carolnavarropereira (approved):

PR #1083, opened 2026-07-08
Centralize sdk-core templates and align project/file creation with Java parity

alisyed-19:
This MR moves project and SuiteScript template ownership from Node CLI into sdk-core, providing one source of truth for project and file creation.
It aligns generated projects and SuiteScript files with the Java implementation, preserves Node CLI-specific scaffolding, centralizes errors and messages, and ensures templates are included when sdk-core is packaged. It also adds validation, path-safety checks, cleanup on failed creation, and expanded unit coverage for ACP and SuiteApp workflows.


packages/sdk-core/src/templates/TemplateLoader.ts
```diff
@@ -0,0 +1,44 @@
...
+		templateCache.set(relativePath, template);
+	}
+	return template;
+}
+
+export function renderTemplate(template: string, values: Record<string, string>): string {
+	return Object.entries(values).reduce((content, [name, value]) => {
+		return content
+			.replaceAll(`\${${name}}`, value)
+			.replaceAll(`{{${name}}}`, value);
```
TheJorgeBorras: WebStorm keeps complaining that `TS2550: Property replaceAll does not exist on type string. Do you need to change your target library? Try changing the lib compiler option to es2021 or later.` for `.replaceAll(...)`.

After compiling the code seems to work fine, but I'm wondering if we should:
a.  just ignore it
b. change some of the complier lib options (like WebStorm suggests)
c. go for a more neutral approach that is supported everywhere, e.g:

```
		return content
			.split(`\${${name}}`).join(value)
			.split(`{{${name}}}`).join(value);
```

packages/sdk-core/src/templates/TemplateLoader.ts
```diff
@@ -0,0 +1,44 @@
...
+	}, template);
+}
+
+function escapeXml(value: string): string {
+	return value
+		.replaceAll('&', '&amp;')
+		.replaceAll('<', '&lt;')
+		.replaceAll('>', '&gt;')
+		.replaceAll('"', '&quot;')
+		.replaceAll("'", '&apos;');
```
TheJorgeBorras: Same thing as earlier comment.
`
TS2550: Property replaceAll does not exist on type string. Do you need to change your target library? Try changing the lib compiler option to es2021 or later.`

carolnavarropereira (approved):

packages/sdk-core/resources/templates/suitescript/ss_2_x_templates.xml
```diff
@@ -0,0 +1,87 @@
...
+      <types>
+        <files label="Client Script" bodyFilename="ss_2_0_client.js" />
+      </types>
+    </template>
+
+    <template id="CustomModule" label="Custom Module" defaultFilename="CustomModule.js" headerFilename="ss_header.js">
+      <types>
+        <files label="Custom Module" bodyFilename="ss_2_1_custom_module.js" />
+      </types>
+    </template>
```
TheJorgeBorras: Personally, I see no reason why this should remain a `.xml` file.
I think it would be far better to translate it into a .ts service that supports a native data-structure.
This way we can avoid parsing this file every time.

I would go into a similar direction for `suitescript_modules.xml` as well.

For reference, I was thinking about something like this:

```
 /** Copyright (c) 2026 Oracle and/or its affiliates.  All rights reserved.
 ** Licensed under the Universal Permissive License v 1.0 as shown at https://oss.oracle.com/licenses/upl.
 */
'use strict';

export type TemplateMatcherEntry = {
	name: string;
	defaultFilename: string;
	path: string;
};

export const DEFAULT_SCRIPT_TYPE = 'CustomModule';

export const SDK_SUITESCRIPT_TEMPLATE_TYPE_MATCHER: Record<string, TemplateMatcherEntry> = {
	BundleInstallationScript: {
		name: 'Bundle Installation Script',
		defaultFilename: 'BundleInstallationScript.js',
		path: 'ss_2_1_bundle_installation.js',
	},
	ClientScript: {
		name: 'Client Script',
		defaultFilename: 'ClientScript.js',
		path: 'ss_2_0_client.js',
	},
	CustomModule: {
		name: 'Custom Module',
		defaultFilename: 'CustomModule.js',
		path: 'ss_2_1_custom_module.js',
	},
	plugintypeimpl: {
		name: 'Custom Plug-in',
		defaultFilename: 'CustomPlugin.js',
		path: 'ss_2_0_plugin_impl.js',
	},
	customRecordAction: {
		name: 'Custom Record Action Script',
		defaultFilename: 'CustomRecordActionScript.js',

...
```

packages/sdk-core/src/commands/project/create/CreateProjectHandler.ts
```diff
@@ -129,7 +130,8 @@ export async function executeCreateProjectCommand(
 }
```
TheJorgeBorras: This function is no longer being used anywhere

packages/sdk-core/src/commands/project/create/CreateProjectHandler.ts
```diff
@@ -129,7 +130,8 @@ export async function executeCreateProjectCommand(
...
 
 export async function executeCreateProjectWorkflowCommand(
-	input: CreateProjectWorkflowInput
+	input: CreateProjectWorkflowInput,
+	dependencies?: CreateProjectWorkflowDependencies
 ): Promise<CreateProjectWorkflowOperationResult> {
-	return executeCreateProjectWorkflow(input);
+	return executeCreateProjectWorkflow(input, dependencies);
```
TheJorgeBorras: In my opinion this isn't very consistent with the rest of commands.

I think `CreateProjectAction.js` should call `CreateProjectExecutor.executeCreateProjectWorkflow()`, just like most other commands already do.

This just adds unnecessary code and makes the flow harder to track unnecessarily.

packages/sdk-core/resources/templates/suitescript/ss_2_x_templates.xml
alisyed-19: The catalogs are currently cached and parsed only once, but I agree that native TypeScript is simpler. I will move SuiteScript template and module metadata into sdk-core and update Node CLI to consume the same exported metadata, avoiding another duplicated source of truth.

TheJorgeBorras (approved):

PR #1086, opened 2026-07-12
Remove deprecated server flag and add a warning

alisyed-19:


TheJorgeBorras (approved):

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -4,7 +4,6 @@
  */
 'use strict';
 
-const { ActionResult } = require('../../../services/actionresult/ActionResult');
 const DeployActionResult = require('../../../services/actionresult/DeployActionResult');
```
carolnavarropereira, line 8: ActionResult is used. It must not be removed.

PR #1087, opened 2026-07-13
TS SDK-Core: Transfer all user messages to Translation Service

TheJorgeBorras:


packages/sdk-core/src/commands/project/ProjectApiClient.ts
```diff
@@ -150,7 +152,7 @@ async function sendHttpsMultipartRequest(input: {
 	hostName: string;
 	pathname: string;
 	accessToken: string;
-	payload: { payload: Buffer; boundary: string };
+	payload: { payload: Buffer; boundary: string };COMPARE_FILE
```
alisyed-19: COMPARE_FILE
Why is it here?

TheJorgeBorras: No idea, I must have pasted it here without knowing.

Removed in latest commit.

packages/sdk-core/src/commands/project/ProjectCommandExecutor.ts
```diff
@@ -18,6 +18,8 @@ import {
 	PROJECT_COMMAND,
 	SDK_OPERATION_STATUS,
 } from './ProjectCommandTypes';
+import { PROJECT_COMMAND } from '../../services/translation/TranslationKeys';
```
alisyed-19: PROJECT_COMMAND is already imported from ProjectCommandTypes.

packages/sdk-core/src/commands/file/FileCommandExecutor.ts
```diff
@@ -768,3 +773,7 @@ function asString(value: unknown): string {
 	}
 	return '';
 }
+
```
carolnavarropereira: Is it necessary this method? If it is, move it to a Utils class . I have also seen it in FileCommandExcutor.ts and ObjectCommandExecutor.ts, ProjectResultFormatter.ts

alisyed-19 (approved):

packages/sdk-core/src/commands/file/FileCommandExecutor.ts
carolnavarropereira: Changing the name of the method is not the point. You should move it to a Utils class.

TheJorgeBorras: This is why I don't like working with AI.
I removed twice and then it rewrote the function once again

smh

carolnavarropereira (approved):

PR #1088, opened 2026-07-13
Feature/sdk core architecture refactor

alisyed-19:


packages/node-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@oracle/suitecloud-cli",
-	"version": "4.2.0",
+	"version": "4.3.0",
```
TheJorgeBorras: This shouldn't be `"4.2.0"` nor `"4.3.0"`...

It should be `"4.0.0"` right? Might as well fix it here...

TheJorgeBorras (approved):

carolnavarropereira (approved):

PR #1100, opened 2026-07-29
TS-Core: Improve project command output summaries

alisyed-19:


carolnavarropereira (approved):

packages/node-cli/src/commands/project/deploy/DeployAction.js
```diff
@@ -133,6 +133,11 @@ module.exports = class DeployAction extends (
 	}
 
 	async _deploy(params, flags) {
+		const isApplyInstallationPreferences = isApplyInstallationPreferencesForDeploy(
+			this._projectType,
+			flags,
+			PROJECT_SUITEAPP
+		);
 		try {
```
TheJorgeBorras: This validation is redundant, since the important checks already happen in the `preExecute();`
```
# line 77
ApplyInstallationPreferencesUtils.validate(params, this._projectFolder, this._commandMetadata.name, this._log);
```


Moreover, this isn't consistend with ValidationAction.js.

I propose we remove this function (the reference here and definition in sdk-core) and instead go with the ValidationAction.js approach: 

```let installationPreferencesApplied = !!params[COMMAND_OPTIONS.APPLY_INSTALLATION_PREFERENCES];```

alisyed-19: Thanks, agreed. I removed the redundant helper and derive the state from the prepared deployment flags. I also replaced the helper-specific test with focused validation coverage for valid SuiteApps, ACP rejection, and SuiteApps without installation-preference files.

TheJorgeBorras (approved):

PR #1137, opened 2026-08-17
config:import first version

carolnavarropereira:


packages/sdk-core/src/commands/config/import/ImportConfigurationHandler.ts
```diff
@@ -0,0 +1,17 @@
...
+	ConfigurationCommandOperationResult,
+	ImportConfigurationExecutionInput,
+} from '../../../api/config/ConfigurationCommand';
+import { executeImportConfiguration } from './ImportConfigurationExecutor';
+
+export function executeImportConfigurationCommand(
+	input: ImportConfigurationExecutionInput
+): Promise<ConfigurationCommandOperationResult> {
+	return executeImportConfiguration(input);
+}
```
TheJorgeBorras: I suggest we avoid keeping this empty class in our code.
From a practical viewpoint, It's doesn't contribute any value in any way. It's obfuscating the real logic within `ImportConfigurationExecutor`.

I propose we either:
- remove this class.
- or move the contents of `ImportConfigurationExecutor.ts` into this one and delete the latter instead.

carolnavarropereira: All the classes use the same design, if we remove it we will break the consistency, I would rather keep it this way even if empty.

TheJorgeBorras: The thing is that the rest Handlers do "something" and have (sometimes stronger sometimes weaker) reasons for being there.
In this case it's obfuscating logic and that's it. 

But I don't have a strong preference about it. If you feel it's cleaner to keep it then let's leave it as it is.

packages/node-cli/src/metadata/CommandGenerators.json
```diff
@@ -1,4 +1,9 @@
 [
+	{
+		"commandName": "config:import",
+		"supportsInteractiveMode": false,
+		"generator": "commands/config/import/ImportConfigurationCommand"
+	},
```
TheJorgeBorras: I suggest we maintain command alphabetical order.

Therefore, let's declare `config:import` under `account:setup:ci` (line 22)

packages/node-cli/src/metadata/NodeCommandsMetadata.json
```diff
@@ -1,4 +1,19 @@
...
+		"options": {
+			"authid": {
+				"name": "authid",
+				"description": "References the custom alias you give to a specific account-role combination. Referred to as authentication ID and auth ID in the documentation.",
+				"type": "SINGLE",
+				"mandatory": true,
+				"disableInIntegrationMode": false
+			}
+		}
+	},
```
TheJorgeBorras: I suggest we maintain command alphabetical order.

Therefore, Let's declare `config:import` under `account:setup` (line 34)

packages/node-cli/src/metadata/CommandGenerators.json
carolnavarropereira: I don't think it is important, but done it.

packages/node-cli/src/metadata/NodeCommandsMetadata.json
carolnavarropereira: Done it.

antoniofruci (approved):

packages/sdk-core/src/commands/config/import/ImportConfigurationExecutor.ts
```diff
@@ -0,0 +1,149 @@
...
+		const statusXml = await readOptionalFile(statusFilePath);
+		if (!statusXml) {
+			return errorResult(translationService.getMessage(CONFIGURATION.ERROR.UNKNOWN_SERVER_RESPONSE));
+		}
+
+		await rm(statusFilePath, { force: true });
+		await mkdir(targetFolder, { recursive: true });
+		await copyDirectoryContents(unzipFolder, targetFolder);
+
+		const statusItems = await parseImportObjectStatus(statusXml);
```
alisyed-19: Could we parse and validate status.xml (line 108) before copying the extracted files into targetFolder? If status parsing fails, this currently returns an error after potentially overwriting files under AccountConfiguration.

alisyed-19 (approved):

PR #1147, opened 2026-08-28
PDPDEVTOOL-6612 - Bugfix 1146 object xml root tag

### LINKED ISSUE #1146: Unable to parse object XML root tag - error when trying to update saved search from account
### Select the SuiteCloud tool where you encountered the issue

SuiteCloud Extension for VSCode

### Version of the SuiteCloud Tool

4.0.0

### Operating system

Windows 11

### Node Version / VSCode Version

1.134.0 (user setup)

### JAVA Version

java 17.0.12+8-LTS-286

### Bug description

Getting the following error when trying to update a saved search from account:
The following objects were not updated from the account to your project:
    - customsearch_xxxxxxx: Unable to parse object XML root tag.

### Steps To Reproduce

1. In Visual Studio Code using the SuiteCloud Extension
2. Right click on a saved search in your project and select "Update Object from Account"
3. Click "Continue" when prompted that the object will be overwritten
4. Observe the error

### Actual Output

The following objects were not updated from the account to your project:
    - customsearch_xxxxxxx: Unable to parse object XML root tag.

### Expected Output

Successful update of the object

### Anything else?

Reverting the extension to 3.1.3 caused this functionality to work successfully.

packages\sdk-core\src\commands\object\ObjectCommandXml.ts function extractRootTagName does not account for the comment at the top of saved search XML which is causing this issue:

alisyed-19:
This PR fixes #1146

### DIFF
--- a/packages/sdk-core/src/commands/object/ObjectCommandXml.ts
+++ b/packages/sdk-core/src/commands/object/ObjectCommandXml.ts
@@ -19,6 +19,10 @@ const CUSTOM_SEGMENT_TYPE = 'customsegment';
 const CUSTOM_RECORD_TYPE = 'customrecordtype';
 const CUSTOM_RECORD_PREFIX = 'customrecord';
 const IDE_RESULT_KEY = 'result';
+const UPDATE_OBJECT_TYPE_BY_ROOT_TAG: Readonly<Record<string, string>> = {
+	pluginimplementation: 'plugintypeimpl',
+	savedcsvimport: 'csvimport',
+};
 
 export function buildCustomObjectsXml(customObjects: CustomObjectInfo[]): string {
 	const expandedObjects: CustomObjectInfo[] = [];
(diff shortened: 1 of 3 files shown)


packages/sdk-core/src/commands/object/ObjectCommandXml.ts
```diff
@@ -128,14 +132,27 @@ export function extractScriptFileReferences(xmlText: string): string[] {
...
 	const tagMatch = normalizedXml.match(/^<([a-zA-Z0-9_:-]+)/);
 	if (!tagMatch) {
 		throw new Error(translationService.getMessage(OBJECT.ERROR.ROOT_TAG_PARSE_FAILED));
 	}
 	return tagMatch[1];
 }
 
+export function extractObjectTypeForUpdate(xmlText: string): string {
+	const rootTag = extractRootTagName(xmlText).toLowerCase();
+	return UPDATE_OBJECT_TYPE_BY_ROOT_TAG[rootTag] ?? rootTag;
```
TheJorgeBorras, line 157: I did a little bit of digging in the Java codebase because this part looked funny to me (only checking for 2 out of the 30+ record types). Turns out the same two mappings are also repeated in Java metadata lookup code (`SuiteCloudUtil.java:126`).


One relevant difference: Java has an additional, non-root-tag special case: for `customrecord`. When the requested script ID begins with "`customrecordxxxx`", it forces the type to `customrecord`.

We should consider if we also want to front-port that logic or not.

TheJorgeBorras: fixed

TheJorgeBorras (approved):

carolnavarropereira (approved):

PR #1148, opened 2026-08-31
PDPDEVTOOL-5514: Fix problems with glob v10.5.0 (bump up jest to v30.5.0)

TheJorgeBorras:
Overriding Jest dependencies in root package.json (what we did for minimatch) doesn't seem to solve the glob v10.5.0 issue for us.
However, by bumping up the jest version to v30.5.0 we can get rid of it (new dependencies use glob v13.0.6 instead).

### DIFF
--- a/packages/sdk-core/resources/templates/project/unittest/package.json.template
+++ b/packages/sdk-core/resources/templates/project/unittest/package.json.template
@@ -5,8 +5,8 @@
     "test": "jest"
   },
   "devDependencies": {
-	"jest": "30.0.0",
-    "@types/jest": "30.5.0",
+	"jest": "30.5.0",
+    "@types/jest": "30.0.0",
     "@oracle/suitecloud-unit-testing": "^1.9.0"
   }
 }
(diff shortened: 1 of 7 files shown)


packages/sdk-core/resources/templates/project/unittest/package.json.template
```diff
@@ -5,8 +5,8 @@
     "test": "jest"
   },
   "devDependencies": {
-	"jest": "30.4.2",
-    "@types/jest": "30.4.2",
+	"jest": "30.5.0",
+    "@types/jest": "30.5.0",
```
davidecorreu: Latest version for `@types/jest` seems to be 30.0.0 at this moment.

TheJorgeBorras: Done :+1: 

I'm not sure why I was under the impression that `"@types/jest"` needed to match `"jest"` version :sweat_smile: 

The previous 30.4.2 doesn't exist either so this is most likely broken on the last released version as well.

davidecorreu (approved):

carolnavarropereira (approved):

PR #1157, opened 2026-09-08
Feature/pdpdevtool 6615 minor release 26.2 preparation

carolnavarropereira:


TheJorgeBorras (approved):

packages/unit-testing/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@oracle/suitecloud-unit-testing",
-	"version": "1.9.0",
+	"version": "1.10.0",
```
alisyed-19: I think its just libs change, so this should be 1.9.1

